PR #1, opened 2022-07-28
Skeleton

mifrandir:
Adds a sufficient skeleton for users to start adding code.

# ToDo
- [x] libs
- [x] clang-format
- [x] executables
- [x] tests
- [x] git hooks
- [x] code-owners

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,0 @@
-{
-    "files.associations": {
-        "*.ml": "ocaml",
-        "chrono": "cpp"
-    }
-}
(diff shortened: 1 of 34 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
kshxtij, line 1: .vscode can be ignored imo

mifrandir: Since I was planning on making the out of the box experience as good as possible. Assuming you'll stick with VSCode being the recommend editor, we should probably extend that file in order to facilitate that. 

Luckily, CMake Tools handles most of the things so this may not be as required.

kshxtij (approved):

PR #4, opened 2022-08-14
NAV - Basic navigation simulator

mifrandir:


kshxtij (approved):

lib/navigation/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+set(target hyped_navigation)
+file(GLOB headers "${CMAKE_CURRENT_SOURCE_DIR}/*.hpp")
+file(GLOB code "${CMAKE_CURRENT_SOURCE_DIR}/*.cpp")
+add_library(${target} STATIC ${headers} ${code})
+target_include_directories(${target}
+    INTERFACE
+    "${CMAKE_SOURCE_DIR}/lib"
+    "${CMAKE_CURRENT_SOURCE_DIR}"
+    PUBLIC
+    "/usr/include/eigen3"
```
maxguy2001, line 10: Referencing local path?

maxguy2001 (changes requested):
Doesn't compile

kshxtij:
Installing Eigen and then replacing lib/navigation/CMakeLists.txt with 

```
set(target hyped_navigation)
file(GLOB headers "${CMAKE_CURRENT_SOURCE_DIR}/*.hpp")
file(GLOB code "${CMAKE_CURRENT_SOURCE_DIR}/*.cpp")
add_library(${target} STATIC ${headers} ${code})
find_package (Eigen3 3.4 REQUIRED NO_MODULE)
include_directories(${target}
    INTERFACE
    "${CMAKE_SOURCE_DIR}/lib"
    "${CMAKE_CURRENT_SOURCE_DIR}"
)
target_link_libraries(${target} PUBLIC Eigen3::Eigen)
``` 
seems to work. 

However, due to there being no cpp (only hpp) files within lib/navigation, I get the following error while trying to build 
```
CMake Error: Cannot determine link language for target "hyped_navigation".
CMake Error: CMake can not determine linker language for target: hyped_navigation
``` 
This is easily fixed by even an empty c++ file within the folder however `set_target_properties(${target} PROPERTIES LINKER_LANGUAGE C++)` doesn't seem to fix it.

Any thoughts or fixes welcome, because i'd like to get this merged ASAP.

mifrandir:
1. Adding an empty C++ file shouldn't be an issue. We expect navigation to have some of those eventually anyways so I don't see a reason to not havea placeholder.
2. There must be a way to have header only libraries depend on other header only libraries. Actually, Eigen3 is header only so why are we linking against it? You should only have to add its include dir.
3. This is far from done or functional so I'd advise against merging it. Someone from navigation should have a go at this. There are non-trivial decisions to be made on how to model the sensors with the Kalman filter.

kshxtij:
I was linking libraries cause the official Eigen page recommends that https://eigen.tuxfamily.org/dox/TopicCMakeGuide.html. 

We could link dirs but the variable install location on mac vs linux/wsl could potentially cause problems?

kshxtij:
Screenshot 2022-10-20 at 12 29 15

This is what the Eigen3 target is exported as

maxguy2001:
> 1. Adding an empty C++ file shouldn't be an issue. We expect navigation to have some of those eventually anyways so I don't see a reason to not havea placeholder.
> 2. There must be a way to have header only libraries depend on other header only libraries. Actually, Eigen3 is header only so why are we linking against it? You should only have to add its include dir.
> 3. This is far from done or functional so I'd advise against merging it. Someone from navigation should have a go at this. There are non-trivial decisions to be made on how to model the sensors with the Kalman filter.

I know this isn't fully functional and there are still decisions to be make on kalman but we're looking to get started on some of the easier implementations to get the new team used to contributing to hyped and the overall workflow. So ideally looking to merge this soon and start on basic implementations, returning to finish this work later on. Maybe merge and open issues to quantify what needs changed?

kshxtij:
+1 on opening issues to quantify what needs to be changed

maxguy2001:
ok I had to change a cmakelist and define an operator because gtest threw a tantrum. Does this still build for you @mifrandir?

maxguy2001 (approved):

PR #6, opened 2022-10-01
SNS - ADC

SnickeyX:


lib/io/adc.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <vector>
+#include <cstdio>
+
+namespace hyped::io {
+class Adc {
+ public:
+  /**
+   * @param pin
+   */
+  explicit Adc(const uint32_t pin);
```
mifrandir, line 13: What purpose does the `explicit` server here?

lib/io/adc.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <cstdint>
+#include <vector>
+#include <cstdio>
+
+namespace hyped::io {
+class Adc {
```
mifrandir, line 8: Newline between class and namespace.

lib/io/adc.cpp
```diff
@@ -0,0 +1,32 @@
...
+}
+uint16_t Adc::read()
+{
+  char buf[100];
+  snprintf(buf, sizeof(buf), "/sys/bus/iio/devices/iio:device0/in_voltage%i_raw", pin_);
+  file_ = open(buf, O_RDONLY);
+  if (file_ < 0) { /* log that there is a problem reading raw voltage for pin_*/ ;}
+  /*log_.debug("fd: %d", file_); */
+  uint16_t val = readHelper(file_);
+  /* log_.debug("val: %d", val); */
```
mifrandir, line 27: commented code?

lib/io/adc.cpp
```diff
@@ -0,0 +1,32 @@
...
+{
+}
+uint16_t Adc::read()
+{
+  char buf[100];
+  snprintf(buf, sizeof(buf), "/sys/bus/iio/devices/iio:device0/in_voltage%i_raw", pin_);
+  file_ = open(buf, O_RDONLY);
+  if (file_ < 0) { /* log that there is a problem reading raw voltage for pin_*/ ;}
+  /*log_.debug("fd: %d", file_); */
+  uint16_t val = readHelper(file_);
```
mifrandir, line 26: `const`

lib/io/adc.cpp
```diff
@@ -0,0 +1,32 @@
+#include "adc.hpp"
+
+#include <fcntl.h>   // define O_WONLY and O_RDONLY
+#include <unistd.h>  // close()
+
+namespace hyped::io {
+    
+uint16_t readHelper(int fd)
```
mifrandir, line 8: Terrible name. Maybe `resetAndRead`? Or `resetAndRead4`? ...

lib/io/adc.hpp
```diff
@@ -0,0 +1,34 @@
...
+   *         because the BBB has 12-bit ADCs (2^12 = 4096)
+   */
+  uint16_t read();
+
+
+  /**
+   * @param fd is a file path to where we read the raw values from
+   * @return   uint16_t returns two bytes of current voltage data
+   */ 
+  uint16_t readHelper(int fd);
```
mifrandir, line 28: `const int fd`

lib/io/adc.hpp
```diff
@@ -0,0 +1,34 @@
...
+   * @brief reads AIN value from file system
+   *
+   * @return uint16_t return two bytes for [0,4095] range 
+   *         because the BBB has 12-bit ADCs (2^12 = 4096)
+   */
+  uint16_t read();
+
+
+  /**
+   * @param fd is a file path to where we read the raw values from
```
mifrandir, line 25: `fd` is not a file path. It's a file descriptor which isn't the same.

lib/io/adc.hpp
mifrandir: Also, maybe `file_descriptor` for style? I know fd is pretty common but still.

lib/io/adc.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <cstdint>
+#include <vector>
```
mifrandir, line 4: Is this required?

mifrandir (changes requested):

lib/io/adc.hpp
SnickeyX: Thought this was to prevent implicit conversions to uint32_t. Don't know if we really need to enforce that but kept it jic

mifrandir: I don't believe this is what `explicit` does.

This is how I understand it:

Say you have a function

```cpp
void f(const Adc adc) { /* ... */ }
```

and you call it like 

```cpp
const uint32_t pin = 8;
f(pin);
```

this won't work if your constructor is `explicit`. Rather, you have to do 

```cpp
f(Adc(pin));
```

I would consider the former bad practice but I don't see any reason to enforce it explicitly. (Mainly because it adds to confusions such as yours.)

TomLonergan03 (approved):
Looks good to me, test it some time

mifrandir (changes requested):

lib/io/adc.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include <fcntl.h>   // define O_WONLY and O_RDONLY
+#include <unistd.h>  // close()
+
+namespace hyped::io {
+    
+uint16_t resetAndRead4(int file_descriptor)
+{
+  char buf[4];                              // buffer size 4 for fs value
+  lseek(file_descriptor, 0, SEEK_SET);      // reset file pointer
+  read(file_descriptor, buf, sizeof(buf));  // actually consume new data, changes value in buffer
```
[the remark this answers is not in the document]
SnickeyX: added this

lib/io/adc.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+Adc::Adc(const uint32_t pin) : pin_(pin)
+{
+}
+
+uint16_t Adc::read()
+{
+  char buf[100];
+  snprintf(buf, sizeof(buf), "/sys/bus/iio/devices/iio:device0/in_voltage%i_raw", pin_);
+  file_ = open(buf, O_RDONLY);
```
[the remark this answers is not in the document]
SnickeyX: keeping the descriptor from the constructor now and closing with the deconstructor

lib/io/adc.cpp
```diff
@@ -0,0 +1,46 @@
...
+}
+
+Adc::~Adc()
+{
+  close(file_); 
+}
+
+std::optional<uint16_t> Adc::read()
+{
+  const std::optional<uint16_t> val = resetAndRead4(file_);
```
maxguy2001, line 24: Naming. What is `val`?

lib/io/adc.cpp
```diff
@@ -0,0 +1,46 @@
...
+      return *val;
+  }
+  return std::nullopt; 
+}
+
+std::optional<uint16_t> resetAndRead4(int file_descriptor)
+{
+  char buf[4];                                                 // buffer size 4 for fs value
+  lseek(file_descriptor, 0, SEEK_SET);                         // reset file pointer
+  int bytes_read = read(file_descriptor, &buf, sizeof(buf));   // actually consume new data, changes value in buffer
```
maxguy2001, line 37: `const`?

lib/io/adc.cpp
```diff
@@ -0,0 +1,46 @@
...
+  if(bytes_read != sizeof(buf)) 
+  {
+    /*TODO: log that we failed in reading 4 bytes or reached EOF*/
+    return std::nullopt;                                       // returning NULL since we did not get any value
+  }
+  return std::atoi(buf);
```
maxguy2001, line 43: `atoi` returns int but you declared optional return type uint16_t. Also is `std::atoi(buf)` strictly positive if return type is uint?

maxguy2001 (changes requested):
Small changes

lib/io/adc.cpp
SnickeyX: `buf` will strictly have nonnegative values but that `atoi` point does give me pain - appreciate the catch tho

lib/io/adc.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace hyped::io { 
+
+Adc::Adc(const uint32_t pin) : pin_(pin)
+{
+  char buf[100];
+  snprintf(buf, sizeof(buf), "/sys/bus/iio/devices/iio:device0/in_voltage%i_raw", pin_);
+  file_ = open(buf, O_RDONLY);
+  if (file_ < 0) { /* TODO: log that there is a problem reading raw voltage for pin_*/ ;}
+  /* TODO: log the file descriptor */
```
kshxtij, line 14: What do you mean by `log the file descriptor`, that doesn't make sense imo

lib/io/adc.cpp
```diff
@@ -21,26 +20,32 @@ Adc::~Adc()
...
+  }
+  const int bytes_read = read(file_descriptor, &read_buffer, sizeof(read_buffer));   // actually consume new data, changes value in buffer
+  if(bytes_read != sizeof(read_buffer)) 
   {
     /*TODO: log that we failed in reading 4 bytes or reached EOF*/
-    return std::nullopt;                                       // returning NULL since we did not get any value
+    return std::nullopt;                                                             // returning NULL since we did not get any value
   }
-  return std::atoi(buf);
+  const int raw_voltage = std::atoi(read_buffer);
```
mifrandir, line 47: I think your use of `atoi` is wrong altogether. Surely the `read_buffer` doesn't contain an ASCII representation of the value? 

https://en.cppreference.com/w/cpp/string/byte/atoi

mifrandir (changes requested):

lib/io/adc.cpp
SnickeyX: Its supposed to be in ASCII since we are reading numbers from a file as characters

kshxtij (approved):

lib/io/adc.cpp
mifrandir: surely the file is just raw binary, though?

kshxtij: That's how the interface for ADC works.

```
[root@beaglebone:/sys/devices/]ocp.3/helper.12# cat AIN0
0
[root@beaglebone:/sys/devices/]ocp.3/helper.12# cat AIN0
427
[root@beaglebone:/sys/devices/]ocp.3/helper.12# cat AIN0
746
[root@beaglebone:/sys/devices/]ocp.3/helper.12# cat AIN0
921
```

https://www.linux.com/training-tutorials/how-get-analog-input-beaglebone-black/

mifrandir: hmmm, this seems wrong but i don't know how to disprove it

mifrandir (approved):

maxguy2001 (commented):
Check `uint` values. If definitely strictly positive then all good.

maxguy2001 (approved):

PR #8, opened 2022-10-16
MTC - Inital setup

TomLonergan03:
Mostly copied class structure from last year
Added io files for CAN
Defined basic types for can frames


lib/core/types.hpp
```diff
@@ -1,7 +1,17 @@
 #pragma once
+#include <stdint.h>
 
 namespace hyped::core {
 
 enum class LowOrHigh { kLow = 0, kHigh };
 
+struct CanFrame
```
kshxtij, line 8: Is this struct from a datasheet/spec?

TomLonergan03: This is the [SocketCAN spec](https://docs.kernel.org/networking/can.html)

kshxtij (approved):
LGTM as an initial setup

lib/core/types.hpp
```diff
@@ -1,7 +1,17 @@
 #pragma once
+#include <stdint.h>
```
maxguy2001, line 2: cstdint?

lib/core/types.hpp
```diff
@@ -1,7 +1,17 @@
...
 enum class LowOrHigh { kLow = 0, kHigh };
 
+struct CanFrame
+{
+    uint32_t can_id; /* 32 bit CAN_ID + EFF/RTR/ERR flags */
+    uint8_t can_dlc; /* frame payload length in byte (0 .. CAN_MAX_DLEN) */
+    uint8_t __pad;   /* padding */
+    uint8_t __res0;  /* reserved / padding */
+    uint8_t __res1;  /* reserved / padding */
+    uint8_t data[8];
```
maxguy2001, line 15: `std::array<int8_t, 8>`?

lib/motorcontrollers/can_processor.cpp
```diff
@@ -0,0 +1,22 @@
...
+    //TODO implement
+}
+
+bool CanProcessor::sendMessage()
+{
+    //TODO implement
+    return false;
+}
+
+void CanProcessor::processMessage(core::CanFrame frame)
```
maxguy2001, line 16: No idea how  this area works in general but can `frame` be `const`? Is the frame actually modified in the function or is it just read in and used?

lib/motorcontrollers/can_processor.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+class CanProcessor
+{
+public:
+    CanProcessor();
+
+    bool sendMessage();
+    void processMessage(core::CanFrame frame);
+}; 
+}
```
maxguy2001, line 16: `//hyped::motorcontrollers`? maybe run clang-format?

lib/motorcontrollers/messages.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <stdint.h>
```
maxguy2001, line 3: `cstdint`?

maxguy2001 (changes requested):
A few stylistic changes but otherwise looking good!

lib/core/types.hpp
```diff
@@ -1,7 +1,17 @@
 #pragma once
+#include <stdint.h>
 
 namespace hyped::core {
 
 enum class LowOrHigh { kLow = 0, kHigh };
 
+struct CanFrame
+{
+    uint32_t can_id; /* 32 bit CAN_ID + EFF/RTR/ERR flags */
```
maxguy2001, line 10: Are we certain that that all these numbers are strictly positive? I don't know CAN, just a quick sanity check.

kshxtij: Don't think CAN ids can be negative

TomLonergan03: The [socketCAN spec](https://docs.kernel.org/networking/can.html) defines them this way so I'm pretty sure it is true.
CAN frame IDs are essentially bitmasks so don't really have negatives, CAN frames can't have negative lengths, the data could be negative but this is representing the actual bytes we receive, rather than the specific values. Pad, res0, and res1 are for things like a checksum and are dealt with by socketCAN instead of us.

lib/core/types.hpp
TomLonergan03: is that functionally different from data[8]? if it is, then will need to be tested to see if it works with socketCAN

lib/core/types.hpp
kshxtij: Yeah i reckon this is fine

lib/core/types.hpp
maxguy2001: Not 100% sure. Just that `uint8_t data[8]` defines a C-style array but we're writing C++ so usually want to use C++ style arrays?

lib/core/types.hpp
maxguy2001: Ok fair enough, just checking that we're not defaulting to `uint` without considering data values.

lib/core/types.hpp
TomLonergan03: i'll change it over, once we get socketCAN working then can test for sure

kshxtij: std::array is just a wrapper for a generic c style array afaik so it shouldn't make a difference but yes testing it is a good idea

maxguy2001 (approved):
Good stuff

PR #9, opened 2022-10-19
SNS - I2C

SnickeyX:


lib/io/i2c.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+  const auto ret = read(fd_, data, len);
+  if (ret != static_cast<int>(len)) {
+    /*log "Could not read from i2c device"*/;
+    return -1;
+  }
+  return ret;
+}
+
+int I2c::writeData(const uint32_t address, uint8_t *data, const size_t len)
```
mifrandir, line 62: The point of this class is to abstract away the ugliness. Returning an `int` in C-fashin isn't helping with that. Consider making an enum class `I2c::WriteResult` or possibly a struct (if you really care about how many bytes were written, which I doubt).

lib/io/i2c.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+  sensor_address_ = address;
+  const int ret   = ioctl(fd_, I2C_SLAVE, address);
+  if (ret < 0) {
+    /*log "Could not set sensor address" */;
+    return;
+  }
+}
+
+int I2c::readData(const uint32_t address, uint8_t *data, const size_t len)
```
mifrandir, line 45: `len` isn't in line with the style guide ... (neither is `fd` so not sure how rigorous you guys are here).

I am also wondering, and this may be going too far, how often do we anticipate that the data is of variable length? Wouldn't it be better to parametrise over the length there? What I'm thinking about is 

```cpp
template<int length>
I2c::ReadResult I2c::readData(const uint32_t address, const std::array<uint8_t, length> data) { /* ... */ }
```

Of course this only works if we always know the size of the message we are sending at compile time. For sensors readings this will be the case!

One needs to be careful not to ruin the callsites (and I don't know how C++ handles this). What we want is something like 

```
i2c.readData(address, message.toArray());
```

Rather than 
```cpp
const auto data = message.toArray();
i2c.readData<data.size()>(address, data);
```

But even in that case the latter is still not any more verbose than your interface!

lib/io/i2c.cpp
```diff
@@ -0,0 +1,77 @@
...
+{
+  if (fd_ < 0) {
+    /* log "Could not find i2c device"*/;
+    return -1;
+  }
+
+  if (sensor_address_ != address) { setSensorAddress(address); }
+
+  const auto ret = read(fd_, data, len);
+  if (ret != static_cast<int>(len)) {
```
mifrandir, line 55: On 64-bit machines, you have chosen to cast `size_t` down to `int` rather than casting `int` up to `size_t`. Why?

lib/io/i2c.cpp
```diff
@@ -0,0 +1,77 @@
...
+  close(fd_);
+}
+
+void I2c::setSensorAddress(const uint32_t address)
+{
+  if (fd_ < 0) {
+    /* log "Could not find i2c device"*/;
+    return;
+  }
+
```
mifrandir, line 36: empty line

lib/io/i2c.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+int I2c::writeData(const uint32_t address, uint8_t *data, const size_t len)
+{
+  if (fd_ < 0) {
+    /*log "Could not find i2c device"*/ ;
+    return -1;
+  }
+  if (sensor_address_ != address) { setSensorAddress(address); }
+
```
mifrandir, line 69: empty line

mifrandir (changes requested):

lib/io/i2c.cpp
kshxtij: Data length depends on the size of the hardware register (on the sensor/other auxilarry device). Like the temp sensor for example, has 8 bit registers; so to transmit 16 bits of temp data it uses two registers DATA_T_H and DATA_T_L. 

So i don't think we should ever be dealing with variable length data?

lib/io/i2c.cpp
kshxtij: +1 on the result enum and yes don't really care about the number of bytes written

SnickeyX: will do

lib/io/i2c.cpp
mifrandir: This is variable at compile time, not at runtime. You would know what sensor your are using by the time you write the code, no?

mifrandir: Alright, try the above and see what it looks/feels like?

mifrandir:
On a more general note, this isn't particularly testable. I just remembered why I wrote the `Gpio` dummy the way I did: You want to be able to change the particular type of GPIO/ADC/I2C instance that the code runs with in a single place. This then allows you to do several things:

1. Test whether the system works in principle, without any hardware.
2. Write some code that uses the interface and play around with it without requiring hardware. Making these interfaces is hard. Not because the code for reading I2C is particularly tricky, but because you want the interface to be pleasant to use while satisfying all users. One step towards that is to try and use it yoruself in some tests (which may not test for any property other than showing how the interface should be used)

SnickeyX:
The way I was approaching this currently is testing the code straight away with hardware (since we have the temperature sensor and accelerometer already bread boarded and ready to test with). Though I must mention, this PR was simply to get me started working on I2C properly, since this code from last year will not actually work (we need to be able to read from and write to partiuclar registers which we can't do here). At the moment, I am testing some I2C code locally with a temperature sensor and seeing if it actually works in giving me the expected values (expect this PR latest by the end of next week) (and I want to be able to do this with all the other comms protocols). 

As for the general interfaces you mentioned, I will try to get working on those but as you said its difficult to know whether it works in a practical setting which is exactly why I want to straight away push for testing with actual hardware (luckily electronics agrees with us this time around).

lib/io/i2c.cpp
```diff
@@ -0,0 +1,77 @@
+#include "i2c.hpp"
+
+#include <fcntl.h>
+#include <unistd.h>
+
+#include <sys/ioctl.h>
+
```
maxguy2001, line 7: A lot of C headers? Can these be replaced by better C++ headers or are these better?

lib/io/i2c.cpp
```diff
@@ -0,0 +1,77 @@
...
+#endif
+
+namespace hyped::io {
+
+I2c::I2c(const uint8_t bus_address)
+    : sensor_address_(0)
+{
+  char path[13];  // up to "/dev/i2c-255"
+  sprintf(path, "/dev/i2c-%d", bus_address);
+  fd_ = open(path, O_RDWR, 0);
```
maxguy2001, line 21: naming?

maxguy2001 (commented):
As someone who does not visit this area of the codebase very often, some of the naming such as `ret` and `fd_` makes no sense to me. Are these really common and intuitive abbreviations that I'm not aware of or could they be named better?

lib/io/i2c.cpp
kshxtij: ioctl has no equivalent

lib/io/i2c.cpp
SnickeyX: will fix once I get working I2C code ready

SnickeyX:
The code now actually works and has been tested with a temperature sensor ＼(~o~)／

lib/io/i2c.cpp
```diff
@@ -0,0 +1,94 @@
...
+  }
+  if (sensor_address_ != device_address) { setSensorAddress(device_address); }
+  // Contains data which we write: first the register address then the data which we write to it
+  uint8_t write_buffer[2] = {register_address, data};	
+  // Writing 2 bytes to our write buffer 
+  const auto num_bytes_written = write(file_descriptor_, write_buffer, 2);
+  if (num_bytes_written != 2) 
+  {
+    /* log "Could not write to i2c device" */;
+    I2cWriteResult::kFailure;
```
kshxtij, line 89: Missing a return?

lib/io/i2c.cpp
```diff
@@ -0,0 +1,94 @@
...
+  if (file_descriptor_ < 0) 
+  {
+    /* log "Could not find i2c device while reading"*/;
+    return std::nullopt; 
+  }
+  if (sensor_address_ != device_address) { setSensorAddress(device_address); }
+  // Contains data which we will read 
+  uint8_t read_buffer[1];		
+  // Contains data which we need to write
+  uint8_t write_buffer[1] = {reg_address};
```
kshxtij, line 57: const?

kshxtij (changes requested):
Good job on testing directly with hardware 👍

kshxtij (approved):
LGTM

lib/io/i2c.cpp
```diff
@@ -0,0 +1,95 @@
...
+  // Writing the register address so we switch to it
+  const int num_bytes_written = write(file_descriptor_, write_buffer, 1);
+  if(num_bytes_written != 1) 
+  {
+    /* log "could not write to i2c device"*/
+    return std::nullopt; 
+  }
+  // Reading one byte to our read buffer 
+  const int num_bytes_read = read(file_descriptor_, read_buffer, 1); 
+  if( num_bytes_read != 1 )
```
mifrandir, line 67: Formatting?

lib/io/i2c.cpp
```diff
@@ -0,0 +1,95 @@
...
+{
+  if (file_descriptor_ < 0) 
+  {
+    /*log "Could not find i2c device while writing"*/ ;
+    return I2cWriteResult::kFailure;
+  }
+  if (sensor_address_ != device_address) { setSensorAddress(device_address); }
+  // Contains data which we write: first the register address then the data which we write to it
+  const uint8_t write_buffer[2] = {register_address, data};	
+  // Writing 2 bytes to our write buffer
```
mifrandir, line 85: Pointless comment

lib/io/i2c.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+I2cWriteResult I2c::writeByte(const uint8_t device_address, const uint8_t register_address, const uint8_t data)
+{
+  if (file_descriptor_ < 0) 
+  {
+    /*log "Could not find i2c device while writing"*/ ;
+    return I2cWriteResult::kFailure;
+  }
+  if (sensor_address_ != device_address) { setSensorAddress(device_address); }
+  // Contains data which we write: first the register address then the data which we write to it
```
mifrandir, line 83: Hmmm... I can see that?

mifrandir (changes requested):

lib/io/i2c.cpp
```diff
@@ -62,9 +62,8 @@ std::optional<uint8_t> I2c::readByte(const uint8_t device_address, const uint8_t
     /* log "could not write to i2c device"*/
     return std::nullopt; 
   }
-  // Reading one byte to our read buffer 
   const int num_bytes_read = read(file_descriptor_, read_buffer, 1); 
-  if( num_bytes_read != 1 )
+  if(num_bytes_read != 1)
```
mifrandir, line 66: Are you using clang-format? This still doesn't look correct.

lib/io/i2c.cpp
mifrandir: What about this @kshxtij @SnickeyX ?

mifrandir (changes requested):
Merge master and lmk what we're doing about the templating.

lib/io/i2c.cpp
kshxtij: I think we changed it to readByte so the length stuff doesn't really matter

lib/io/i2c.cpp
```diff
@@ -0,0 +1,85 @@
...
+  }
+  sensor_address_        = device_address;
+  const int return_value = ioctl(file_descriptor_, I2C_SLAVE, device_address);
+  if (return_value < 0) {
+    /*log "Could not set sensor address" */;
+    return;
+  }
+}
+
+std::optional<uint8_t> I2c::readByte(const uint8_t device_address, const uint8_t register_address)
```
mifrandir, line 43: Do we expect to mostly read single bytes? Doing repeated reads like that may end up being both inefficient and ugly from the callees perspective.

kshxtij: All of the i2c hardware only uses 8 bit registers and single byte communication so we're not concerned about multiple bytes read (atleast yet)

SnickeyX: Also, though it's possible that we need to read from two registers that are back to back in memory (for example data_high and data_low being at 0x06 and 0x07) and having  `readBytes` would help, but I think having one function read from both makes it harder to understand what the code is doing. (Plus I don't think we are going to need to read more than 2 registers at one time)

mifrandir (approved):

PR #10, opened 2022-10-22
UTILS - Logger

kshxtij:
- [x] Added ILogger Interface
- [x] Implement Basic Logger 
- [x] Unit Tests

Verbosity levels are defined based on the ros and roscpp spec found [here](http://wiki.ros.org/Verbosity%20Levels).

While in a Ros node enviroment, we'd pass in RosLogger into the classes which implements logging to stdout, stderr, text file as well as /rosout (a ros topic that runs centrally and collects all logging message). RosCpp defines these using macros within the standard ros library so RosLogger would just wrap the ros provided macro/definitions.

### DIFF
--- a/lib/core/logger.cpp
+++ b/lib/core/logger.cpp
@@ -14,12 +14,12 @@ Logger::Logger(const char *const label, const LogLevel level, const core::ITimeS
 void Logger::printHead(FILE *file, const char *title)
 {
   const auto time_point         = time_source_.now();
-  const auto ttime_t            = std::chrono::system_clock::to_time_t(time_point);
-  const auto time_point_seconds = std::chrono::system_clock::from_time_t(ttime_t);
+  const auto timestamp            = std::chrono::system_clock::to_time_t(time_point);
+  const auto time_point_seconds = std::chrono::system_clock::from_time_t(timestamp);
   const std::chrono::milliseconds time_point_milliseconds
     = duration_cast<std::chrono::milliseconds>(time_point - time_point_seconds);
   const long long time_milliseconds = time_point_milliseconds.count();
-  const std::tm *time_struct        = localtime(&ttime_t);
+  const std::tm *time_struct        = localtime(&timestamp);
   fprintf(file, "%02d:%02d:%02d.%03lld %s[%s] ", time_struct->tm_hour, time_struct->tm_min,
           time_struct->tm_sec, time_milliseconds, title, label_);
 }
--- a/lib/core/logger.cpp
+++ b/lib/core/logger.cpp
@@ -13,15 +13,18 @@ Logger::Logger(const char *const label, const LogLevel level, const core::ITimeS
 
 void Logger::printHead(FILE *file, const char *title)
 {
-  const auto time_point = time_source_.now();
-  const auto ttime_t = std::chrono::system_clock::to_time_t(time_point);
+  const auto time_point         = time_source_.now();
+  const auto ttime_t            = std::chrono::system_clock::to_time_t(time_point);
   const auto time_point_seconds = std::chrono::system_clock::from_time_t(ttime_t);
-  const std::chrono::milliseconds time_point_milliseconds = duration_cast<std::chrono::milliseconds>(time_point - time_point_seconds);
+  const std::chrono::milliseconds time_point_milliseconds
+    = duration_cast<std::chrono::milliseconds>(time_point - time_point_seconds);
   const std::tm *time_struct = localtime(&ttime_t);
-  fprintf(file, "%02d:%02d:%02d.%03lld %s[%s] ", time_struct->tm_hour, time_struct->tm_min, time_struct->tm_sec, time_point_milliseconds.count(), title, label_);
+  fprintf(file, "%02d:%02d:%02d.%03lld %s[%s] ", time_struct->tm_hour, time_struct->tm_min,
+          time_struct->tm_sec, time_point_milliseconds.count(), title, label_);
 };
 
-void Logger::log(const LogLevel level, const char *format, ...){
+void Logger::log(const LogLevel level, const char *format, ...)
+{
   FILE *file;
   if (level_ <= level && level_ != LogLevel::kNone) {
     switch (level) {
@@ -48,4 +51,4 @@ void Logger::log(const LogLevel level, const char *format, ...){
   }
 }
 
-};  // namespace hyped::utils
+};  // namespace hyped::core
--- a/lib/core/logger.cpp
+++ b/lib/core/logger.cpp
@@ -37,10 +37,10 @@ void Logger::intToLevel(const int level)
 
 void Logger::printHead(FILE *file, const char *title)
 {
-  const auto tp = timer_.now();
-  const auto ttime_t = std::chrono::system_clock::to_time_t(tp);
+  const auto time_point = timer_.now();
+  const auto ttime_t = std::chrono::system_clock::to_time_t(time_point);
   const auto tp_seconds = std::chrono::system_clock::from_time_t(ttime_t);
-  const std::chrono::milliseconds tp_milliseconds = duration_cast<std::chrono::milliseconds>(tp - tp_seconds);
+  const std::chrono::milliseconds tp_milliseconds = duration_cast<std::chrono::milliseconds>(time_point - tp_seconds);
   const std::tm *time_struct = localtime(&ttime_t);
   fprintf(file, "%02d:%02d:%02d.%03lld %s[%s] ", time_struct->tm_hour, time_struct->tm_min, time_struct->tm_sec, tp_milliseconds.count(), title, module_);
 };
--- a/lib/core/logger.cpp
+++ b/lib/core/logger.cpp
@@ -24,9 +24,6 @@ void Logger::intToLevel(const int level)
       level_ = Level::kWarn;
       break;
     case 4:
-      level_ = Level::kError;
-      break;
-    case 5:
       level_ = Level::kFatal;
       break;
     default:
@@ -88,19 +85,6 @@ void Logger::warn(const char *format, ...)
   FILE *file = stderr;
   if (level_ == Level::kWarn || level_ == Level::kInfo || level_ == Level::kDebug) {
     printHead(file, "WARN");
-    va_list args;
-    va_start(args, format);
-    print(file, format, args);
-    va_end(args);
-  }
-};
-
-void Logger::error(const char *format, ...)
-{
-  FILE *file = stderr;
-  if (level_ == Level::kError || level_ == Level::kWarn || level_ == Level::kInfo
-      || level_ == Level::kDebug) {
-    printHead(file, "ERROR");
     va_list args;
     va_start(args, format);
     print(file, format, args);
--- a/lib/core/logger.cpp
+++ b/lib/core/logger.cpp
@@ -18,10 +18,11 @@ void Logger::printHead(FILE *file, const char *title)
   const auto time_point_seconds = std::chrono::system_clock::from_time_t(ttime_t);
   const std::chrono::milliseconds time_point_milliseconds
     = duration_cast<std::chrono::milliseconds>(time_point - time_point_seconds);
-  const std::tm *time_struct = localtime(&ttime_t);
+  const long long time_milliseconds = time_point_milliseconds.count();
+  const std::tm *time_struct        = localtime(&ttime_t);
   fprintf(file, "%02d:%02d:%02d.%03lld %s[%s] ", time_struct->tm_hour, time_struct->tm_min,
-          time_struct->tm_sec, time_point_milliseconds.count(), title, label_);
-};
+          time_struct->tm_sec, time_milliseconds, title, label_);
+}
 
 void Logger::log(const LogLevel level, const char *format, ...)
 {
@@ -51,4 +52,4 @@ void Logger::log(const LogLevel level, const char *format, ...)
   }
 }
 
-};  // namespace hyped::core
+}  // namespace hyped::core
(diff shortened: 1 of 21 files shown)


lib/io/gpio.cpp
```diff
@@ -2,20 +2,22 @@
 
 namespace hyped::io {
 
-Gpio::Gpio()
+Gpio::Gpio(hyped::utils::ILogger *log) : log_(log)
```
mifrandir, line 5: I dont like that this is a pointer. The logger wont hold a lot of data, right? So passing it by value should be fine.

lib/utils/logger.cpp
```diff
@@ -0,0 +1,124 @@
...
+{
+  switch (level) {
+    case 0:
+      level_ = Level::kNone;
+      break;
+    case 1:
+      level_ = Level::kDebug;
+      break;
+    case 2:
+      level_ = Level::kInfo;
```
mifrandir, line 21: Way too many levels to be practical. At most have fatal, warnings, info, and debug.

mifrandir (changes requested):
Maybe logging should live in core? This seems pretty essential.

It also seems like a step back to wrap macros in function calls but I dont suppose there's any way around this without linking against ROS?

lib/utils/logger.cpp
kshxtij: Yeah that's what i thought too, i just followed ROS's levels but i'll remove kError.

kNone is there for tests when we don't want output to stdout/stderr

kshxtij:
I think ROS's logging utility is too useful to not link against especially when it has the ability to centralise logging as well, I agree that wrapping macros with functions isn't ideal so i'll look into calling ros's logger directly.

kshxtij:
The only logging api exposed by ROS is through the macros, while it isn't ideal to wrap them with functions i think that's the most realistic way to do it.  Especially since we definitely want to be using RosLogger within a multi-node, multi-controller setup.

It also directly links with some other ros extensions we might want to use in the future.

kshxtij:
As soon as this PR is reviewed and merged, i'll go through the preexisting code as well as currently open PRs to add logging where needed (like the comments in adc/i2c by Ishan)

lib/core/logger.cpp
```diff
@@ -0,0 +1,108 @@
...
+    va_start(args, format);
+    print(file, format, args);
+    va_end(args);
+  }
+};
+
+void Logger::fatal(const char *format, ...)
+{
+  FILE *file = stderr;
+  if (level_ == Level::kFatal || level_ == Level::kError || level_ == Level::kWarn
```
TomLonergan03, line 98: What's the point of kError as a level? It only seems to be used here and isn't accessible through `intToLevel`.
Could it be removed and just use kFatal instead?

lib/core/logger.cpp
```diff
@@ -0,0 +1,108 @@
...
+  vfprintf(file, format, args);
+  fprintf(file, "\n");
+};
+
+void Logger::debug(const char *format, ...)
+{
+  FILE *file = stdout;
+  if (level_ == Level::kDebug) {
+    printHead(file, "DEBUG");
+    va_list args;
```
TomLonergan03, line 64: Is there a reason for doing `va_list args`, `va_start` etc. in each logging level instead of within print itself?

TomLonergan03 (changes requested):

lib/core/logger.cpp
kshxtij: Forgot to remove it from the enum and these functions when i deleted log.error()

lib/core/logger.cpp
kshxtij: Can only use `va_start` within a function that accepts `...` (or ellipsis) as a parameter

TomLonergan03: makes sense

TomLonergan03 (approved):

lib/core/logger.cpp
```diff
@@ -0,0 +1,107 @@
...
+    case 4:
+      level_ = Level::kFatal;
+      break;
+    default:
+      level_ = Level::kNone;
+      break;
+  }
+};
+
+void Logger::printHead(FILE *file, const char *title)
```
maxguy2001, line 35: Can parts of this function not be `const`? Also naming in this function?

lib/core/logger.cpp
```diff
@@ -0,0 +1,107 @@
...
+{
+  const std::time_t result = std::time(nullptr);
+  const tm *current_time        = localtime(&result);
```
mifrandir, line 38: This is not how we usually get timestamps, right? This should use a time source instead. This might make it deterministic in testing, too?

lib/core/logger.cpp
```diff
@@ -0,0 +1,107 @@
...
+  fprintf(file, "%02d:%02d:%02d", current_time->tm_hour, current_time->tm_min, current_time->tm_sec);
+
+  static const bool print_micro = true;
+  if (print_micro) {
```
mifrandir, line 42: What is  this `print_micro` good for?

lib/core/logger.cpp
```diff
@@ -0,0 +1,107 @@
...
+      level_ = Level::kNone;
+      break;
+  }
+};
+
+void Logger::printHead(FILE *file, const char *title)
+{
+  const std::time_t result = std::time(nullptr);
+  const tm *current_time        = localtime(&result);
+  fprintf(file, "%02d:%02d:%02d", current_time->tm_hour, current_time->tm_min, current_time->tm_sec);
```
mifrandir, line 39: In general, I'm not a fan of the `printf` style. I know there are some merits in terms of performance but I'm not sure that's worth it?

lib/core/logger.cpp
```diff
@@ -0,0 +1,107 @@
...
+  if (level_ == Level::kWarn || level_ == Level::kInfo || level_ == Level::kDebug) {
+    printHead(file, "WARN");
+    va_list args;
+    va_start(args, format);
+    print(file, format, args);
+    va_end(args);
+  }
+};
+
+void Logger::fatal(const char *format, ...)
```
mifrandir, line 95: All of these functions look identical. Why not have a private `void log(const std::string label, const Level level, std::ofstream output_stream)`?

The strings can be static in their parent calls so don't worry about allocations.

lib/core/logger.hpp
```diff
@@ -0,0 +1,46 @@
...
+#include <optional>
+#include <string>
+
+namespace hyped::utils {
+
+enum class Level { kNone, kDebug, kInfo, kWarn, kFatal };
+
+class ILogger {
+ public:
+  ILogger(const char *const module, const Level level) : module_(module), level_(level){};
```
mifrandir, line 13: Do we really want constructors here?

mifrandir (changes requested):

lib/core/logger.hpp
kshxtij: I think i did this because module_ and level_ are members of ILogger and not Logger so to set them i'd need to defer to ILoggers constructor?

lib/core/logger.cpp
kshxtij: You'd need to have the 

```cpp
    va_list args;
    va_start(args, format);
    
    va_end args;
```

within the `fatal`, `info`, `debug`, `warn` functions due to the ellipsis parameters to accept multiple arguments so i don't see what you could move to another function

lib/core/logger.cpp
kshxtij: I don't think this is a problem because we use printf style args within the logging functions themselves unless we replace everything with std::format and Variadic templates but im not sure that's needed

lib/core/logger.cpp
kshxtij: Will switch over to the constructor accepting a ITimeSource and using now().

lib/core/logger.hpp
kshxtij: Changed the location of module_ and level_ to Logger instead of ILogger

lib/core/logger.cpp
SnickeyX, line 42: Is this supposed to always be true?

SnickeyX (changes requested):

lib/core/logger.cpp
kshxtij: Removed, was an artifact of last year

SnickeyX (changes requested):

test/core/logger.cpp
```diff
@@ -0,0 +1,59 @@
+#include <gtest/gtest.h>
+
+#include <utils/manual_time.hpp>
+#include <core/logger.hpp>
+
+namespace hyped::test {
+
+void logHelper(core::LogLevel level, utils::ManualTime &manual_time, std::string expected_output, bool stdout)
```
mifrandir, line 8: `const`


`use_stdout`

mifrandir: Also, maybe call it `testLog`?

test/core/logger.cpp
```diff
@@ -0,0 +1,59 @@
...
+TEST(Logger, DebugStdout)
+{
+  utils::ManualTime manual_time;
+  logHelper(core::LogLevel::kDebug, manual_time, "01:00:00.000 DEBUG[test] test\n", true);
+};
+
+TEST(Logger, InfoStdout)
+{
+  utils::ManualTime manual_time;
+  logHelper(core::LogLevel::kInfo, manual_time, "01:00:00.000 INFO[test] test\n", true);
```
mifrandir, line 33: Maybe combine these into one test? Just have subsequent calls to logHelper.

lib/core/logger.hpp
```diff
@@ -0,0 +1,30 @@
...
+enum class LogLevel { kNone = 0, kDebug, kInfo, kFatal };
+
+class ILogger {
+ public:
+  virtual void log(const LogLevel level, const char *format ...){};
+};
+
+class Logger : public ILogger {
+ public:
+  Logger(const char *const module, const LogLevel level, const core::ITimeSource &timer);
```
mifrandir, line 20: module

lib/core/logger.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+  void log(const LogLevel level, const char *format, ...);
+
+  private:
+    void printHead(FILE *file, const char *title);
+    const char *const label_;
+    LogLevel level_;
```
mifrandir, line 27: const

lib/core/logger.hpp
```diff
@@ -0,0 +1,30 @@
...
+    const core::ITimeSource &timer_;
```
mifrandir, line 28: `time_source`

lib/core/logger.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include "time.hpp"
+
+namespace hyped::core {
+
+enum class LogLevel { kNone = 0, kDebug, kInfo, kFatal };
+
+class ILogger {
+ public:
+  virtual void log(const LogLevel level, const char *format ...){};
```
mifrandir, line 15: `=0`

also check the ROS interface but it may now be possible to put a `const` at the end here, too.

mifrandir (changes requested):

TomLonergan03 (approved):
This is much nicer having ditched all the duplicate functions

lib/core/logger.cpp
```diff
@@ -0,0 +1,51 @@
...
+      time_source_(time_source_)
+{
+}
+
+void Logger::printHead(FILE *file, const char *title)
+{
+  const auto time_point = time_source_.now();
+  const auto ttime_t = std::chrono::system_clock::to_time_t(time_point);
+  const auto tp_seconds = std::chrono::system_clock::from_time_t(ttime_t);
+  const std::chrono::milliseconds tp_milliseconds = duration_cast<std::chrono::milliseconds>(time_point - tp_seconds);
```
SnickeyX, line 19: naming

SnickeyX (changes requested):

SnickeyX (approved):

lib/core/logger.cpp
```diff
@@ -0,0 +1,54 @@
...
+    }
+    va_list args;
+    va_start(args, format);
+    vfprintf(file, format, args);
+    fprintf(file, "\n");
+    va_end(args);
+  }
+}
+
+};  // namespace hyped::core
```
maxguy2001, line 54: ...

lib/core/logger.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+void Logger::printHead(FILE *file, const char *title)
+{
+  const auto time_point         = time_source_.now();
+  const auto ttime_t            = std::chrono::system_clock::to_time_t(time_point);
+  const auto time_point_seconds = std::chrono::system_clock::from_time_t(ttime_t);
+  const std::chrono::milliseconds time_point_milliseconds
+    = duration_cast<std::chrono::milliseconds>(time_point - time_point_seconds);
+  const std::tm *time_struct = localtime(&ttime_t);
+  fprintf(file, "%02d:%02d:%02d.%03lld %s[%s] ", time_struct->tm_hour, time_struct->tm_min,
```
maxguy2001, line 22: :(

lib/core/logger.cpp
```diff
@@ -0,0 +1,54 @@
...
+    : label_(label),
+      level_(level),
+      time_source_(time_source_)
+{
+}
+
+void Logger::printHead(FILE *file, const char *title)
+{
+  const auto time_point         = time_source_.now();
+  const auto ttime_t            = std::chrono::system_clock::to_time_t(time_point);
```
mifrandir, line 17: Identifiers with suffix `_t` usually refer to a type.

maxguy2001 (approved):
Builds

mifrandir (approved):

PR #12, opened 2022-10-24
SNS - GPIO 

Knhwong:


kshxtij (commented):
Why were the files within scripts/hooks removed?

lib/io/gpio.cpp
```diff
@@ -1,22 +1,289 @@
...
+
+  char buf[10];
+  for (uint32_t pin : exported_pins) {
+    snprintf(buf, sizeof(buf), "%d", pin);
+    write(fd, buf, strlen(buf) + 1);
+  }
+  close(fd);
 }
+//Pins must be set before we export and attach
+void Gpio::exportGPIO()
```
mifrandir, line 68: exportGpio

lib/io/gpio.cpp
```diff
@@ -1,22 +1,289 @@
...
+    if (fdset.revents & POLLERR) { log_.error("an error on wait of gpio %d: %d", pin_, errno); }
+  }
+
+  log_.error("some error on wait of gpio %d", pin_);
+  return -1;
+}
+
+
+
+uint8_t Gpio::Write(const uint8_t pin, const bool SetOrClear) {
```
mifrandir, line 191: Gpio::write

mifrandir (changes requested):
Id get rid of most of this old interface. What was so bad about the dummy that was there before?

Knhwong:
#34

PR #14, opened 2022-10-24
MTC - preliminary error handling functions 

syntax:
Added skeleton functions for error handling; processErrorMessage, processEmergencyMessage, processSdoMessage, processNmtMessage.

Implementing error codes and their description from https://www.emdrive-mobility.com/portfolio/emdrive-500/ controller.

Logging will be fixed when logging is updated on master, just being used as a placeholder.

Working with @TristCrocker 

WIP


lib/motorcontrollers/controller.cpp
```diff
@@ -3,5 +3,73 @@
...
+      case 0xFF10:
+        log_.error("ERROR_CONTROLLER_OVERTEMPERATURE");
+        break;
+      case 0x3210:
+        log_.error("ERROR_DC_LINK_OVERVOLTAGE");
+        break;
+    }
+  }
+
+void processSdoMessage(core::types::Canframe data) {
```
mifrandir, line 67: const

lib/motorcontrollers/controller.cpp
```diff
@@ -3,5 +3,73 @@
 namespace hyped::motorcontrollers {
```
mifrandir, line 3: i don't like the namespace name

motors?

lib/motorcontrollers/controller.cpp
```diff
@@ -3,5 +3,73 @@
...
 Controller::Controller()
 {
+void processEmergencyMessage(core::types::Canframe data) {
+    //TODO: Implement
 }
-}  // namespace hyped::motorcontrollers
\ No newline at end of file
+
+  void processErrorMessage(const uint16_t error_code) {
+    switch (error_code)
```
mifrandir, line 11: this switch is missing a default case

mifrandir (changes requested):
please make this compile asap and add a header.

kshxtij:
Another more in general thing, we wouldn't want to be logging things like "ERROR_DRV_HS_L1" or "ERROR_DRV_HS_L2". 

Log messages should be clear enough that we immediately know what's happening instead of needing to check the datasheet/some other manual

mifrandir:
I would like the error to stay in the log message, though. This makes it easier to understand whats going on with respect to the manual.

lib/motorcontrollers/controller.cpp
TomLonergan03: Isn't this an issue for a different pr? And is motors actually better?
I chose motorcontrollers as it:
a) directly attaches the namespace to us, as all the other software teams do, 
and b) because the code is for interacting with the controllers, not directly the motors. Though I recognise this point is kinda debatable at best

mifrandir: I don't like `motorcontrollers` because it's (a) too long. That's it. Also, of course the job of the software is to control the motors (just like sensors controls the sensors). This involves talking to the dedicated controllers as well as things like the STN board (which I suppose is a controller in some sense?)

I agree that this is the wrong PR, it's just the first time I see it.

mifrandir: #16

lib/motorcontrollers/controller.cpp
```diff
@@ -0,0 +1,130 @@
...
+namespace hyped::motorcontrollers {
+
+  
+
+Controller::Controller(core::ILogger &logger) : logger_(logger))
+{
+  //TODO: Implement Enum for warning codes
+  // Implement function to iterate through warning codes given a warning bit selection
+  // and output a vector of enums
+  enum warnings {ControllerTemperatureExceeded, MotorTemperatureExceeded, DCLinkUnderVoltage,
```
mifrandir, line 13: I would consider enums inside functions to be bad practice.

mifrandir:
The most recent commit does a lot more than just merging...

lib/motorcontrollers/controller.cpp
TristCrocker: I will get that changed asap

TristCrocker:
Going to recommit with a more explicit explanation on the changes made

lib/motors/controller.cpp
```diff
@@ -1,9 +1,132 @@
...
+      break;
+  }
 }
 
+controller_status Controller::processWarningMessage(const std::uint8_t warning_code)
+{
+  controller_status curr = controller_status::Nominal;
+
+  // Flag specific warnings in binary
+  if ((warning_code & 0b0000'0001) == 0b0000'0001) {
```
TomLonergan03, line 84: this should have the trailing 0s for consistency imo

lib/motors/controller.cpp
```diff
@@ -1,9 +1,132 @@
...
+      break;
+    default:
+      logger_.log(core::LogLevel::kFatal, "GENERIC_ERROR: Unspecific error occurred");
+      break;
+  }
 }
 
+controller_status Controller::processWarningMessage(const std::uint8_t warning_code)
+{
+  controller_status curr = controller_status::Nominal;
```
TomLonergan03, line 81: `curr` is a bad name, something like `current_status` maybe? not really happy with that either but don't have a better idea

lib/motors/controller.hpp
```diff
@@ -1,10 +1,19 @@
 #pragma once
+#include <cstdint>
+
+#include "core/logger.hpp"
 
 namespace hyped::motors {
+enum class controller_status { ControllerTemperatureExceeded, GeneralWarning, Nominal };
```
TomLonergan03, line 7: ControllerStatus
variables are lowercase underscores, classes and enums are upper camel case

TomLonergan03 (changes requested):
Looks pretty good, mostly style issues and spelling to be corrected

lib/motors/controller.hpp
TomLonergan03: Probably call it `kUrecoverableWarning`, or similar to make it clear what it is

lib/motors/controller.cpp
syntax: maybe current_error_priority/current_error_weight/current_error_status ?

TomLonergan03: `priority_error` seems decent

lib/motors/controller.cpp
```diff
@@ -1,9 +1,132 @@
...
+      break;
+    case 0xFF0C:
+      logger_.log(core::LogLevel::kFatal,
+                  "ERROR_DC_LINK_UNDERVOLTAGE: DC voltage not applied to bridge or to low");
+      break;
+    case 0xFF0D:
+      logger_.log(core::LogLevel::kFatal, "ERROR_PULS_MODE_FINISHED: Puls mode finished");
+      break;
+    case 0xFF0E:
+      logger_.log(core::LogLevel::kFatal, "ERROR_APP_ERROR");
```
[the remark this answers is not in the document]
syntax: Unfortunately after cross-referencing all manuals, no.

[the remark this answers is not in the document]
TomLonergan03: that's annoying, i guess this will just have to stay as-is then

TomLonergan03 (approved):
I'm happy with this

TomLonergan03:
Just as a note, "some changes implemented" is a very unhelpful commit message

lib/motors/controller.hpp
```diff
@@ -1,10 +1,19 @@
 #pragma once
+#include <cstdint>
+
+#include "core/logger.hpp"
 
 namespace hyped::motors {
+enum class controllerStatus { kControllerTemperatureExceeded, kUnrecoverableWarning, kNominal };
```
mifrandir, line 7: Naming. This should be `ControllerStatus`.

lib/motors/controller.cpp
```diff
@@ -2,8 +2,128 @@
...
+    logger_.log(core::LogLevel::kFatal, "I2T protection");
+    priority_error = controllerStatus::kUnrecoverableWarning;
+  }
+  if ((warning_code & 0b0100'0000'0000) == 0b0100'0000'0000) {
+    logger_.log(core::LogLevel::kFatal, "Field weakening active");
+    priority_error = controllerStatus::kUnrecoverableWarning;
+  }
+
+  return priority_error;
+}
```
mifrandir, line 128: So this function needs some work...

1. I'm not a massive fan of the mutable `priority_error` here but I see why we would do it like this (just for logging); how about this: catch the errors in decreasing order of priority (reverse from what you're doing here), log the most important error (i.e. in english) AND the full error code (this will allow us to debug other errors/warnings that may have occurred). 
2. Don't use binary; hex does everything you need.
3. `if ((warning_code & MASK) == MASK) { /* ... */ }` is equivalent to `if (warning_code & MASK) { /* ... */  }`

Pytting these together, here's what I think this should look like:

```cpp
if (warning_code & 0x40) {
  logger_.log(core::LogLevel::kFatal, "Controller warning: Field weakening active (code: %x)", warning_code);
  return ControllerStatus::kUnrecoverableWarning;
}
if (warning_code & 0x20) {
  // ...
}
// ...
return ControllerStatus::kNominal;
```

mifrandir (changes requested):

lib/motors/controller.cpp
syntax: I agree this is cleaner, will get this implemented.

syntax: did not consider this, perhaps could roll back processWarningMessage to last commit and just change the bitwise conditions to be neater. I suppose a more readable output is probable preferable.

lib/motors/controller.cpp
```diff
@@ -2,8 +2,142 @@
...
+    return ControllerStatus::kUnrecoverableWarning;
+  }
+  if (warning_code & 0x8) {
+    logger_.log(
+      core::LogLevel::kFatal, "Contoller Warning: DC link over voltage (code: %x)", warning_code);
+    return ControllerStatus::kUnrecoverableWarning;
+  }
+  if (warning_code & 0x10) {
+    logger_.log(core::LogLevel::kFatal,
+                "Contoller Warning: DC over priority_errorent (code: %x)",
```
TomLonergan03, line 98: This should be over current

lib/motors/controller.cpp
```diff
@@ -2,8 +2,142 @@
...
+  }
+  if (warning_code & 0x200) {
+    logger_.log(
+      core::LogLevel::kFatal, "Contoller Warning: I2T protection (code: %x)", warning_code);
+    return ControllerStatus::kUnrecoverableWarning;
+  }
+  if (warning_code & 0x400) {
+    logger_.log(
+      core::LogLevel::kFatal, "Contoller Warning: Field weakening active (code: %x)", warning_code);
+    return ControllerStatus::kUnrecoverableWarning;
```
TomLonergan03, line 132: And I am not at all keen on having to look up error codes in the manual to understand our own logs in every case where there is more than 1 warning active

TomLonergan03 (changes requested):

lib/motors/controller.cpp
```diff
@@ -2,8 +2,142 @@
...
+  }
+  if (warning_code & 0x80) {
+    logger_.log(
+      core::LogLevel::kFatal, "Contoller Warning: BMS Proposed Power (code: %x)", warning_code);
+    return ControllerStatus::kUnrecoverableWarning;
+  }
+  if (warning_code & 0x100) {
+    logger_.log(core::LogLevel::kFatal,
+                "Contoller Warning: Capacitor temporature exceeded (code: %x)",
```
TomLonergan03, line 120: I'm not sure what capacitors this is talking about but possibly they are also something we could increase cooling to deal with

TomLonergan03: Also temperature is spelled wrong

syntax: Do we want this to be made into another ControllerStatus enum? or shall we just leave as fatal

TomLonergan03: and i just noticed controller is spelled wrong in every warning message log

lib/motors/controller.cpp
mifrandir: > My problem with this is that we then have to look it up in the manual in the case of 2 warnings. We should log all warnings, especially fatal ones

Hm, you're kind of right? There is a problem here, though: Logging is technically an I/O operation so I don't really want all of this work to happen here. Now, this is a problem for telemetry but I'm a bit confused about what they've been up to this year because there literally isn't anything in this repository yet. @kshxtij ?

kshxtij: They're working on the UI end right now because they don't need pod-side code beyond stuff that plugs directly into ros.

syntax: Cool, Tristan just rolled back the function to the mutable variable version, think this is the best outcome

mifrandir: Why does the UI not live in this repo? 

Also, this is exactly what scares me! We have a failure and what do we do? We decide to send stupid things over the network.

lib/motors/controller.cpp
```diff
@@ -76,68 +76,61 @@ void Controller::processErrorMessage(const std::uint16_t error_code)
...
 {
-  // Flag specific warnings in binary
+  ControllerStatus priority_error = ControllerStatus::kNominal;
+
+  // In the event any warning are found, print entire error code.
+  if (warning_code != 0) {
+    logger_.log(core::LogLevel::kInfo, "Controller Error found, (code: %x)", warning_code);
+  } else {
+    return priority_error;
+  }
```
TomLonergan03, line 86: better to do if warning == 0 return
then log error code after the if statement finishes

TomLonergan03 (changes requested):

lib/motors/controller.cpp
```diff
@@ -2,8 +2,134 @@
...
+      logger_.log(core::LogLevel::kFatal, "ERROR_APP_ERROR");
+      break;
+    case 0xFF0F:
+      logger_.log(core::LogLevel::kFatal, "ERROR_EMERGENCY_BUTTON: Emergency button pressed");
+      break;
+    case 0xFF10:
+      logger_.log(core::LogLevel::kFatal,
+                  "ERROR_CONTROLLER_OVERTEMPERATURE: Controller overtemperature");
+      break;
+    case 0x3210:
```
mifrandir, line 65: Is this correct? Looks very different to all the other values.

lib/motors/controller.cpp
```diff
@@ -2,8 +2,134 @@
...
+  }
+  if (warning_code & 0x200) {
+    logger_.log(core::LogLevel::kFatal, "Controller Warning: I2T protection");
+    priority_error = ControllerStatus::kUnrecoverableWarning;
+  }
+  if (warning_code & 0x400) {
+    logger_.log(core::LogLevel::kFatal, "Controller Warning: Field weakening active");
+    priority_error = ControllerStatus::kUnrecoverableWarning;
+  }
+
```
mifrandir, line 132: whitespace

lib/motors/controller.cpp
```diff
@@ -2,8 +2,134 @@
...
+}
+
+ControllerStatus Controller::processWarningMessage(const std::uint8_t warning_code)
+{
+  ControllerStatus priority_error = ControllerStatus::kNominal;
+
+  // In the event norminal warning found, return.
+  if (warning_code == 0) { return priority_error; }
+
+  // Log general statement incl. error code.
```
mifrandir, line 84: useless comment, whitespace

mifrandir (changes requested):

mifrandir:
Seems like you added `_deps/googletest-src` by accident. please remove again.

lib/motors/controller.cpp
syntax: image
yep!

mifrandir (approved):
it's fine now

lib/motors/controller.cpp
```diff
@@ -2,8 +2,131 @@
...
+  }
+}
+
+ControllerStatus Controller::processWarningMessage(const std::uint8_t warning_code)
+{
+  ControllerStatus priority_error = ControllerStatus::kNominal;
+
+  // In the event norminal warning found, return.
+  if (warning_code == 0) { return priority_error; }
+  logger_.log(core::LogLevel::kInfo, "Controller Error found, (code: %x)", warning_code);
```
TomLonergan03, line 83: This should be controller warning found

TomLonergan03 (changes requested):
Just the one change, then it should be good

TomLonergan03 (approved):
I'm happy with this

kshxtij (approved):

PR #20, opened 2022-10-27
Started basic outline of navgation system

maxguy2001:
Making outline of navigation system.

Still TODO:
- Setup calibrating
- Figure out data flow
- Decide on how/if we use historic data
- Setup main navigation class (what to use when we want to do navigation)
- Define custom nav types
- Decide where to store navigation constants (track length, number of encoders, etc)


lib/navigation/consts.hpp
```diff
@@ -0,0 +1,46 @@
+#include <cstdint>
+
+
+namespace hyped::navigation{
```
mifrandir, line 4: formatting

lib/navigation/consts.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+  using nav_t = float;
+  static constexpr nav_t kTrackLength       = 100.0; //m
+  static constexpr nav_t kBrakingDistance   = 20.0;  //m
+  static constexpr uint8_t kNumberImus      = 4;
+  static constexpr uint8_t kNumberEncoders  = 4;
+  static constexpr uint8_t kNumberKeyence   = 2;
```
mifrandir, line 11: I think the convention is `kNumKeyence` now?

lib/navigation/consts.hpp
```diff
@@ -0,0 +1,46 @@
...
+  
+  //current trajectory struct
+  struct Trajectory {
+  nav_t displacement;
+  nav_t velocity;
+  nav_t acceleration;
+};
+
+//inline trajectory equal to oprator for use in testing
+inline bool operator==(const Trajectory &lhs, const Trajectory &rhs)
```
mifrandir, line 24: Maybe we should move these closer to the tests? There's a problem here because comparing floats with equality doesn't make much sense. I'm not sure the tests are always going to work anyways when we do it.

lib/navigation/main.cpp
```diff
@@ -0,0 +1,26 @@
...
+    - if kcalibrating, run nav calibrations
+    - if kready, do nothing
+    - if kstart, navigate
+    - if krunning, navigate
+    - if basically anything else, naviagte!
+    */
+  }
+
+
+
```
mifrandir, line 25: white space

lib/navigation/main.cpp
```diff
@@ -0,0 +1,26 @@
...
+//TODO: move/ rename/ make this less bad
+
+namespace hyped::navigation{
+
+  Main::Main(){
+    //TODO: this
+  }
+
+  void Main::navigate(){
```
mifrandir, line 11: I don't think we need a main at all?

This is supposed to be a library, i.e. it needs to have all the nontrivial functionality that is not related to ROS.

In `src/` there will then be a node file which instantiates a particular navigator and does navigation.

lib/navigation/navigator.hpp
```diff
@@ -1,34 +1,19 @@
...
-  core::Float velocity;
-  core::Float acceleration;
-};
+  public:
+
+  Navigator();
+
+  void navigate();
+
+  Trajectory trajectory_;
```
mifrandir, line 13: This shouldn't be public?

mifrandir (changes requested):

lib/navigation/main.cpp
kshxtij: +1

lib/navigation/consts.hpp
kshxtij: gtest has `ASSERT_FLOAT_EQ` so if this is just for testing, rewrite this within the test/navigation using the asserts

lib/navigation/consts.hpp
maxguy2001: yeah formatting is on the to-do list. Will make sure it's done before making this pr ready for review

lib/navigation/main.cpp
maxguy2001: Yeah realised this at the end of the meeting. main.[ch]pp will be leaving us shortly

lib/navigation/navigator.hpp
maxguy2001: Wasn't sure what format data would be published by the navigation system over ROS so left it public until it was decided. Do you think nav should publish trajectory elements separately instead?

mifrandir: This is not a question about ROS. If you want to expose it, do it in a function. Otherwise there is external write access to this class member!

lib/navigation/consts.hpp
mifrandir: Pro-tip: Run `$ ./scripts/setup` to get formatting on commits.

lib/navigation/navigator.hpp
maxguy2001: Fair point...

lib/navigation/crosscheck.hpp
```diff
@@ -0,0 +1,22 @@
+#include <array>
+
+namespace hyped::navigation{
+
+  class Crosscheck{
```
mifrandir, line 5: formatting

mifrandir: also, what's the purpose of this? maybe add a doc comment? in particular, comment in the header what the class is supposed to achieve and comment on the functions in the source file how it is achieve, if that's not obvious from reading the code (example of obvious method: linear regression; example of non-obvious method: Kalman filter)

lib/navigation/crosscheck.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+    public:
+
+      Crosscheck();
```
mifrandir, line 9: None of the functions here take any arguments. How does the crosschecker (whatever it is) obtain data?

mifrandir (changes requested):

lib/navigation/crosscheck.hpp
maxguy2001: Purpose of this is to check the sensor readings against each other (i.e. does the double integrated value of acceleration agree with the value form the wheel encoders). This will be called in the main navigation function between the preprocessing step and updating the current trajectory. Ngl the name might be a bit misleading but couldn't think of anything better so open to suggestions.

lib/navigation/crosscheck.hpp
maxguy2001: TBD. From what I currently understand from @kshxtij, the data will be brought in through the use of callback functions which do not yet exist. Currently undecided as to how to deal with data once we have fresh readings somewhere in central navigation class. Might just pass directly to other functionalities or store as a class member and define some function to access it or maybe some other way.

maxguy2001: Not looking to have that sorted right now though. More just looking to get an outline done so that we can start splitting tasks between nav team and start implementing.

kshxtij: Even when data is being passed in through callbacks, you need functions within your class that accept `std::function<something>` to actually pass in functions.

We can discuss this more tonight.

lib/navigation/navigator.cpp
```diff
@@ -0,0 +1,50 @@
...
+  - calculate current trajecotory
+  - update current trajectory
+  */
+}
+
+void Navigator::publishTrajectory(){
+  /*
+  TODO: implement
+
+  idk how ros works but make sure everyone who needs to know trajectory has it
```
kshxtij, line 27: This would need a callback function passed to it with the type `std::function<void(hyped::navigation::Trajectory)>` and you would call that function with the current trajectory that you wanna send out

kshxtij (changes requested):

lib/navigation/consts.hpp
```diff
@@ -0,0 +1,47 @@
...
+ return !(lhs == rhs);
+}
+
+//explicitly define zero trajectory
+inline Trajectory zero_trajectory = {0, 0, 0};
+
+
+
+template<typename SensorData>
+class INavigator {
```
mifrandir, line 42: Do we still need this?

If so, your `Navigator` should implement `INavigator<WheelEncoderData>`, `INavigator<KeyenceData>`, `INavigator<ImuData>`.

The reason this was here to begin with was so that we could compare navigators in a simulated environment. I.e. we can just swap them out and test them against some data. I still believe that would be a decent idea, but it doesn't seem unfixable later on. Just make sure that your interface is intuitive (input the things that you can handle nicely internally and output the things that you produce directly).

mifrandir (changes requested):

lib/navigation/consts.hpp
```diff
@@ -0,0 +1,47 @@
...
+  static constexpr nav_t kBrakingDistance   = 20.0;  //m
+  static constexpr uint8_t kNumImus         = 4;
+  static constexpr uint8_t kNumEncoders     = 4;
+  static constexpr uint8_t kNumKeyence      = 2;
+  static constexpr nav_t kPi                = 3.14159265359;
+  static constexpr nav_t kWheelCicumference = kPi * 0.1; //m //TODO: check!
+  static constexpr nav_t kStripeDistance    = 10.0; //m
+  
+  //current trajectory struct
+  struct Trajectory {
```
mifrandir, line 16: This probably wants to live somewhere else? E.g. state machine and telemetry will want to use this type and they don't want to link against navigation.

lib/navigation/navigator.cpp
```diff
@@ -28,4 +28,23 @@ void Navigator::publishTrajectory(){
...
 
+void Navigator::setImuData(const core::ImuData imu_data){
+  imu_data_.at(0) = static_cast<nav_t>(imu_data.imu0);
+  imu_data_.at(1) = static_cast<nav_t>(imu_data.imu1);
+  imu_data_.at(2) = static_cast<nav_t>(imu_data.imu2);
+  imu_data_.at(3) = static_cast<nav_t>(imu_data.imu3);
+}
+
+void Navigator::setEncoderData(const core::WheelEncoderData encoder_data){
+  encoder_data_.at(0) = encoder_data.encoder0;
```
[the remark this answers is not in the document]
kshxtij: +1, just pass in the array then if that's what you need

lib/navigation/navigator.cpp
kshxtij: Yeah that makes sense, could be a function in the node that calls `getTrajectory` and publishes it at `x` frequency

lib/navigation/consts.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+  using nav_t = float;
+  static constexpr nav_t kTrackLength       = 100.0; //m
+  static constexpr nav_t kBrakingDistance   = 20.0;  //m
+  static constexpr uint8_t kNumImus         = 4;
+  static constexpr uint8_t kNumEncoders     = 4;
+  static constexpr uint8_t kNumKeyence      = 2;
+  static constexpr nav_t kPi                = 3.14159265359;
+  static constexpr nav_t kWheelCicumference = kPi * 0.1; //m //TODO: check!
+  static constexpr nav_t kStripeDistance    = 10.0; //m
```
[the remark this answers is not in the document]
kshxtij: Think that's a good idea but we'd need to fix all the `// TODO implements` already in the codebase

lib/navigation/consts.hpp
```diff
@@ -0,0 +1,47 @@
+#include <cstdint>
+
+namespace hyped::navigation{
+
+  using nav_t = float;
```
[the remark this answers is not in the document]
maxguy2001: Could do but also quite like having a separate type for navigation tho...

lib/navigation/crosscheck.cpp
```diff
@@ -0,0 +1,49 @@
+#include "crosscheck.hpp"
+
+namespace hyped::navigation{
+
+  Crosscheck::Crosscheck(){
+    //TODO: implement
+  }
+
+  bool Crosscheck::checkTrajectoryAgreement(const std::array<nav_t, kNumImus> imu_data,
```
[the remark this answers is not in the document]
maxguy2001: No I you're right for preprocessing! I need to make sure we deal with 3d imu entries. Will have to add functionality to get resultant and subtract calibration readings. Also will probably implement outlier detection on all 3 axis.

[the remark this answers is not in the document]
maxguy2001: But by the time we get to crosscheck, we'll be dealing with resultant accelerations. Will fix types tho. Since we will want to iterate over the sensors at some point, a struct is probably not the best solution?

lib/navigation/crosscheck.hpp
maxguy2001: Yeah not really much point in it if all we do is pass in the integrated estimates from each sensor and just make a function to compare them, surely this could be done in about 3 lines in the navigator class without having to push a load of data around? Still working on how much/ how/ where to store historic data but it would probably be in here and we would use this class to call integrations/ complete all conversions and then compare.

maxguy2001: The point of this is to keep the messy background work that is needed to compare the sensor's estimates away from the neat navigator class so that all we need to do is call a function from here, pass over the current data and let it do it's job. Idk if there's a nicer way of doing things tho?

lib/navigation/crosscheck.hpp
```diff
@@ -28,17 +31,8 @@ namespace hyped::navigation{
...
-      /**
-       * @brief Checks the integrated camera value of displacement against the 
-       * encoder value of displacement to some tolerance
-       * TODO: update comment with tolerance once updated
-       * 
-       * @return true Camera and wheel encoders agree
-       * @return false Camera and wheel encoders disagree
-       */
-      bool checkEncoderCamera();
+      bool checkEncoderImu(const std::array<nav_t, kNumImus> imu_data,
```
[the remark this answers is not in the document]
maxguy2001: sure. Just `kAgree` and `kDisagree` or similar?

lib/navigation/navigator.cpp
maxguy2001: Yeah that was just a placeholder for something that may end up happening. Maybe better to delete it for now since how the nav data will be accessed is still TBD and we may never need it?

lib/navigation/navigator.cpp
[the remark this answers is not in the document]
maxguy2001: Just make an encoder array in types.hpp, something along lines of :
`using encoder_array = std::array<int32_t, kNumEncoders>;`?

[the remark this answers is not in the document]
mifrandir: Modulo style guide and naming this seems fair.

lib/navigation/crosscheck.hpp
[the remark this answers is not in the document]
mifrandir: Yeah. Though the semantic here are important. If this succeeds it doesn't mean the values agree but rather that they dont disagree too much. Food for thought.

lib/navigation/crosscheck.hpp
mifrandir: Sounds like it would require some experimentation. I will say though: A class should have exactly one job. If that job is doing numerical calculations (integration, differentiation, ...), storing data, and comparing it then something has gone wrong.

Classes aren't expensive so feel free to make more of them. If some of them turn out to be useless we'll clean them up.

lib/navigation/consts.hpp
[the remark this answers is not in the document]
mifrandir: Why? Navigation is the only place that needs floating point numbers, really, so `core::Float` is your type. However, we don't want to force everyone to link against navigation to get a type alias.

lib/navigation/crosscheck.cpp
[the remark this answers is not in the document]
mifrandir: I don't have a definitive answer but here's how you may want to decide which one to go with for now (switching later down the line shouldn't be hard): Which type of access do you expect to see more? Iteration or point wise (e.g. imu_data.first or imu_data.at(0))? I would argue that the only thing you care about with the IMU array is that the order doesn't change; otherwise you are indifferent about which IMU goes where. So an array should be fine, as long as this guarantee is understood. 

One way you may want to fix this is by giving each `ImuData` an explicit ID (maybe talk to @SnickeyX to see whether there is an obvious choice, e.g. an address for this). This would allow you to verify that the data is indeed coming from the place that you thought.

lib/navigation/crosscheck.hpp
maxguy2001: The integration itself would probably get put in utils or something and then just used within this class. The singular purpose of this class is to check that we still know what's going on. Will tidy up data a bit for now and then leave this until the point of implementation since we know what it should do and can volt on anything extra we need at that point in time?

lib/navigation/consts.hpp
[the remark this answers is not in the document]
maxguy2001: Fair enough. Will change to the globally accessible type

lib/navigation/consts.hpp
maxguy2001: This is used in tests contained in several different subdirectories so on second thoughts, probably neater to leave this where it is?

lib/navigation/crosscheck.cpp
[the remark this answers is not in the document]
maxguy2001: have changed to arrays in core types

lib/navigation/navigator.cpp
```diff
@@ -20,31 +20,23 @@ void Navigator::navigate(){
...
+  }
 }
 
-void Navigator::setEncoderData(const core::WheelEncoderData encoder_data){
-  encoder_data_.at(0) = encoder_data.encoder0;
-  encoder_data_.at(1) = encoder_data.encoder1;
-  encoder_data_.at(2) = encoder_data.encoder2;
-  encoder_data_.at(3) = encoder_data.encoder3;
+void Navigator::setEncoderData(const core::EncoderData encoder_data){
+  for(size_t i = 0; i < core::kNumEncoders; ++i){
```
mifrandir, line 31: `std::copy(encoder_data.begin(), encoder_data.end(), raw_encoder_data.begin());`

lib/core/types.hpp
```diff
@@ -16,23 +16,23 @@ struct CanFrame {
...
+static constexpr uint8_t kNumKeyence   = 2;
 
-struct KeyenceData{
-  uint32_t keyence0;
-  uint32_t keyence1;
-};
+//data format for raw sensor data
+using RawImuData  = std::array<std::array<Float, kNumImuAxis>, kNumImus>; 
+using ImuData     = std::array<Float, kNumImus>;
+using EncoderData = std::array<int, kNumEncoders>;
```
mifrandir, line 35: `uint32_t`?

lib/navigation/consts.hpp
```diff
@@ -1,47 +1,36 @@
...
-         && lhs.acceleration == rhs.acceleration;
-}
+  static constexpr core::Float kTrackLength       = 100.0; //m
+  static constexpr core::Float kBrakingDistance   = 20.0;  //m TODO:check!
+  static constexpr core::Float kPi                = 3.14159265359;
+  static constexpr core::Float kWheelCicumference = kPi * 0.1; //m TODO: check!
+  static constexpr core::Float kStripeDistance    = 10.0; //m TODO:check!
 
+  //define sensor checks return type
+  enum class SensorChecks { kUnacceptable = 0, kAcceptable};
```
mifrandir, line 14: [hehe](https://i.pinimg.com/originals/80/69/e0/8069e046bfeeeaa71c8da6f2279517a0.jpg)

lib/navigation/navigator.cpp
mifrandir: Actually, since they have the same type just `raw_encoder_data_ = encoder_data` is good enough; the above is for cases where you may be going from `std::array` to `std::vector` or similar.

You may have to drop the `const` here, which is fine; or add a reference.

mifrandir (changes requested):
please fix your formatting for god's sake

lib/core/types.hpp
```diff
@@ -32,7 +32,7 @@ static constexpr uint8_t kNumKeyence   = 2;
 //data format for raw sensor data
 using RawImuData  = std::array<std::array<Float, kNumImuAxis>, kNumImus>;
```
mifrandir, line 33: `kNumAxes`?

lib/navigation/preprocessing.hpp
```diff
@@ -1,83 +1,82 @@
...
+ private:
+  // raw sensor data
+  core::RawImuData raw_imu_data_;
+  core::EncoderData raw_encoder_data_;
+  core::KeyenceData raw_keyence_data_;
+
+  // processed sensor data
+  core::ImuData imu_data_;
+  core::EncoderData encoder_data_;
+  core::KeyenceData keyence_data_;
```
mifrandir, line 43: I'm unsure why we need to keep all this data here? The preprocessing seems to only rely on the input and the known outliers.

Secondly, I think this class is now doing three things, not one:
1. Outlier detection + flattening on IMU data
2. Outlier detection on encoders.
3. Preprocessing on Keyence (unsure what you would even be doing here?)

As far as I can tell these three are entirely unrelated at this stage, so why do they need to be in the same class?

lib/navigation/preprocessing.hpp
```diff
@@ -1,83 +1,82 @@
...
+   * TODO: update comment with what we do with data once implemented
+   */
+  void checkEncodersReliable();
+
+  // arrays denoting which sensors are reliable
+  std::array<bool, core::kNumImus> reliable_imus_;
+  std::array<bool, core::kNumEncoders> reliable_encoders_;
+
+  // arrays to keep track of how many times a sensor has been an outlier consecutively
+  std::array<int, core::kNumImus> outlier_imus_         = {0, 0, 0, 0};
```
mifrandir, line 76: This initialisation should happen in the constructor, not here.

lib/navigation/preprocessing.hpp
```diff
@@ -1,83 +1,82 @@
...
+  std::array<int, core::kNumEncoders> oultier_encoders_ = {0, 0, 0, 0};
+
+  // flag for if keyence sensors are in disagreement
+  bool keyence_disagreement_;
```
mifrandir, line 80: Flags should start with a verb. E.g. `is_faulty_` or `has_keyence_disagreement_`.

lib/navigation/preprocessing.cpp
```diff
@@ -1,123 +1,129 @@
...
+  -if outlier_imus_[i] > n (tbd), mark imu as unreliable
+  in reliable imus.
+  -if number unreliable in reliable_imus > 1, fail state
+
+  -also figure out return type/ what we update and update
+  documentation as appropriate
+  */
+}
+
+void Preprocessing::checkEncodersReliable()
```
mifrandir, line 60: You're working with state too much. This should be a function taken some raw encoder data and returning whether they the data points agree or not. Possibly with past data? (E.g. encoders should never decrease so that's something that would rely on past state)

However, what is the caller going to do with this information? Just knowing that the encoders are faulty doesn't tell me how to fix it. Are we going to throw away the whole reading? Or are we going to compensate somehow?

I would argue that the transformation in `preprocessEncoders` should go from `std::array<uint32_t, kNumEncoders>` to `std::vector<uint32_t>` where we discard all the values that we deem safe (possibly none!). I don't think late stages will care where each reading came from, only how many we have.

mifrandir (changes requested):
I think you should:

1. Separate preprocessing into three individual classes.
2. Remove unnecessary state in those classes.
3. Rethink your interfaces, similar to what I've described for the encoders.

mifrandir (changes requested):
Some more stuff regarding tests.

Also, formatting (some of these are in navigation files that you have touched, `test/navigation/navigation.cpp` in particular):

```
$ clang-format -Werror --dry-run $(find lib src test -type f | grep ".[ch]pp$")  
lib/utils/manual_time.cpp:10:25: error: code should be clang-formatted [-Wclang-format-violations]
ManualTime::ManualTime()
                        ^
lib/utils/manual_time.hpp:9:7: error: code should be clang-formatted [-Wclang-format-violations]
  void                    set_time(const core::TimePoint time_point);
      ^
lib/core/logger.cpp:17:23: error: code should be clang-formatted [-Wclang-format-violations]
  const auto timestamp            = std::chrono::system_clock::to_time_t(time_point);
                      ^
lib/core/logger.cpp:23:16: error: code should be clang-formatted [-Wclang-format-violations]
  fprintf(file, "%02d:%02d:%02d.%03lld %s[%s] ", time_struct->tm_hour, time_struct->tm_min,
               ^
lib/core/logger.cpp:23:49: error: code should be clang-formatted [-Wclang-format-violations]
  fprintf(file, "%02d:%02d:%02d.%03lld %s[%s] ", time_struct->tm_hour, time_struct->tm_min,
                                                ^
lib/core/logger.cpp:23:71: error: code should be clang-formatted [-Wclang-format-violations]
  fprintf(file, "%02d:%02d:%02d.%03lld %s[%s] ", time_struct->tm_hour, time_struct->tm_min,
                                                                      ^
lib/core/logger.cpp:24:31: error: code should be clang-formatted [-Wclang-format-violations]
          time_struct->tm_sec, time_milliseconds, title, label_);
                              ^
lib/core/logger.cpp:24:50: error: code should be clang-formatted [-Wclang-format-violations]
          time_struct->tm_sec, time_milliseconds, title, label_);
                                                 ^
lib/core/logger.cpp:24:57: error: code should be clang-formatted [-Wclang-format-violations]
          time_struct->tm_sec, time_milliseconds, title, label_);
                                                        ^
lib/core/types.hpp:19:3: error: code should be clang-formatted [-Wclang-format-violations]
//current trajectory struct
  ^
lib/core/types.hpp:19:28: error: code should be clang-formatted [-Wclang-format-violations]
//current trajectory struct
                           ^
lib/core/types.hpp:26:3: error: code should be clang-formatted [-Wclang-format-violations]
//number of each type of sensors
  ^
lib/core/types.hpp:27:34: error: code should be clang-formatted [-Wclang-format-violations]
static constexpr uint8_t kNumImus      = 4;
                                 ^
lib/core/types.hpp:28:37: error: code should be clang-formatted [-Wclang-format-violations]
static constexpr uint8_t kNumImuAxis   = 3;
                                    ^
lib/core/types.hpp:29:38: error: code should be clang-formatted [-Wclang-format-violations]
static constexpr uint8_t kNumEncoders  = 4;
                                     ^
lib/core/types.hpp:30:37: error: code should be clang-formatted [-Wclang-format-violations]
static constexpr uint8_t kNumKeyence   = 2;
                                    ^
lib/core/types.hpp:32:3: error: code should be clang-formatted [-Wclang-format-violations]
//data format for raw sensor data
  ^
lib/core/types.hpp:33:74: error: code should be clang-formatted [-Wclang-format-violations]
using RawImuData  = std::array<std::array<Float, kNumImuAxis>, kNumImus>; 
                                                                         ^
lib/core/types.hpp:35:56: error: code should be clang-formatted [-Wclang-format-violations]
using EncoderData = std::array<uint32_t, kNumEncoders>; 
                                                       ^
lib/core/types.hpp:36:55: error: code should be clang-formatted [-Wclang-format-violations]
using KeyenceData = std::array<uint32_t, kNumKeyence>; 
                                                      ^
lib/io/adc.cpp:13:87: error: code should be clang-formatted [-Wclang-format-violations]
  if (file_ < 0) { /* TODO: log that there is a problem reading raw voltage for pin_*/; }
                                                                                      ^
lib/io/adc.cpp:13:88: error: code should be clang-formatted [-Wclang-format-violations]
  if (file_ < 0) { /* TODO: log that there is a problem reading raw voltage for pin_*/; }
                                                                                       ^
lib/io/adc.cpp:33:59: error: code should be clang-formatted [-Wclang-format-violations]
  const auto offset = lseek(file_descriptor, 0, SEEK_SET);        // reset file pointer
                                                          ^
lib/io/adc.cpp:38:23: error: code should be clang-formatted [-Wclang-format-violations]
  char read_buffer[4];                                            // buffer size 4 for fs value
                      ^
lib/io/adc.cpp:39:27: error: code should be clang-formatted [-Wclang-format-violations]
  const int num_bytes_read = read(file_descriptor, &read_buffer,  
                          ^
lib/io/adc.cpp:39:51: error: code should be clang-formatted [-Wclang-format-violations]
  const int num_bytes_read = read(file_descriptor, &read_buffer,  
                                                  ^
lib/io/adc.cpp:39:65: error: code should be clang-formatted [-Wclang-format-violations]
  const int num_bytes_read = read(file_descriptor, &read_buffer,  
                                                                ^
lib/io/adc.cpp:40:56: error: code should be clang-formatted [-Wclang-format-violations]
                                  sizeof(read_buffer));           // actually consume new data, changes value in buffer
                                                       ^
lib/io/adc.cpp:43:25: error: code should be clang-formatted [-Wclang-format-violations]
    return std::nullopt;                                          // returning NULL since we did not get any value
                        ^
lib/io/adc.cpp:46:45: error: code should be clang-formatted [-Wclang-format-violations]
  return static_cast<uint16_t>(raw_voltage);                      // max value is 2^12-1 = 4095
                                            ^
src/pod/main.cpp:1:1: error: code should be clang-formatted [-Wclang-format-violations]
#include <iostream>
^
test/utils/naive_navigator.cpp:1:1: error: code should be clang-formatted [-Wclang-format-violations]
#include <gtest/gtest.h>
^
test/core/logger.cpp:1:1: error: code should be clang-formatted [-Wclang-format-violations]
#include <gtest/gtest.h>
^
test/core/logger.cpp:8:47: error: code should be clang-formatted [-Wclang-format-violations]
void testStdoutLog(const core::LogLevel level, const utils::ManualTime &manual_time, const std::string expected_output)
                                              ^
test/core/logger.cpp:8:85: error: code should be clang-formatted [-Wclang-format-violations]
void testStdoutLog(const core::LogLevel level, const utils::ManualTime &manual_time, const std::string expected_output)
                                                                                    ^
test/core/logger.cpp:16:47: error: code should be clang-formatted [-Wclang-format-violations]
void testStderrLog(const core::LogLevel level, const utils::ManualTime &manual_time, const std::string expected_output)
                                              ^
test/core/logger.cpp:16:85: error: code should be clang-formatted [-Wclang-format-violations]
void testStderrLog(const core::LogLevel level, const utils::ManualTime &manual_time, const std::string expected_output)
                                                                                    ^
test/core/logger.cpp:22:2: error: code should be clang-formatted [-Wclang-format-violations]
} 
 ^
test/core/logger.cpp:30:2: error: code should be clang-formatted [-Wclang-format-violations]
}
 ^
test/navigation/kalman_filter.cpp:1:1: error: code should be clang-formatted [-Wclang-format-violations]
#include <gtest/gtest.h>
^
test/navigation/navigation.cpp:1:1: error: code should be clang-formatted [-Wclang-format-violations]
#include <gtest/gtest.h>
^
test/navigation/navigation.cpp:9:2: error: code should be clang-formatted [-Wclang-format-violations]
{ 
 ^
```

lib/navigation/preprocessing.cpp
maxguy2001: This checks if an individual encoder is consistently faulty. If so, there  is  no fixing, we just disregard any further readings from that sensor. I see what you mean about probably working with state too much but the idea of working with state here is to keep track of which sensors are giving bad data. I'll modify the new classes I'm going to make to be less reliant on states.

test/CMakeLists.txt
```diff
@@ -7,6 +7,6 @@ FetchContent_Declare(
 FetchContent_MakeAvailable(googletest)
 include(GoogleTest)
 
-add_subdirectory(utils)
-add_subdirectory(navigation)
+#add_subdirectory(utils)
```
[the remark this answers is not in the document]
maxguy2001: The inline operator definition was giving me the runaround and I was writing this at 4am so decided that I'd disable it for now and make it a problem for another day. I know about it and will fix before merge.

lib/navigation/navigator.cpp
```diff
@@ -0,0 +1,42 @@
...
+  - preprocess
+  - check sensor agreement
+  - calculate current trajecotory
+  - update current trajectory
+  */
+}
+
+void Navigator::setImuData(const core::RawImuData imu_data)
+{
+  for (size_t i = 0; i < core::kNumImus; ++i) {
```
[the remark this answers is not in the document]
maxguy2001: have added `#inlude <cstdint>`, will that not fix it?

test/navigation/navigation.cpp
```diff
@@ -1,26 +1,27 @@
...
 
 TEST(Trajectory, equality)
-{
-  navigation::Trajectory reference = {100.0, 10.0, 1.0};
+{ 
+  core::Trajectory reference = {100.0, 10.0, 1.0};
   {
-    navigation::Trajectory other = {100.0, 10.0, 1.0};
+    core::Trajectory other = {100.0, 10.0, 1.0};
     ASSERT_EQ(reference, other);
```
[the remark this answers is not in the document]
maxguy2001: Yeah I know why they were breaking but was working on other parts so temporarily disabled them. Will make sure they're all running again before merge.

lib/navigation/preprocessing.hpp
maxguy2001: Fair point. Also, think I'm going to change keyence agreement implementation to a function within navigation since it really only needs to check equality of results and allow one iteration where they disagree I think (can be modified easily if we decide to fail on disagreement). Thinking about it, it makes more sense in general to separate the preprocessing out since we probably won't be receiving new data  at the same point in time from multiple sensors - it will more likely be at different points in time.

lib/navigation/navigator.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+  - look at new data
+  - preprocess
+  - check sensor agreement
+  - calculate current trajecotory
+  - update current trajectory
+  */
+}
+
+void Navigator::setImuData(const core::RawImuData imu_data)
```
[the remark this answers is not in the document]
mifrandir: I agree that this needn't be implemented now, but the interface (including the names) should reflect this as much as possible.

lib/navigation/navigator.cpp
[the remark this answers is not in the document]
mifrandir: Why do that if you can just use `std::size_t`? I prefer this anyways, though I don't think there's a rule.

lib/navigation/preprocessing.cpp
maxguy2001: Just changed implementation. No raw data is at any point stored in preprocessing classes. Data is passed in, processed and output. The only states used are tracking credibility of sensors

lib/io/adc.cpp
```diff
@@ -10,8 +10,8 @@ Adc::Adc(const uint8_t pin, hyped::core::ILogger &logger) : pin_(pin), logger_(l
   char buf[100];
   snprintf(buf, sizeof(buf), "/sys/bus/iio/devices/iio:device0/in_voltage%i_raw", pin_);
   file_ = open(buf, O_RDONLY);
-  if (file_ < 0) {
-    logger_.log(hyped::core::LogLevel::kFatal, "Unable to open ADC file");
+  if (file_ < 0) { /* TODO: log that there is a problem reading raw voltage for pin_*/
```
kshxtij, line 13: ?

lib/navigation/crosscheck.cpp
```diff
@@ -0,0 +1,52 @@
...
+{
+  // TODO: implement
+}
+
+SensorChecks Crosscheck::checkTrajectoryAgreement(const core::ImuData imu_data,
+                                                  const core::EncoderData encoders_data,
+                                                  const core::KeyenceData keyence_data)
+{
+  /*
+  TODO: implement
```
kshxtij, line 15: TODOLater for TODOs that are out of scope from this pr, otherwise the CI will flag it.

lib/navigation/crosscheck.cpp
```diff
@@ -0,0 +1,52 @@
...
+    Also need to figure out how data flow is going to work with the historic data and what we use.
+    The basic infrastrucutre is there for now so will be a problem for another day.
+  */
+  return SensorChecks::kAcceptable;
+}
+
+SensorChecks checkEncoderImu(const core::ImuData imu_data, const core::EncoderData encoders_data)
+{
+  /*
+  TODO: implement.
```
kshxtij, line 30: Same as above

lib/navigation/crosscheck.cpp
```diff
@@ -0,0 +1,52 @@
...
+  - otherwise all good, return true
+  */
+  return SensorChecks::kAcceptable;
+}
+
+SensorChecks checkEncoderKeyence(const core::EncoderData encoder_data,
+                                 const core::KeyenceData keyence_data)
+{
+  /*
+  TODO: implement.
```
kshxtij, line 44: One more

lib/navigation/crosscheck.cpp
kshxtij, line 30: Same as above

lib/navigation/preprocess_imu.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+ private:
+  core::ImuData imuOutlierDetection(const core::RawImuData imu_data);
+
+  void checkImusReliable(const core::ImuData imu_data);
+
+  // initialised as {0, 0, 0, 0}, count of consecutive outliers
+  std::array<uint16_t, core::kNumImus> outlier_imus_;
+
+  // initialised as all true, bool mask of reliable sensors
```
kshxtij, line 21: Is this a todo? otherwise where is it being initialised

test/CMakeLists.txt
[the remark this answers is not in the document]
kshxtij: Can look at this tonight

kshxtij (changes requested):
Looks good, TODOLater for anything that isn't meant to be implemented/fixed within this PR

lib/navigation/crosscheck.cpp
maxguy2001: Do we not want them all flagged? Then we know what still needs done before anything critical is run on this codebase? All my TODO comments will definitely need implemented before we run and isn't this the point of the TODO flag?

lib/navigation/preprocess_imu.hpp
maxguy2001: Will be initialized in constructor which is TODO

lib/io/adc.cpp
maxguy2001: shit. Formatted and got conflict. Will fix now

mifrandir (changes requested):

test/navigation/navigation.cpp
```diff
@@ -5,24 +5,39 @@
...
+}
+
+void checkTrajectoriesEquality(const core::Trajectory &trajectory1,
+                               const core::Trajectory &trajectory2,
+                               bool equality)
+{
+  const bool comparision = is_float_eq(trajectory1.acceleration, trajectory2.acceleration)
+                           && is_float_eq(trajectory1.velocity, trajectory2.velocity)
+                           && is_float_eq(trajectory1.displacement, trajectory2.displacement);
+  ASSERT_EQ(equality, comparision);
```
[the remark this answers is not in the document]
kshxtij: No equivalent `ASSERT_FLOAT_NE` and we can't just have three `ASSERT_NE` calls because if we have 2 trajectories {1,2,3} and {1,3,4} then it wouldn't work because `ASSERT_NE(1, 1)` would fail the test.

This was the easiest way i could think of during the meeting

lib/navigation/navigator.cpp
```diff
@@ -0,0 +1,44 @@
...
+  - calculate current trajecotory
+  - update current trajectory
+  */
+}
+
+void Navigator::recordImuData(const core::RawImuData imu_data)
+{
+  for (std::size_t i = 0; i < core::kNumImus; ++i) {
+    raw_imu_data_.at(i) = imu_data.at(i);
+  }
```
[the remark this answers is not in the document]
maxguy2001: No, this is where data is taken into the navigation class to be passed around to everyone who needs it. Makes sense to wipe this for now tho since how to handle multiple data points at once is still tbd.

lib/navigation/navigator.hpp
```diff
@@ -1,34 +1,54 @@
...
  public:
-  virtual Trajectory currentTrajectory()             = 0;
-  virtual void update(const SensorData &sensor_data) = 0;
+  Navigator();
+
+  /**
+   *@brief Does navigation. Combines all navigation functionality to produce the best
+   *current estimate of trajectory
+   */
+  void navigate();
```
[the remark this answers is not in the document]
maxguy2001: This is where all the main navigation decisions are made and processes are called. The rough plan will be to check what new data is there and act accordingly (preprocess, cross check agreement and update trajectory as appropriate).

test/utils/naive_navigator.cpp
```diff
@@ -1,14 +1,18 @@
 #include <gtest/gtest.h>
 
+#include <core/types.hpp>
 #include <utils/naive_navigator.hpp>
 
 namespace hyped::test {
 
 void test_with_trajectory(utils::NaiveNavigator &naive_navigator,
```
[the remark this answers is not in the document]
maxguy2001: ?

lib/navigation/crosscheck.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * @brief checks that all sensors agree with wheel encoders. This is a safety
+   * check for the reliability of the current trajectory (i.e. the sensors are
+   * probably not all wrong...)
+   *
+   * @return true signifies trajectory agreement
+   * @return false signifies trajectory disagreement. We enter fail state
+   */
+  SensorChecks checkTrajectoryAgreement(const core::ImuData imu_data,
```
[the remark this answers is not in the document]
maxguy2001: Integration functionality of some format will probably be called in here but is still up for debate. Not 100% sure what to pass in here but currently passing in preprocessed data. What would probably make sense at some point is to define a struct for sensor estimates for displacement somewhere and update that with integrated values. However this is further down todo list, really just want basic outline merged for now because this pr is starting to bottleneck nav progress

lib/navigation/navigator.hpp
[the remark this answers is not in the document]
maxguy2001: This implementation can be revisited later tho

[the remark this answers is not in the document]
mifrandir: My problem is this: Why are there decisions to be made if there isn't any new data and nobody is checking any outputs? We are not in threaded-land anymore where we have a loop that runs internally forever so I don't see why a caller would ever call this function.

lib/navigation/crosscheck.hpp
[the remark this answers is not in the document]
mifrandir: You can open PRs into this branch; there's no rush in getting this merged. (The base will change once this is merge so people can start work on the areas that seem stable.)

Implementation is the easy bit and this is the hard bit that you should get right. It's worth taking time to get a good interface done because then the implementation will write itself.

test/navigation/navigation.cpp
```diff
@@ -5,7 +5,7 @@
 
 namespace hyped::test {
 
-bool is_float_eq(float a, float b)
+bool checkFlaotEquality(float a, float b)
```
mifrandir, line 8: :eyes:

lib/navigation/navigator.hpp
```diff
@@ -4,51 +4,41 @@
...
  public:
   Navigator();
 
   /**
-   *@brief Does navigation. Combines all navigation functionality to produce the best
-   *current estimate of trajectory
+   *@brief runs cross checking and returns trajectory
    */
-  void navigate();
+  core::Trajectory currentTrajectory();
```
mifrandir, line 14: I believe this can fail so it should be `std::optional`

test/utils/naive_navigator.cpp
```diff
@@ -1,26 +1,51 @@
...
+  core::Float sum_imu = 0;
+  for (size_t i = 0; i < core::kNumImus; ++i) {
+    for (size_t j = 0; j < core::kNumAxis; ++j) {
+      sum_imu += imu_data.at(i).at(j);
+    }
+  }
+  core::Float new_acceleration = sum_imu / core::kNumImus;
+  core::Float new_velocity     = 0;
+  core::Float new_displacement
+    = std::accumulate(encoder_data.begin(), encoder_data.end(), 0.0) / core::kNumEncoders;
```
mifrandir, line 28: Hm, I don't like this test... it's just the code copied and pasted, it doesn't verify any properties.

lib/utils/naive_navigator.cpp
```diff
@@ -11,9 +13,30 @@ core::Trajectory NaiveNavigator::currentTrajectory()
...
-void NaiveNavigator::update(const core::Trajectory &current_trajectory)
+void NaiveNavigator::keyenceUpdate(const core::KeyenceData &keyence_data)
+{
+  // Do nothing. Keyence has no direct influence on trajectory
+}
+
+void NaiveNavigator::encoderUpdate(const core::EncoderData &encoder_data)
+{
+  core::Float sum                  = std::accumulate(encoder_data.begin(), encoder_data.end(), 0.0);
+  core::Float encoder_average      = sum / core::kNumEncoders;
```
mifrandir, line 24: explicit cast to `core::Float`?

lib/navigation/navigator.hpp
```diff
@@ -4,51 +4,41 @@
...
+  void imuUpdate(const core::RawImuData &imu_data);
 
  private:
-  core::RawImuData raw_imu_data_;
-  core::EncoderData raw_encoder_data_;
-  core::KeyenceData raw_keyence_data_;
-
-  core::ImuData imu_data_;
+  // previous readings
   core::EncoderData encoder_data_;
```
mifrandir, line 37: call them previous or last then

lib/navigation/consts.hpp
```diff
@@ -17,11 +17,11 @@ enum class SensorChecks { kUnacceptable = 0, kAcceptable };
 // explicitly define zero trajectory
 inline core::Trajectory zero_trajectory = {0, 0, 0};
 
-// TODOLater fix this
-template<typename SensorData>
-class INavigator {
+class INavigator : public core::KeyenceData, core::EncoderData, core::RawImuData {
```
mifrandir, line 20: Why are you inheriting from `core::KeyenceData`? Don't do this, just remove.

mifrandir (changes requested):

test/utils/naive_navigator.cpp
maxguy2001: Yeah not quite sure what to do with this since the actual behavior of the class has changed?

lib/navigation/consts.hpp
```diff
@@ -17,9 +18,10 @@ enum class SensorChecks { kUnacceptable = 0, kAcceptable };
 // explicitly define zero trajectory
```
mifrandir, line 18: useless comment

lib/navigation/consts.hpp
```diff
@@ -17,9 +18,10 @@ enum class SensorChecks { kUnacceptable = 0, kAcceptable };
...
 inline core::Trajectory zero_trajectory = {0, 0, 0};
 
-class INavigator : public core::KeyenceData, core::EncoderData, core::RawImuData {
+// Navigator interface
```
mifrandir, line 21: ...

mifrandir (changes requested):

lib/navigation/consts.hpp
maxguy2001: ?

kshxtij: Im guessing pointless comment again 😂

test/utils/naive_navigator.cpp
maxguy2001: Think I'm going to make testing this properly something to add to the TODOLater pile.

lib/navigation/preprocess_imu.hpp
```diff
@@ -0,0 +1,25 @@
+#include <array>
+#include <cstdint>
+
+#include "core/types.hpp"
+
+namespace hyped::navigation {
+class PreprocessImus {
```
mifrandir, line 7: Surely this isn't quite a word?

`ImuPreprocessor`?

lib/navigation/preprocess_imu.hpp
```diff
@@ -0,0 +1,25 @@
...
+ public:
+  PreprocessImus();
+
+  core::ImuData processData(const core::RawImuData raw_imu_data);
+
+ private:
+  core::ImuData imuOutlierDetection(const core::RawImuData imu_data);
+
+  void checkImusReliable(const core::ImuData imu_data);
```
mifrandir, line 16: You're in the IMU class. Just `checkReliable`. Also reference.

lib/navigation/preprocess_imu.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+namespace hyped::navigation {
+class PreprocessImus {
+ public:
+  PreprocessImus();
+
+  core::ImuData processData(const core::RawImuData raw_imu_data);
+
+ private:
+  core::ImuData imuOutlierDetection(const core::RawImuData imu_data);
```
mifrandir, line 14: `detectOutliers`? `filter`?  depending on how this is supposed to work.

lib/navigation/preprocess_imu.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+  void checkImusReliable(const core::ImuData imu_data);
+
+  // initialised as {0, 0, 0, 0}, count of consecutive outliers
+  std::array<uint16_t, core::kNumImus> outlier_imus_;
```
mifrandir, line 19: this doesn't contain IMUs so the name shouldn't suggest that. how about `num_outliers_per_imu_`?

lib/navigation/preprocess_imu.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+  // initialised as all true, bool mask of reliable sensors
+  std::array<bool, core::kNumImus> reliable_imus_;
```
mifrandir, line 22: similar to the above. `is_reliable_`?

lib/navigation/preprocess_keyence.hpp
```diff
@@ -0,0 +1,20 @@
+#include "consts.hpp"
+
+#include <array>
+#include <cstdint>
+
+#include "core/types.hpp"
+
+namespace hyped::navigation {
+
+class PreprocessKeyence {
```
mifrandir, line 10: Similar to the IMU case.

lib/navigation/preprocess_encoders.hpp
```diff
@@ -0,0 +1,25 @@
+#include <array>
+#include <cstdint>
+
+#include "core/types.hpp"
+
+namespace hyped::navigation {
+class PreprocessEncoders {
```
mifrandir, line 7: Similar to the IMU case.

mifrandir (changes requested):
Last round I hope; mostly naming.

mifrandir (approved):

BenET01 (approved):

lib/io/adc.cpp
```diff
@@ -42,6 +42,7 @@ std::optional<uint16_t> Adc::resetAndRead4(const int file_descriptor)
            sizeof(read_buffer));  // actually consume new data, changes value in buffer
   if (num_bytes_read != sizeof(read_buffer)) {
     logger_.log(hyped::core::LogLevel::kFatal, "Failed to read 4 bytes");
+
```
SnickeyX, line 45: ??

lib/navigation/preprocess_encoders.hpp
```diff
@@ -0,0 +1,25 @@
...
+ private:
+  core::EncoderData detectOutliers(const core::EncoderData encoder_data);
+
+  void checkReliable(const core::EncoderData &encoder_data);
+
+  // initialised as {0, 0, 0, 0}, count of consecutive outliers
+  std::array<uint16_t, core::kNumEncoders> num_outliers_per_encoder_;
+
+  // initialised as all true, bool mask of reliable sensors
+  std::array<bool, core::kNumEncoders> is_encoder_reliable_;
```
SnickeyX, line 22: since these are multiple, `are_encoders_reliable_` ?

lib/navigation/preprocess_imu.hpp
```diff
@@ -0,0 +1,25 @@
...
+ private:
+  core::ImuData detectOutliers(const core::RawImuData imu_data);
+
+  void checkReliable(const core::ImuData &imu_data);
+
+  // initialised as {0, 0, 0, 0}, count of consecutive outliers
+  std::array<uint16_t, core::kNumImus> num_outliers_per_imu_;
+
+  // initialised as all true, bool mask of reliable sensors
+  std::array<bool, core::kNumImus> is_imu_reliable_;
```
SnickeyX, line 22: `are_imus_reliable`?

SnickeyX (changes requested):
very small requests, look good overall!

lib/io/adc.cpp
maxguy2001: Was failing formatter...

SnickeyX: oh fair, then dw

lib/navigation/preprocess_encoders.hpp
maxguy2001: those cheeky plurals catching me out again :(

lib/navigation/crosscheck.hpp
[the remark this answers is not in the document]
Fayz-S: @maxguy2001 if this is the case, we could look at implementing now instead of waiting for merge? We'll probably be able to work out some of these kinks as we implement the other more stable parts of the skeleton.

lib/navigation/consts.hpp
```diff
@@ -0,0 +1,27 @@
...
+namespace hyped::navigation {
+
+static constexpr core::Float kTrackLength       = 100.0;  // m
+static constexpr core::Float kBrakingDistance   = 20.0;   // m TODOLater:check!
+static constexpr core::Float kPi                = 3.14159265359;
+static constexpr core::Float kWheelCicumference = kPi * 0.1;  // m TODOLater: check!
+static constexpr core::Float kStripeDistance    = 10.0;       // m TODOLater:check!
+
+// define sensor checks return type
+enum class SensorChecks { kUnacceptable = 0, kAcceptable };
```
Fayz-S, line 16: I'm unfamiliar with this notation: what value does kAcceptable take on?

lib/navigation/crosscheck.hpp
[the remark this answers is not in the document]
Fayz-S: @maxguy2001 if this is the case, we could look at implementing now instead of waiting for merge? We'll probably be able to work out some of these kinks as we implement the other more stable parts of the skeleton.

lib/utils/naive_navigator.cpp
```diff
@@ -1,19 +1,42 @@
...
+void NaiveNavigator::imuUpdate(const core::RawImuData &imu_data)
 {
-  current_trajectory_ = current_trajectory;
+  core::Float sum = 0.0;
+  for (size_t i = 0; i < core::kNumImus; ++i) {
+    for (size_t j = 0; j < core::kNumAxis; ++j) {
+      sum += imu_data.at(i).at(j);
+    }
+  }
+  core::Float imu_average          = static_cast<core::Float>(sum / core::kNumImus);
```
Fayz-S, line 36: White space?

test/navigation/navigation.cpp
```diff
@@ -1,27 +1,44 @@
...
+    core::Trajectory other = {100.0, 20.0, 1.0};
+    checkTrajectoriesEquality(reference, other, false);
   }
   {
-    navigation::Trajectory other = {100.0, 10.0, 2.0};
-    ASSERT_NE(reference, other);
+    core::Trajectory other = {100.0, 10.0, 2.0};
+    checkTrajectoriesEquality(reference, other, false);
   }
 }
```
Fayz-S, line 43: The file ending here with the bracketsdoesn't seem to be consistent with your other files.

kshxtij (approved):

SnickeyX (approved):

test/navigation/navigation.cpp
maxguy2001: Formatting was not done manually. Formatting checks have passed so all good

lib/utils/naive_navigator.cpp
maxguy2001: Auto formatting is happy so we're all good

lib/navigation/consts.hpp
maxguy2001: 1

maxguy2001: (basically like bool but more descriptive)

PR #19, opened 2022-10-27
SNS - PWM

kshxtij:

PR #21, opened 2022-10-27
MTC - socketCAN implementation

TomLonergan03:
Working, tested socketCAN code.

TODO:
- [x] Receiving messages ~~using separate thread? This was how it was last year at least~~ synchronously
- [x] Something to call when messages are received
- [x] Figure out receive buffer


mifrandir:
Do not use threads. You can do the same stuff synchronously. The messages will be put in a buffer somewhere and you can read them the same way the separate threads did previously. Of course we need to be careful not to wait too long, but that was the case with threads, too.

lib/io/can.cpp
```diff
@@ -1,8 +1,73 @@
...
+    return;
+  }
+
+  if (write(socket_, &message, sizeof(can_frame)) != sizeof(can_frame)) {
+    logger_.log(hyped::core::LogLevel::kFatal, "Failed to send CAN message");
+  }
+
+  logger_.log(hyped::core::LogLevel::kDebug,
+              "CAN message sent, ID:%i, DATA: %i %i %i %i %i %i %i %i",
+              (int)message.can_id,
```
mifrandir, line 54: use `static_cast<int>` instead

mifrandir: also this look ugly 

maybe we can add support for printing arrays to the logger? all we have to do is add a conversion from array to args (which is then converted to an array). Maybe a function `vlog` @kshxtij ?

lib/io/can.hpp
```diff
@@ -1,8 +1,15 @@
+#include <core/logger.hpp>
 #include <core/types.hpp>
 
 namespace hyped::io {
 class Can {
  public:
-  Can();
+  Can(hyped::core::ILogger &logger);
+  int sendCanFrame(can_frame message);
```
mifrandir, line 8: Not a good return type. Use an enum or a named struct.

lib/io/can.hpp
```diff
@@ -1,8 +1,15 @@
...
+  can_frame receiveCanFrame();
```
mifrandir, line 9: This is an I/O operation. It can probably go wrong. Use `std::optional` to indicate that.

lib/io/can.cpp
```diff
@@ -1,8 +1,73 @@
...
+}
+
+int Can::sendCanFrame(can_frame message)
+{
+  if (socket_ < 0) {
+    logger_.log(hyped::core::LogLevel::kFatal,
+                "Trying to send CAN message but no CAN socket found");
+    return;
+  }
+
```
mifrandir, line 47: whitespace

lib/io/can.cpp
```diff
@@ -1,8 +1,73 @@
...
+  logger_.log(hyped::core::LogLevel::kInfo, "CAN socket successfully created");
+}
+
+int Can::sendCanFrame(can_frame message)
+{
+  if (socket_ < 0) {
+    logger_.log(hyped::core::LogLevel::kFatal,
+                "Trying to send CAN message but no CAN socket found");
+    return;
```
mifrandir, line 45: undefined behaviour because you don't have a return value. this should be `std::nullopt`

lib/io/can.cpp
```diff
@@ -1,8 +1,73 @@
...
+#include <unistd.h>
+
+#include <linux/can.h>
+#include <net/if.h>
+#include <sys/ioctl.h>
+#include <sys/socket.h>
+
 namespace hyped::io {
-Can::Can()
+Can::Can(hyped::core::ILogger &logger) : logger_(logger)
```
mifrandir, line 12: I think you should extract the I/O out of the constructor into a function, e.g. `Can::initialise`.

The reason is the following: I/O may fail and you want to indicate that to your callee. No, logging doesn't count! Thus have a function that returns an enum value `kFailure` or `kSuccess`. In fact, this type has become so common that we might want to move it into core @kshxtij . But then again, they don't need to be interchangable so maybe not?

lib/io/can.cpp
```diff
@@ -1,8 +1,73 @@
...
+    return;
+  }
+
+  if (bind(socket_, (sockaddr *)&socket_address, sizeof(socket_address)) < 0) {
+    logger_.log(hyped::core::LogLevel::kFatal, "Unable to bind CAN socket");
+    close(socket_);
+    socket_ = -1;
+    return;
+  }
+
```
mifrandir, line 36: whitespace; you get the idea

lib/io/can.cpp
```diff
@@ -1,8 +1,73 @@
...
+  socket_address.can_ifindex = if_nametoindex("can1");
+
+  if (socket_address.can_ifindex == 0) {
+    logger_.log(hyped::core::LogLevel::kFatal, "Unable to find CAN1 network interface");
+    close(socket_);
+    socket_ = -1;
+    return;
+  }
+
+  if (bind(socket_, (sockaddr *)&socket_address, sizeof(socket_address)) < 0) {
```
mifrandir, line 30: there's another C-cast here. use `static_cast` (or similar)

mifrandir (changes requested):

lib/io/can.cpp
kshxtij: Yeah i was thinking of creating a core::types::WriteResult enum with kFailure and kSucess while doing #19 anyways

lib/io/can.cpp
kshxtij: I'll look into it

kshxtij: I think converting an array to va_args/va_lists doesn't seem like a good idea

lib/io/can.cpp
kshxtij: Done in #19, will retroactively fix everything else once that's merged

lib/io/can.cpp
```diff
@@ -1,8 +1,74 @@
...
+
+CanResult Can::initialiseCanSocket()
+{
+  if ((socket_ = socket(PF_CAN, SOCK_RAW, CAN_RAW)) < 0) {
+    logger_.log(hyped::core::LogLevel::kFatal, "Unable to open CAN socket");
+    return hyped::io::CanResult::kFailure;
+  }
+  sockaddr_can socket_address;
+  socket_address.can_family  = AF_CAN;
+  socket_address.can_ifindex = if_nametoindex("can1");
```
kshxtij, line 24: Are we not working with any other can interfaces? If we are you should pass this to the function as an argument

lib/io/can.cpp
```diff
@@ -1,8 +1,74 @@
...
+
 namespace hyped::io {
-Can::Can()
+Can::Can(hyped::core::ILogger &logger) : logger_(logger)
+{
+}
+
+CanResult Can::initialiseCanSocket()
+{
+  if ((socket_ = socket(PF_CAN, SOCK_RAW, CAN_RAW)) < 0) {
```
kshxtij, line 18: Split these into two lines for better readability?

lib/io/can.cpp
```diff
@@ -65,7 +69,8 @@ CanResult Can::sendCanFrame(can_frame message)
 std::optional<can_frame> Can::receiveCanFrame()
 {
   can_frame received_message;
-  if (read(socket_, &received_message, sizeof(can_frame)) < sizeof(can_frame)) {
+  int numBytesRead = read(socket_, &received_message, sizeof(can_frame));
```
kshxtij, line 72: const

lib/io/can.cpp
```diff
@@ -45,7 +48,8 @@ CanResult Can::sendCanFrame(can_frame message)
                 "Trying to send CAN message but no CAN socket found");
     return hyped::io::CanResult::kFailure;
   }
-  if (write(socket_, &message, sizeof(can_frame)) != sizeof(can_frame)) {
+  int numBytesWritten = write(socket_, &message, sizeof(can_frame));
```
kshxtij, line 51: const

lib/io/can.cpp
```diff
@@ -13,22 +13,25 @@ Can::Can(hyped::core::ILogger &logger) : logger_(logger)
...
-  socket_address.can_ifindex = if_nametoindex("can1");
+  socket_address.can_ifindex = if_nametoindex(can_network_interface.c_str());
   if (socket_address.can_ifindex == 0) {
     logger_.log(hyped::core::LogLevel::kFatal, "Unable to find CAN1 network interface");
     close(socket_);
     socket_ = -1;
     return hyped::io::CanResult::kFailure;
   }
-  if (bind(socket_, reinterpret_cast<sockaddr *>(&socket_address), sizeof(socket_address)) < 0) {
+  int bindStatus
```
kshxtij, line 32: const

kshxtij, line 32: bind_status

lib/io/can.cpp
kshxtij, line 51: num_bytes_written

lib/io/can.cpp
kshxtij, line 72: num_bytes_read

kshxtij (changes requested):

lib/io/can.cpp
```diff
@@ -1,8 +1,79 @@
...
+#include <sys/ioctl.h>
+#include <sys/socket.h>
+
 namespace hyped::io {
-Can::Can()
+Can::Can(hyped::core::ILogger &logger) : logger_(logger)
+{
+}
+
+CanResult Can::initialiseCanSocket(std::string can_network_interface)
```
mifrandir, line 16: Just `initialise`?

Re semantics: We don't accommodate for the possibility of being on more than one CAN network, right?

lib/io/can.cpp
```diff
@@ -1,8 +1,79 @@
...
+  if (socket_ < 0) {
+    logger_.log(hyped::core::LogLevel::kFatal,
+                "Trying to send CAN message but no CAN socket found");
+    return hyped::io::CanResult::kFailure;
+  }
+  const int num_bytes_written = write(socket_, &message, sizeof(can_frame));
+  if (num_bytes_written != sizeof(can_frame)) {
+    logger_.log(hyped::core::LogLevel::kFatal, "Failed to send CAN message");
+  }
+  logger_.log(hyped::core::LogLevel::kDebug,
```
mifrandir, line 55: Can we get a `// TODO` here to make this elegant? No need to fix it now.

lib/io/can.cpp
```diff
@@ -1,8 +1,79 @@
...
+    logger_.log(hyped::core::LogLevel::kFatal, "Unable to bind CAN socket");
+    close(socket_);
+    socket_ = -1;
+    return hyped::io::CanResult::kFailure;
+  }
+  logger_.log(hyped::core::LogLevel::kInfo, "CAN socket successfully created");
+  return hyped::io::CanResult::kSuccess;
+}
+
+CanResult Can::sendCanFrame(const core::CanFrame message)
```
mifrandir, line 44: Why does `CanFrame` live in `core` rather than `io`? I'm not saying that its necessarily bad but we want to have some good heuristics for what goes where? @kshxtij

lib/io/can.hpp
```diff
@@ -1,8 +1,22 @@
+#include <string>
+
+#include <linux/can.h>
+
+#include <core/logger.hpp>
 #include <core/types.hpp>
 
 namespace hyped::io {
+
+enum CanResult { kFailure, kSuccess };
```
mifrandir, line 10: `enum class`

lib/io/can.hpp
```diff
@@ -1,8 +1,22 @@
...
 class Can {
  public:
-  Can();
+  Can(hyped::core::ILogger &logger);
+  CanResult initialiseCanSocket(std::string can_network_interface);
```
mifrandir, line 14: `const`

maybe make it a reference?

lib/io/can.hpp
```diff
@@ -1,8 +1,22 @@
...
+  CanResult sendCanFrame(const core::CanFrame message);
```
mifrandir, line 15: I believe it is possible to make this reference, too.

lib/io/can.hpp
```diff
@@ -1,8 +1,22 @@
...
+  std::optional<can_frame> receiveCanFrame();
```
mifrandir, line 16: why `can_frame` rather than `core::CanFrame`?

mifrandir (changes requested):

lib/io/can.cpp
mifrandir: But `io/` will be shared between the two?

lib/io/can.hpp
mifrandir: I think we should stick with one of the options wherever we can. I believe that `can_frame` is not Darwin friendly so we should define `core::CanFrame` to be `can_frame` on Linux and custom on Darwin. Once we have done that, we only ever refer to `core::CanFrame` again, regardless of architecture.

lib/io/can.cpp
TomLonergan03: I'm pretty sure, but haven't tested, that this all works with multiple networks if we instantiate a new Can class and initialise with a new network interface. But also we shouldn't ever have more than one bus, containing BMS, controller, and (maybe) nucleo board

lib/core/types.hpp
```diff
@@ -7,6 +7,10 @@ enum class DigitalSignal { kLow = 0, kHigh };
 
 using Float = float;
 
+#if LINUX
+#include <linux/can.h>
+struct CanFrame = can_frame;
```
TomLonergan03, line 12: Is this the best way to do this? I'm not confident with preprocessor directives but it seems weird to have to include `<linux/can.h>` here

TomLonergan03: Actually this straight up doesn't compile, but I'm not sure how to do it right

lib/io/can.cpp
kshxtij: That is fair, @TomLonergan03 could you move the definition to can.hpp?

lib/io/can.hpp
```diff
@@ -1,8 +1,33 @@
+#include <string>
+
+#include <core/logger.hpp>
 #include <core/types.hpp>
 
+#ifndef LINUX
```
kshxtij, line 6: `#if LINUX`

lib/io/can.hpp
```diff
@@ -1,8 +1,33 @@
...
+#include <linux/can.h>
+#define CanFrame can_frame
+#else
+struct CanFrame {
```
kshxtij, line 10: You want these within the namespace so the types are actually accessibly outside this file

kshxtij (changes requested):

lib/motors/can_processor.hpp
```diff
@@ -3,15 +3,16 @@
 #include <cstdint>
 
 #include <core/types.hpp>
+#include <io/can.hpp>
 
 namespace hyped::motors {
 
 class CanProcessor {
```
mifrandir, line 10: Based on what I assume is going to happen, you will require an interface `ICanProcessor` in `can.hpp`. The class here should implement that interface. (You will probably want to have a collection like `std::unordered_map<uint32_t,std::vector<std::shared_ptr<ICanProcessor>>>` in the main `Can` instance.

I am doubtful whether `CanProcessor` is a good name after these changes have been made. For one, I would call the interface `ICanSubscriber`, because each subscriber listens to a particular set of IDs. This particular instance will then be responsible for what? Handling messages from the motors? So maybe it's just a `Controller`? Food for thought.

lib/motors/can_processor.hpp
```diff
@@ -3,15 +3,16 @@
...
  public:
   CanProcessor();
 
-  bool sendMessage(const core::CanFrame frame);
-  void processMessage(const core::CanFrame frame);
+  bool sendMessage(const CanFrame frame);
```
mifrandir, line 14: How is this different from `Can::sendMessage`?

lib/io/can.hpp
```diff
@@ -1,19 +1,30 @@
...
+  uint8_t __res0;  /* reserved / padding */
+  uint8_t __res1;  /* reserved / padding */
+  std::array<uint8_t, 8> data;
+};
+#endif
 namespace hyped::io {
 
-enum CanResult { kFailure, kSuccess };
+enum class CanResult { kFailure, kSuccess };
 class Can {
```
mifrandir, line 22: You probably also want to abstract this into a `ICan` instance, similar to what's happened to GPIO. However, note that this is different from the `CanProcessor` interface below! (I can see an implementation where we have the latter but not the former, so maybe this one goes into the next PR? I might even be inclined to do it some time...)

mifrandir (changes requested):

lib/motors/can_processor.hpp
TomLonergan03: What is the purpose of the collection?
I'm also not 100% sure the canprocessor class is needed at all tbh, my original thinking was that when `can.cpp` receives a message then `CanProcessor` would take that and make it into what we want to use, but probably that would just happen in `controller` anyway

TomLonergan03:
Another todo, I'm not sure exactly how the receive buffer works, the documentation seems kinda sparse. Will try to test.

lib/motors/can_processor.hpp
mifrandir: So the original point of `CanProcessor` was the following:
- you may have many different CAN devices with many different IDs
- some of them you are interested in, some of them you aren't interested in (because CAN messages get sent to everybody)
- for those that you are interested in, there are many different clients (e.g. motors vs bms)
- in order to handle this elegantly you may allow clients to tell you what they're interested in
- so motors may implement something called a `CanProcessor` and give it to `Can` saying "hi there, we would like to receive messages to ID 0xabcde" 
- `Can` says "okay, cool, I'll write that down"
- when `Can` receives a message, it checks the ID and finds all the `CanProcessor`s that registered interest

Now, you need the data structure to do the last step. Previously we hade a vector of these and just went through them in linear time. That's not happening on my watch.

TomLonergan03: That makes sense. And because motors vs BMS will want to do different things when they receive a message it's an interface so there are different motor can processor and bms can processor and whatever else.

So when a message is received, the ID is checked against the map then it calls `canprocessor.newMessage()` or something similar for every canprocessor in the vector?

mifrandir: Exactly. In your current interface it would be `processMessage`, I believe.

lib/io/can.hpp
```diff
@@ -23,10 +27,18 @@ struct CanFrame {
 
 enum class CanResult { kFailure, kSuccess };
 
+class ICanProcessor {
+ public:
+  virtual void processMessage(const io::CanFrame frame);
```
mifrandir, line 32: `virtual void process(const io::CanFrame &frame) = 0;`?

- name
- pure virutal (`= 0`)
- reference

lib/io/can.hpp
```diff
@@ -23,10 +27,18 @@ struct CanFrame {
...
+};
+
 class ICan {
+ public:
   virtual CanResult initialise(const std::string &can_network_interface);
   virtual CanResult send(const CanFrame &message);
   virtual std::optional<CanFrame> receive();
+  virtual CanResult addCanProcessor(uint16_t ID, ICanProcessor processor);
+  virtual CanResult removeCanProcessor(ICanProcessor processor);
```
mifrandir, line 41: all pure virtual, please

lib/io/can.hpp
```diff
@@ -39,5 +51,6 @@ class Can : ICan {
  private:
   int socket_;
   hyped::core::ILogger &logger_;
+  std::unordered_map<uint32_t, std::vector<std::shared_ptr<ICanProcessor>>> processors_;
```
mifrandir, line 54: If you're here, than you can implement the logic, too.

lib/io/can.hpp
mifrandir: Also `removeCanProcessor` may be hard to realise.

1. it's not something we anticipate to need
2. removing things from a vector is hard so we would require a different data structure (`unordered_set`)

I'd say scratch this.

lib/io/can.hpp
```diff
@@ -23,10 +27,18 @@ struct CanFrame {
...
+ public:
+  virtual void processMessage(const io::CanFrame frame);
+};
+
 class ICan {
+ public:
   virtual CanResult initialise(const std::string &can_network_interface);
   virtual CanResult send(const CanFrame &message);
   virtual std::optional<CanFrame> receive();
+  virtual CanResult addCanProcessor(uint16_t ID, ICanProcessor processor);
```
mifrandir, line 40: `const`

`ID` is upper case

be careful with passing by value vs passing by reference here; copying this over could be a problem? I don't know. The safest thing is to require a `std::shared_ptr`.

lib/motors/can_processor.hpp
```diff
@@ -7,7 +7,7 @@
 
 namespace hyped::motors {
 
-class CanProcessor {
+class CanProcessor : io::ICanProcessor {
```
mifrandir, line 10: `public`

lib/io/can.hpp
```diff
@@ -23,10 +27,18 @@ struct CanFrame {
...
+  virtual CanResult removeCanProcessor(ICanProcessor processor);
 };
 
 class Can : ICan {
```
mifrandir, line 44: `public`

mifrandir (changes requested):

lib/io/can.cpp
```diff
@@ -8,11 +8,13 @@
 #include <sys/socket.h>
 
 namespace hyped::io {
+
 Can::Can(core::ILogger &logger) : logger_(logger)
 {
+  processors_ = {};
```
mifrandir, line 14: Initialise above, similar to `logger_`.

lib/io/can.cpp
```diff
@@ -94,4 +96,24 @@ std::optional<CanFrame> Can::receive()
   return received_message;
 }
 
+void Can::listen()
+{
+  std::optional<CanFrame> data;
+  if (ioctl(socket_, FIONREAD) >= sizeof(CanFrame)) { data = receive(); }
+  if (data == std::nullopt) { return; }
```
mifrandir, line 103: I don't like the logic here.

```cpp
if (ioctl(socket_, FIONREAD) < sizeof(CanFrame)) { return; }
const auto data = receive();
if (!data) {
  return;
}
```

mifrandir: Also, this looks like it can succeed and fail? Maybe you should have a non-void return type here to indicate what happened to the caller.

lib/io/can.cpp
```diff
@@ -94,4 +96,24 @@ std::optional<CanFrame> Can::receive()
...
+  const CanFrame message = data.value();
+  if (!processors_.contains(message.can_id)) {
+    logger_.log(core::LogLevel::kInfo, "No CanProccessor associated with id %i", message.can_id);
+  }
+  for (uint8_t i = 0; i < processors_[message.can_id].size(); i++) {
```
mifrandir, line 108: Lots of things can go wrong here.

Firstly, you need to check whether the key `message.can_id` exists. Then you need to retrieve the value and then you can iterate over it like so:

```cpp
for (auto &processors : subscribed_processors) {
  processor->processMessage(message);
}
```

mifrandir (changes requested):

lib/io/can.cpp
```diff
@@ -9,111 +9,113 @@
...
-              static_cast<int>(message.data[0]),
-              static_cast<int>(message.data[1]),
-              static_cast<int>(message.data[2]),
-              static_cast<int>(message.data[3]),
-              static_cast<int>(message.data[4]),
-              static_cast<int>(message.data[5]),
-              static_cast<int>(message.data[6]),
-              static_cast<int>(message.data[7]));
+  // TODOLater make log_ more elegant
+  log_.log(core::LogLevel::kDebug,
```
mifrandir, line 56: why did you name this back?

lib/io/can.cpp
```diff
@@ -9,111 +9,113 @@
...
+  if (!data) { return io::CanResult::kFailure; }
+  CanFrame message = data.value();
   if (!processors_.contains(message.can_id)) {
-    logger_.log(core::LogLevel::kInfo, "No CanProccessor associated with id %i", message.can_id);
+    log_.log(core::LogLevel::kInfo, "No CanProccessor associated with id %i", message.can_id);
+    return io::CanResult::kFailure;
   }
-  for (uint8_t i = 0; i < processors_[message.can_id].size(); i++) {
-    processors_[message.can_id][i]->processMessage(message);
+  std::vector<std::shared_ptr<ICanProcessor>> &subscribed_processors = processors_[message.can_id];
```
mifrandir, line 108: again, what if `message.can_id` isn't in the map? use `std::unordered_map::find`!

mifrandir (changes requested):

lib/io/can.cpp
TomLonergan03: log_ is the standard used everywhere else as far as I can see

lib/io/can.cpp
TomLonergan03: isn't this already caught by if (!processors_.contains(message.can_id)) on line 104?

lib/io/can.cpp
mifrandir: i think the standard has changed to `logger_`. in particular, there was at least one PR for changing `log_` to `logger_` that I've seen

lib/io/can.cpp
mifrandir: oh, you're right. my bad. here's the idiomatic way to do it (fewer lookups; also `std::map::contains` didn't exist pre C++20 so you always had to do this)

```cpp
auto it = processors_.find(message.can_id);
if (it == processors.end()) {
   // handle error
}
for (auto &processor : it->second) { 
  // do stuff
}
```

I don't think we like the variable name `it` here so call it something like ` canIdAndProcessor`.

mifrandir: It also might be possible to sprinkle in some `const`s here; not sure whether you can put it on the iterator because the semantics of what is and isn't immutable due to const are a bit fuzzy. try it out and see what works.

lib/io/can.cpp
```diff
@@ -1,10 +1,121 @@
...
+  for (auto &processor : subscribed_processors) {
+    processor->processMessage(message);
+  }
+  return io::CanResult::kSuccess;
+}
+
+void Can::addCanProcessor(const uint16_t id, std::shared_ptr<ICanProcessor> processor)
 {
-  // TODOLater implement
+  processors_[id].push_back(processor);
```
mifrandir, line 117: what if this one is empty? :)

lib/io/can.cpp
```diff
@@ -1,10 +1,121 @@
...
+  }
+  sockaddr_can socket_address;
+  socket_address.can_family  = AF_CAN;
+  socket_address.can_ifindex = if_nametoindex(can_network_interface.c_str());
+  if (socket_address.can_ifindex == 0) {
+    log_.log(core::LogLevel::kFatal, "Unable to find CAN1 network interface");
+    close(socket_);
+    socket_ = -1;
+    return io::CanResult::kFailure;
+  }
```
mifrandir, line 31: The following would avoid the mutable `sockaddr_can`:

```cpp
const auto interface_index = if_nametoindex(can_network_interface.c_str());
if (!interface_index) {
  // handle error
}
const sockaddr_can socket_address = { AF_CAN, interface_index };
```

lib/io/can.cpp
```diff
@@ -1,10 +1,121 @@
...
+  const int bind_status
+    = bind(socket_, reinterpret_cast<sockaddr *>(&socket_address), sizeof(socket_address));
```
mifrandir, line 33: maybe put a comment on why we do the `reinterpret_cast` here? I believe it's an acceptable way to do it but it's non-obvious. I had to look up the `bind` man page.  https://www.man7.org/linux/man-pages/man2/bind.2.html

mifrandir (changes requested):
It's starting to look like C++!

lib/io/can.cpp
TomLonergan03: how do i do this properly? i've not been able to get insert to work, is that the right way or is it something else?

lib/io/can.cpp
TomLonergan03: i believe the iterator is alright, it builds and i can't see anything obvious online on why it wouldn't work, but i'm not 100% confident

TomLonergan03:
`CanFrame message = data.value();`
This is failing clang-format for some reason. I'm running it locally and it doesn't get touched so I'm not sure what the issue is

kshxtij:
You need

```
  CanFrame message                 = data.value();
 ```

File: lib/io/can.cpp [hunk 41]
```diff
@@ -9,113 +9,113 @@
...
+              static_cast<int>(data.data[7]));
   return io::CanResult::kSuccess;
 }
 
 std::optional<CanFrame> Can::receive()
 {
   CanFrame received_message;
   if (ioctl(socket_, FIONREAD) < sizeof(CanFrame)) {
-    log_.log(core::LogLevel::kDebug, "No can data in rx queue");
+    logger_.log(core::LogLevel::kDebug, "No can data in rx queue");
```
mifrandir, line 73: capitalise `CAN`

lib/io/can.cpp
```diff
@@ -9,113 +9,113 @@
...
-CanResult Can::listen()
+io::CanResult Can::listen()
 {
   if (ioctl(socket_, FIONREAD) < sizeof(CanFrame)) { return io::CanResult::kFailure; }
   const auto data = receive();
   if (!data) { return io::CanResult::kFailure; }
-  CanFrame message = data.value();
-  if (!processors_.contains(message.can_id)) {
-    log_.log(core::LogLevel::kInfo, "No CanProccessor associated with id %i", message.can_id);
+  CanFrame message                 = data.value();
```
mifrandir, line 102: `const auto`?

lib/io/can.cpp
```diff
@@ -9,113 +9,113 @@
...
     processor->processMessage(message);
   }
   return io::CanResult::kSuccess;
 }
 
 void Can::addCanProcessor(const uint16_t id, std::shared_ptr<ICanProcessor> processor)
 {
-  processors_[id].push_back(processor);
-  log_.log(core::LogLevel::kInfo, "Added processor for id %i", id);
+  if (processors_.contains(id)) { processors_[id].push_back(processor); }
```
mifrandir, line 116: i don't like the square bracket indexing because it hides some complexity. to me it looks like something as simple as pointer arithmetic whereas in fact you are doing a lookup in a hash table.

also you are still doing the lookup twice.

```cpp
const auto id_and_processors = processors_.find(id);
if (id_and_processors == processors_.end()) {
  processors_.emplace(id, std::vector<...>({ processor }));
} else {
  id_and_processors->second.push_back(processor);
}
```

mifrandir (changes requested):
almost there

lib/io/can.hpp
```diff
@@ -1,10 +1,75 @@
...
+    } tp;
+    /* reserved for future CAN protocols address information */
+  } can_addr;
+};
+#define PF_CAN 29
+#define AF_CAN 29
+#define CAN_RAW 1
+#endif
+
+enum class CanResult { kFailure, kSuccess };
```
SnickeyX, line 44: can use core::Result now

lib/io/can.hpp
```diff
@@ -1,10 +1,75 @@
...
+  virtual void processMessage(const io::CanFrame &frame) = 0;
+};
+
+class ICan {
+ public:
+  virtual CanResult initialise(const std::string &can_network_interface) = 0;
+  virtual CanResult send(const CanFrame &message)                        = 0;
+  virtual std::optional<CanFrame> receive()                              = 0;
+  virtual CanResult listen()                                             = 0;
+  virtual void addCanProcessor(uint16_t ID, ICanProcessor &processor)    = 0;
```
SnickeyX, line 57: `const`?

SnickeyX (changes requested):

lib/io/can.hpp
mifrandir: also naming

lib/io/can.cpp
```diff
@@ -99,7 +99,7 @@ io::CanResult Can::listen()
   if (ioctl(socket_, FIONREAD) < sizeof(CanFrame)) { return io::CanResult::kFailure; }
   const auto data = receive();
   if (!data) { return io::CanResult::kFailure; }
-  CanFrame message                 = data.value();
+  const CanFrame message           = data.value();
```
mifrandir, line 102: sorry, i know i've commented on this line before

common practice with `std::optional`s is to dereference it when required. 

so you would do something like `const auto message = receive(); if (!message) { /* ... */  }` 

and then `processors_.find(message->can_id);`

there is a bit of a style question here that i'm not sure how to resolve: why am I fine with using the pointer notation (`*x` and `x->member`) here but not for collections (`x.at(i)` vs `x[i]`)? i genuinely don't know. the honest answer is that this is how i was taught to do it. a possible justification is this: accessing the value inside an optional is O(1) just like a pointer derefrence. the same doesn't hold for collections in general. i suppose by that logic we should use the `x[i]` notation for vectors but not for maps?

mifrandir (changes requested):

lib/io/can.cpp
TomLonergan03: is dereferencing it as `*message` to pass into process message ok? it feels wrong, but `&message` doesn't work.
or is `.value()` the best way?

mifrandir: `*message` is the correct way to do it.

`&message` doesn't make sense because it gives you a reference.

lib/io/can.hpp
```diff
@@ -1,12 +1,74 @@
...
  public:
-  Can();
+  virtual core::Result initialise(const std::string &can_network_interface)      = 0;
+  virtual core::Result send(const CanFrame &message)                             = 0;
+  virtual std::optional<CanFrame> receive()                                      = 0;
+  virtual core::Result listen()                                                  = 0;
+  virtual void addCanProcessor(const std::uint16_t id, ICanProcessor &processor) = 0;
+};
+
+class Can : public ICan {
```
mifrandir, line 59: I think this class should be `HardwareCan` and in a separate file `hardware_can.[ch]pp`. This avoid people accidentally including `can.hpp` and using `Can` rather than `ICan`.

lib/io/can.cpp
```diff
@@ -1,10 +1,124 @@
...
+  }
+  return core::Result::kSuccess;
+}
+
+void Can::addCanProcessor(const std::uint16_t id, std::shared_ptr<ICanProcessor> processor)
 {
-  // TODOLater implement
+  const auto id_and_processors = processors_.find(id);
+  if (id_and_processors == processors_.end()) {
+    processors_.emplace(id, std::vector<std::shared_ptr<ICanProcessor>>({processor}));
```
mifrandir, line 117: Not a fan of this line. Can you split this up nicely in order to avoid complexity?

lib/io/can.hpp
```diff
@@ -1,12 +1,74 @@
...
+
+class ICanProcessor {
+ public:
+  virtual void processMessage(const io::CanFrame &frame) = 0;
+};
+
+class ICan {
  public:
-  Can();
+  virtual core::Result initialise(const std::string &can_network_interface)      = 0;
```
mifrandir, line 52: I don't think we want this function. I think whenever the caller has a CAN object, it should already be initialised.

If there are any strong reasons against this (e.g we need to make an object on init but only call the initialise method much later into the run), which I doubt, the interface should be extended to include something like `IUninitialisedCan` which has an initialise function that optionally returns an `ICan` instance. This would allow the user calling e.g. `ICan::send` before calling `ICan::initialise`.

mifrandir (changes requested):

lib/io/can.hpp
TomLonergan03: is that rename can.cpp and split `ICan` and `Can` into seperate hpp files?

lib/io/can.hpp
TomLonergan03: should this be done in the constructor then?

lib/io/can.hpp
mifrandir: you should end up with three files

- `can.hpp` containing `ICan`
- `hardware_can.hpp` defining `HardwareCan`
- `hardware_can.cpp` implementing `HardwareCan`

lib/io/hardware_can.cpp
```diff
@@ -114,7 +114,8 @@ void Can::addCanProcessor(const std::uint16_t id, std::shared_ptr<ICanProcessor>
 {
   const auto id_and_processors = processors_.find(id);
   if (id_and_processors == processors_.end()) {
-    processors_.emplace(id, std::vector<std::shared_ptr<ICanProcessor>>({processor}));
+    auto new_processor_vector = std::vector<std::shared_ptr<ICanProcessor>>({processor});
```
mifrandir, line 117: don't like the name here; processors_for_id maybe? idk

lib/io/can.hpp
TomLonergan03: if this is done in the constructor how should we handle failing to open a socket as it can't return failures any more

lib/io/hardware_can.cpp
```diff
@@ -0,0 +1,127 @@
...
+    logger.log(core::LogLevel::kFatal, "Unable to find CAN1 network interface");
+    close(socket_id);
+    return std::nullopt;
+  }
+  const sockaddr_can socket_address = {AF_CAN, interface_index};
+  const int bind_status
+    = bind(socket_id, reinterpret_cast<const sockaddr *>(&socket_address), sizeof(socket_address));
+  if (bind_status < 0) {
+    logger.log(core::LogLevel::kFatal, "Unable to bind CAN socket");
+    close(socket_id);
```
mifrandir, line 31: you probably want to do this in a deconstructor, too.

but beware, there's some subtlety here: you return a `HardwareCan`; this means that deallocation happens all the time, whenever the object is copied somewhere; things might also get out of sync between the copies in other ways

solution: return a shared pointer; of course this doesn't quite prevent anybody from copying the object around but it conveys the intention. we could make this impossible on a type level but i don't believe that that's worth the hassle

lib/io/hardware_can.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include <memory>
+#include <optional>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include <core/logger.hpp>
+#include <core/types.hpp>
+namespace hyped::io {
+class HardwareCan : public ICan {
```
mifrandir, line 14: I think we put two empty lines around the namespace here

mifrandir (changes requested):

lib/io/hardware_can.cpp
```diff
@@ -0,0 +1,132 @@
...
+    processor->processMessage(*message);
+  }
+  return core::Result::kSuccess;
+}
+
+void HardwareCan::addCanProcessor(const std::uint16_t id, std::shared_ptr<ICanProcessor> processor)
+{
+  const auto id_and_processors = processors_.find(id);
+  if (id_and_processors == processors_.end()) {
+    auto processors_for_id = std::vector<std::shared_ptr<ICanProcessor>>({processor});
```
mifrandir, line 124: `const`?

lib/io/hardware_can.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+class HardwareCan : public ICan {
+ public:
+  static std::optional<std::shared_ptr<HardwareCan>> create(
+    core::ILogger &logger, const std::string &can_network_interface);
+  ~HardwareCan();
+  core::Result send(const CanFrame &message);
+  std::optional<CanFrame> receive();
+  core::Result listen();
+  void addCanProcessor(const std::uint16_t id, std::shared_ptr<ICanProcessor> processor);
```
mifrandir, line 24: `addProcessor`?

lib/io/hardware_can.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include <net/if.h>
+#include <sys/ioctl.h>
+#include <sys/socket.h>
+
+namespace hyped::io {
+
+std::optional<std::shared_ptr<HardwareCan>> HardwareCan::create(
+  core::ILogger &logger, const std::string &can_network_interface)
+{
+  std::int16_t socket_id = socket(PF_CAN, SOCK_RAW, CAN_RAW);
```
mifrandir, line 15: `const`

also this returns `int`

https://www.man7.org/linux/man-pages/man2/socket.2.html

lib/io/hardware_can.cpp
```diff
@@ -0,0 +1,132 @@
...
+  if (socket_id < 0) {
+    logger.log(core::LogLevel::kFatal, "Unable to open CAN socket");
+    return std::nullopt;
+  }
+  const std::uint8_t interface_index = if_nametoindex(can_network_interface.c_str());
```
mifrandir, line 20: this returns `unsigned int` aka `std::uint32_t` (or just `auto`)

https://www.man7.org/linux/man-pages/man3/if_nametoindex.3.html

don't just make up types!

mifrandir (changes requested):

lib/io/hardware_can.cpp
```diff
@@ -12,18 +12,18 @@ namespace hyped::io {
...
   }
-  const std::uint8_t interface_index = if_nametoindex(can_network_interface.c_str());
+  const auto interface_index = if_nametoindex(can_network_interface.c_str());
   if (!interface_index) {
     logger.log(core::LogLevel::kFatal, "Unable to find CAN1 network interface");
     close(socket_id);
     return std::nullopt;
   }
-  const sockaddr_can socket_address = {AF_CAN, interface_index};
+  const sockaddr_can socket_address = {AF_CAN, static_cast<int>(interface_index)};
```
mifrandir, line 26: why this cast? isn't this already an `int`?

maxguy2001 (changes requested):

lib/io/can.hpp
```diff
@@ -1,12 +1,54 @@
...
+  std::uint8_t __res0;  /* reserved / padding */
+  std::uint8_t __res1;  /* reserved / padding */
+  std::array<std::uint8_t, 8> data;
+};
+struct sockaddr_can {
+  std::uint8_t can_family;
+  std::uint8_t can_ifindex;
+  union {
+    /* transport protocol class address info (e.g. ISOTP) */
+    struct {
```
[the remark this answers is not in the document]
TomLonergan03: i guess i could `define` everything/use macros but i feel like that isn't better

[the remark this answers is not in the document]
maxguy2001: `define` isn't really much better. It's just that the naming here makes it much harder for someone who is new to can/mtc to follow what's actually being input and returned but if that's something we can't change then I guess we'll just have to either accept it or rewrite the entirety of can... hackathon proj?

lib/io/hardware_can.cpp
```diff
@@ -0,0 +1,132 @@
...
+    logger.log(core::LogLevel::kFatal, "Unable to open CAN socket");
+    return std::nullopt;
+  }
+  const int interface_index = if_nametoindex(can_network_interface.c_str());
+  if (!interface_index) {
+    logger.log(core::LogLevel::kFatal, "Unable to find CAN1 network interface");
+    close(socket_id);
+    return std::nullopt;
+  }
+  const sockaddr_can socket_address = {AF_CAN, interface_index};
```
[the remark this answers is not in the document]
mifrandir: yup, looks like you had it the wrong way around to begin with :^)

lib/io/can.hpp
```diff
@@ -1,12 +1,54 @@
...
+// structs and values as defined in <linux/can.h>
+struct CanFrame {
+  std::uint32_t can_id; /* 32 bit CAN_ID + EFF/RTR/ERR flags */
+  std::uint8_t can_dlc; /* frame payload length in byte (0 .. CAN_MAX_DLEN) */
+  std::uint8_t __pad;   /* padding */
+  std::uint8_t __res0;  /* reserved / padding */
+  std::uint8_t __res1;  /* reserved / padding */
+  std::array<std::uint8_t, 8> data;
+};
+struct sockaddr_can {
```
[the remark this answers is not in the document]
mifrandir: this one is fine because it comes from the linux header

lib/io/can.hpp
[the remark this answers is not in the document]
mifrandir: i think this is fine because we're just reimplementing the built-in functions

TomLonergan03:
This works, I have tested it. The receive buffer will buffer at least 3 messages, but further testing is required to determine how large it is.

TomLonergan03:
For the buffer I believe a can frame is around 13 bytes total, as far as I have discovered a can socket seems to have around 160kB of buffer which would imply ~12,000 messages can be in the rx buffer, equivalent to ~0.3 seconds of continuous sending at 500kbaud. This will need to be experimentally validated

lib/io/hardware_can.cpp
```diff
@@ -0,0 +1,138 @@
...
+void HardwareCan::addProcessor(const std::uint16_t id, std::shared_ptr<ICanProcessor> processor)
+{
+  const auto id_and_processors = processors_.find(id);
+  if (id_and_processors == processors_.end()) {
+    const auto processors_for_id = std::vector<std::shared_ptr<ICanProcessor>>({processor});
+    processors_.emplace(id, processors_for_id);
+  } else {
+    id_and_processors->second.push_back(processor);
+  }
+  logger_.log(core::LogLevel::kInfo, "Added processor for id %i", id);
```
SnickeyX, line 135: change to `kDebug`

lib/io/hardware_can.hpp
```diff
@@ -0,0 +1,33 @@
...
+    core::ILogger &logger, const std::string &can_network_interface);
+  HardwareCan(core::ILogger &logger, const int socket);
+  ~HardwareCan();
+  core::Result send(const CanFrame &message);
+  std::optional<CanFrame> receive();
+  core::Result listen();
+  void addProcessor(const std::uint16_t id, std::shared_ptr<ICanProcessor> processor);
+
+ private:
+  int socket_;
```
SnickeyX, line 28: const

SnickeyX: also ordering, so logger_ first

SnickeyX (changes requested):

maxguy2001 (approved):

BenET01 (approved):

lib/io/can.hpp
```diff
@@ -1,12 +1,54 @@
...
-  Can();
+  virtual void processMessage(const io::CanFrame &frame) = 0;
 };
 
-}  // namespace hyped::io
+class ICan {
+ public:
+  virtual core::Result send(const CanFrame &message)                                          = 0;
+  virtual std::optional<CanFrame> receive()                                                   = 0;
+  virtual core::Result listen()                                                               = 0;
```
mifrandir, line 51: The name here isnt very descriptive, esp in the presence of receive.

I'm not even sure why we demand that an ICan instance implements both of these.

Also: What if theres more than one message waiting? Do we ever encounter a case where there's more messages coming in than we are collecting?

mifrandir (changes requested):

SnickeyX (approved):
lgtm, apart from what franz mentioned above

lib/io/can.hpp
TomLonergan03: the socket buffers some number of messages but i haven't determined an exact number, 3 <= n < 12000

kshxtij (approved):

PR #22, opened 2022-10-28
Varying time tests for logger

mifrandir:


mifrandir:
Actually, when we are at time 0, why do we print `01:00:00.000`, @kshxtij?

kshxtij (approved):

kshxtij:
localtime vs gmtime semantics

TomLonergan03 (approved):

mifrandir:
> localtime vs gmtime semantics

but we don't want our tests to depend on the timezone, do we?

kshxtij:
I see what you mean, but we'd want our logs to display timestamps based on the local timezone

mifrandir:
> I see what you mean, but we'd want our logs to display timestamps based on the local timezone

So I've tried to fix this put `std::chrono::parse` seems to not exist yet so this is a future problem. A solution would be to figure out what "00:00:00.000" on 1970-01-01 gets parsed to in local time and then add an offset to that. Not to big an issue, though.

mifrandir:
@kshxtij merge whenever you want

PR #26, opened 2022-10-29
SNS - Interface

SnickeyX:
Goal of the I2c sensor interface is to be able to use this abstract class with the code for the relevant mux. Having this interface allows us to generalize the mux code to handle any I2C sensor connected to it. 

- [x] Agree on the best return type for the read function.


SnickeyX:
The biggest issue for me is setting the correct return type for read - I have temporarily kept it at uint16_t which works with current I2C sensors that I am aware of but this is bound to change.

lib/sensors/sensors.hpp
```diff
@@ -0,0 +1,23 @@
...
+ * If a sensor is to be used with an I2C mux, it must inherit from this abstract class.
+ */
+class II2cSensor {
+ public:
+  /*
+   * This function carries out the initilization steps for a particular sensor.
+   */
+  virtual bool configure()    = 0;
+  virtual WriteResult write() = 0;
+  virtual uint16_t read()     = 0;
```
kshxtij, line 20: https://github.com/Hyp-ed/hyped-2023/blob/9a3285262877e8cb0e3a3bdb00dadacc28cbcf76/lib/io/i2c.cpp#L43

Would we want the sensor returning an optional since the i2c interface does?

SnickeyX: Good point, will change

lib/sensors/sensors.hpp
```diff
@@ -0,0 +1,24 @@
...
+ * If a sensor is to be used with an I2C mux, it must inherit from this abstract class.
+ */
+class II2cSensor {
+ public:
+  /*
+   * This function carries out the initilization steps for a particular sensor.
+   */
+  virtual bool configure()               = 0;
+  virtual WriteResult write()            = 0;
+  virtual std::optional<uint16_t> read() = 0;
```
mifrandir, line 21: Do we need both read and write access at all times? Surely most users will only require one of these.

mifrandir (changes requested):
I perfer the design where we have one toplevel instance/manager/factory (`II2c`) that allows us to create sensors. With this design you need to make all the sensors on startup and pass them in to the main function (rather than providing callees with the means to read to/write from I2c at will). That's what I did for GPIO. Is there any reason why this wouldn't work here? Besides not being excessively verbose, this has the neat effect that we can manage file descriptors within this instance and don't need the callees to be efficient.

lib/sensors/sensors.hpp
SnickeyX: On second thought, the write function is redundant since thats what configure essentially does. Will remove that. We definitely need read though.

mifrandir: Maybe call it `I2cReader` then? For consistency.

SnickeyX: But the classes deriving this will be like `ITemperature` or `IAccelerometer`  which won't just be readers?

kshxtij: FWIW, we need a generic II2cSensor interface specifically because of the Multiplexer class that needs to be implemented soon. The Multiplexer class would need to accept II2c instances as a hardware mux could have any type of i2c sensor connected to it.

mifrandir: > But the classes deriving this will be like `ITemperature` or `IAccelerometer` which won't just be readers?

I don't see anything wrong with this. Also, I don't think these will derive from this but rather wrap around it. I.e. they add some sort of code that interprets the `std::optional<uint16_t>` as a temperature or an acceleration and possibly does some setup.

If you really want to make these abstract (`ITemperature` vs `Temperature`) and it wasn't just a typo I don't think an accelerometer interface should have anything to do with I2C specifically.

> FWIW, we need a generic II2cSensor interface specifically because of the Multiplexer class that needs to be implemented soon. The Multiplexer class would need to accept II2c instances as a hardware mux could have any type of i2c sensor connected to it.

Let's chat tonight? Not sure I agree or understand.

lib/sensors/main.cpp
```diff
@@ -0,0 +1,7 @@
+namespace hyped::sensors {
```
kshxtij, line 1: just leave this file empty, we need a cpp file in the directory but don't need it to have anything. 

This file should be deleted in one of the temperature or accelerometer prs anyways as soon as we have an actual cpp file in here

kshxtij (changes requested):
LGTM, just clear that file imo

lib/sensors/main.cpp
mifrandir: there shouldn't be anything called `main.cpp` or `main.hpp` imo

mifrandir: Maybe an empty `sensors.cpp`?

mifrandir (approved):

lib/sensors/main.cpp
SnickeyX: yeah will delete after either temp or acc get approved and merged into here

kshxtij: @SnickeyX we don't need this pr to be open until the others are done. Replace main.cpp with an empty sensors.cpp, we merge this pr and then the other prs automatically shift bases to master.

SnickeyX: W

kshxtij (approved):

PR #30, opened 2022-11-01
Debug Binary

kshxtij:
- [x] Command Parsing
- [x] IO Commands 
    - [x] I2C
    - [x] ADC
- [x] Config Parsing


src/debugger/main.cpp
```diff
@@ -0,0 +1,21 @@
+#include <iostream>
+
```
maxguy2001, line 2: `#include <string>`. I'm assuming it's not broken since it's included in one of the other headers you use but best practice is include what you use?

kshxtij: Good catch but this is just a skeleton to remind me to fix this

lib/debug/repl.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <core/logger.hpp>
+#include <io/adc.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::debug {
+
+struct Command {
+  std::string name;
+  std::string description;
+  std::function<void(void)> callback;
```
mifrandir, line 15: `handler`?

lib/debug/repl.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+  std::cout << "> ";
+  std::string command;
+  std::getline(std::cin, command);
+  auto identifier = command_map_.find(command);
+  if (identifier == command_map_.end()) {
+    log_.log(hyped::core::LogLevel::kFatal, "Unknown command: %s", command.c_str());
+    return;
+  }
+  identifier->second.callback();
```
mifrandir, line 67: calling `second` on `identifier` seems awkward, suggesting your name is off.

now, I personally don't have a problem with `it` which is the C++ way to name this; however, in the spirit of the style guide something like `nameAndCommand` would be better.

lib/debug/repl.hpp
```diff
@@ -0,0 +1,36 @@
...
+ private:
+  void printCommands();
+  void addCommand(const Command &cmd);
+  void addQuitCommand();
+  void addHelpCommand();
+  void handleCommand();
+
+  void addAdcCommands(const uint8_t pin);
+
+  std::vector<Command> commands_;
```
mifrandir, line 32: Unsure what we need this for? If it's to print help you could just use an ordered map and iterate over it? This would have the benefit of putting the commands in alphabetical order.

mifrandir (changes requested):

kshxtij:
Thoughts on JSON config to enable/disable commands vs other forms of configs (idk cmdline args? flags?)

lib/debug/repl.cpp
```diff
@@ -28,29 +27,30 @@ void Repl::addHelpCommand()
...
+    };
+  }
 }
 
 void Repl::printCommands()
 {
   log_.log(hyped::core::LogLevel::kInfo, "Available commands:");
-  for (const auto &cmd : commands_) {
-    log_.log(hyped::core::LogLevel::kInfo, "  %s: %s", cmd.name.c_str(), cmd.description.c_str());
+  for (const auto &command : command_map_) {
```
mifrandir, line 49: this isn't just a `command` either

mifrandir: actually, you might want to use https://en.cppreference.com/w/cpp/language/structured_binding here and below

mifrandir:
> Thoughts on JSON config to enable/disable commands vs other forms of configs (idk cmdline args? flags?)

JSON seems to be the most versatile and also quite easy to implement; you may then ask people to specify exactly what config they ran their tests with which also makes sense.

lib/debug/repl.cpp
mifrandir: You're right, you can't!

kshxtij:
Will probably change config to

```json
{
  "debugger": {
    "io": {
      "adc": {
        "enabled": true,
        "pins": []
      },
      "i2c": {
        "enabled": true,
        "channels": []
      }
    }
  }
}
```

and refactor `fromFile` to use `parseIO`, `parseSensors`, `parseMotors` etc to avoid one massive function

kshxtij:
Ubuntu build works if i add `target_link_libraries(${target} hyped_io)` to lib/debug's CMake but im not sure why i need that 🤔

src/debugger/main.cpp
```diff
@@ -0,0 +1,22 @@
+#include <core/logger.hpp>
+#include <core/wall_clock.hpp>
+#include <debug/repl.hpp>
+
+int main(int argc, char **argv)
```
maxguy2001, line 5: const

maxguy2001 (changes requested):
doesn't build linux but other than that, all good.

kshxtij:
TODO: maybe abstract away json parsing from the `fromFile` function into `parse` for more unit testability

lib/debug/repl.cpp
```diff
@@ -0,0 +1,213 @@
...
+             "Missing required field 'io.i2c.enabled' in configuration file");
+    return std::nullopt;
+  }
+  if (i2c["enabled"].GetBool()) {
+    if (!i2c.HasMember("channels")) {
+      log_.log(hyped::core::LogLevel::kFatal,
+               "Missing required field 'io.i2c.channels' in configuration file");
+      return std::nullopt;
+    }
+    auto channels = i2c["channels"].GetArray();
```
mifrandir, line 100: `const`?

lib/debug/repl.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+void Repl::addHelpCommand()
+{
+  addCommand(Command{"help", "Print this help message", [this]() { printCommands(); }});
+}
+
+void Repl::addAdcCommands(const uint8_t pin)
+{
+  const auto adc = std::make_shared<hyped::io::Adc>(pin, log_);
+  {
```
mifrandir, line 148: I don't think you need this scope.

mifrandir (changes requested):

lib/debug/repl.cpp
```diff
@@ -0,0 +1,213 @@
...
+}
+
+void Repl::addHelpCommand()
+{
+  addCommand(Command{"help", "Print this help message", [this]() { printCommands(); }});
+}
+
+void Repl::addAdcCommands(const uint8_t pin)
+{
+  const auto adc = std::make_shared<hyped::io::Adc>(pin, log_);
```
mifrandir, line 147: You are using `io` here. So surely you need to link against it?

mifrandir (approved):

maxguy2001 (approved):

lib/debug/repl.cpp
```diff
@@ -0,0 +1,211 @@
...
+    if (value) { log_.log(hyped::core::LogLevel::kInfo, "ADC value: %d", *value); }
+  };
+  addCommand(adc_read_command);
+}
+
+void Repl::addI2cCommands(const uint8_t channel)
+{
+  const auto i2c = std::make_shared<hyped::io::I2c>(channel, log_);
+  {
+    Command i2c_read_command;
```
SnickeyX, line 166: not a change rn, just an fyi: there will have to be a separate readReg and read command once my pr is merged

SnickeyX (changes requested):

lib/debug/repl.cpp
kshxtij: Sure, we can add commands as the interface changes.

SnickeyX (changes requested):

config/template.json
```diff
@@ -0,0 +1,14 @@
+{
+	"debugger": {
+		"io": {
+			"adc": {
+				"enabled": false,
+				"pins": []
+			},
+			"i2c": {
+				"enabled": true,
+				"buses": [12]
```
SnickeyX, line 10: 12? everything else looks okay

SnickeyX (changes requested):

SnickeyX (approved):

PR #31, opened 2022-11-03
SNS - Accelerometer

SnickeyX:
@Platiniom64 
- [x] Implement `read`, `write`  
- [x] Confirm code works by testing with accelerometer


lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,28 @@
...
+  public:
+    Accelerometer(io::I2c &i2c, hyped::core::ILogger &log);
+    ~Accelerometer();
+
+    bool configure(); 
+    std::optional<float> read();
+
+  private:
+    hyped::core::ILogger &log_; 
+    hyped::io::I2c &i2c_; // I2c object for the temperature sensor
```
kshxtij, line 24: Unnecessary comment, even then accelerometer*

maxguy2001 (approved):

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,27 @@
...
+{
+}
+
+Accelerometer::~Accelerometer()
+{
+}
+
+std::optional<float> Accelerometer::read()
+{
+  // TODOLater: read from i2c
```
kshxtij, line 15: TODO for this and below

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include "sensors.hpp"
+
+#include <cstdint>
+#include <cstdio>
```
mifrandir, line 6: Unused import; move to cpp if it's used there.

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,27 @@
...
+#include <optional>
+
+#include <core/logger.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+
+class Accelerometer : II2cSensor<float> {
```
mifrandir, line 14: `core::Float`

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include "sensors.hpp"
+
+#include <cstdint>
```
mifrandir, line 5: Unused import, remove.

mifrandir (changes requested):

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,54 @@
...
+}
+
+Accelerometer::~Accelerometer()
+{
+}
+
+    std::optional<float> Accelerometer::read() 
+    {
+        
+        do
```
mifrandir: don't use do-while loops

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,54 @@
...
+			/*Wait till the value is ready to read*/
+			ITDS_getdataReadyState(&DRDY);
+		} while (DRDY == ITDS_disable);
+
+		ITDS_getRawAccelerationX(&XRawAcc);
+		XRawAcc = XRawAcc >> 2; /* shifted by 2 as 14bit resolution is used in high performance mode */
+		float XAcceleration = (float) (XRawAcc);
+		XAcceleration = XAcceleration / 1000; /* mg to g */
+		XAcceleration = XAcceleration * 1.952; /* Multiply with sensitivity 1.952 in high performance mode, 14bit, and full scale +-16g */
+		printf("Acceleration X-axis %f g \r\n ", XAcceleration);
```
mifrandir, line 75: don't print, log

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,54 @@
...
+		printf("Acceleration Y-axis %f g \r\n ", YAcceleration);
+
+		ITDS_getRawAccelerationZ(&ZRawAcc);
+		ZRawAcc = ZRawAcc >> 2;
+		float ZAcceleration = (float) (ZRawAcc);
+		ZAcceleration = ZAcceleration / 1000;
+		ZAcceleration = ZAcceleration * 1.952;
+		printf("Acceleration Z-axis %f g \r\n ", ZAcceleration);
+
+
```
mifrandir: missing return value

mifrandir (changes requested):

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,40 @@
...
+  Accelerometer(const std::uint8_t mux_address, io::I2c &i2c, hyped::core::ILogger &log);
+  ~Accelerometer();
+
+  bool configure();
+  std::optional<g_result> read();
+
+ private:
+  hyped::core::ILogger &log_;
+  hyped::io::I2c &i2c_;  // I2c object for the accelerometer
+  const std::unit8_t ADDR;
```
kshxtij, line 33: address_

kshxtij (changes requested):

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include <cstdio>
+#include <optional>
+
+#include <core/logger.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+
+
+struct g_result {
```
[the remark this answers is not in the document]
SnickeyX: part of `types.hpp` in core.  @maxguy2001 what struct name would you like for the accelerometer?

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,102 @@
...
+  return rawX;
+}
+
+std::optional<std::int16_t> getRawAccelerationZ()
+{
+  std::optional<std::uint8_t> resultLow  = i2c_.readByte(ADDR, 0x2C);
+  std::optional<std::uint8_t> resultHigh = i2c_.readByte(ADDR, 0x2D);
+
+  if (!resultLow.hasValue() || !resultLow.hasValue()) { return std::nullopt; }
```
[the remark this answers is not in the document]
Platiniom64: that error will already be logged, by the i2c class that does hte reading.

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+  return rawX;
+}
+
+std::optional<std::int16_t> getRawAccelerationY()
+{
+  std::optional<std::uint8_t> resultLow  = i2c_.readByte(ADDR, 0x2A);
+  std::optional<std::uint8_t> resultHigh = i2c_.readByte(ADDR, 0x2B);
+
+  if (!resultLow.hasValue() || !resultLow.hasValue()) { return std::nullopt; }
```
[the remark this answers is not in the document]
Platiniom64: that error will already be logged, by the i2c class that does hte reading.

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+  if (!result.has_value()) { return std::nullopt; }
+
+  // check if sensor says that data is ready to be read
+  if (result.value() == 0x00) {
+    log_.log(core::LogLevel::kFatal, "acceleration data not ready yet to be read");
+    return std::nullopt;
+  }
+
+  // x axis
+  std::optional<std::int16_t> resultX = getRawAccelerationX();
```
SnickeyX, line 66: `const` here and for every other result too, also `result_X` but open to better naming

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+}
+
+Accelerometer::~Accelerometer()
+{
+}
+
+std::optional<std::int16_t> Accelerometer::getRawAccelerationX()
+{
+  std::optional<std::uint8_t> resultLow  = i2c_.readByte(address_, 0x28);
+  std::optional<std::uint8_t> resultHigh = i2c_.readByte(address_, 0x29);
```
SnickeyX, line 19: also, for the addresses of all the registers we read from - could you `static constexpr` them with just the actual register name, e.g `X_OUT_L = 0x28`. Just so we don't have any magic numbers in the code and the register names can easily be cross-referenced with the datasheet.

lib/sensors/accelerometer.cpp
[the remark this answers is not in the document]
mifrandir: but that class doesn't know what was being read, does it? so here you should specify that there was a failure in reading an acceleration value!

lib/core/types.hpp
```diff
@@ -40,4 +41,23 @@ using ImuData     = std::array<Float, kNumImus>;
 using EncoderData = std::array<std::uint32_t, kNumEncoders>;
 using KeyenceData = std::array<std::uint32_t, kNumKeyence>;
 
+// data produced by the accelerometer sensor
+struct acceleration_struct {
```
mifrandir, line 45: Naming.

1. Upper, camel-case.
2. Don't call things struct, int, float, array, or any other thing that is obvious from the declaration.

Just `Acceleration` should do fine. Actually, in line with the others, I'd call it `RawAccelerationData`.

lib/core/types.hpp
```diff
@@ -40,4 +41,23 @@ using ImuData     = std::array<Float, kNumImus>;
...
+        y(y),
+        z(z),
+        time(time)
+  {
+  }
+
+  core::Float x;
+  core::Float y;
+  core::Float z;
+  std::chrono::high_resolution_clock::time_point time;
```
mifrandir, line 60: just `TimePoint` from `core/time.hpp`

lib/sensors/accelerometer.cpp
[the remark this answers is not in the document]
mifrandir: `auto` (please fix everywhere)

mifrandir: `kXOutL` for style reasons

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+}
+
+std::optional<std::int16_t> Accelerometer::getRawAccelerationZ()
+{
+  std::optional<std::uint8_t> resultLow  = i2c_.readByte(address_, 0x2C);
+  std::optional<std::uint8_t> resultHigh = i2c_.readByte(address_, 0x2D);
+
+  if (!resultLow.has_value() || !resultLow.has_value()) { return std::nullopt; }
+
+  std::optional<std::int16_t> rawX = (resultHigh.value() << 8) | resultLow.value();
```
mifrandir, line 47: `const auto` (here and in other places)

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+
```
mifrandir, line 48: whitespace; here and elsewhere; don't have empty lines in functions. if you think it's required for structure then you should split your function

mifrandir (changes requested):
not a proper review yet, just some style things

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
+#include "accelerometer.hpp"
+
+namespace hyped::sensors {
+
+Accelerometer::Accelerometer(std::uint8_t mux_address_ess, io::I2c &i2c, core::ILogger &log)
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+Accelerometer::~Accelerometer()
+{
+}
+
+std::optional<std::int16_t> Accelerometer::getRawAccelerationX()
+{
+  std::optional<std::uint8_t> resultLow  = i2c_.readByte(address_, 0x28);
+  std::optional<std::uint8_t> resultHigh = i2c_.readByte(address_, 0x29);
+
+  if (!resultLow.has_value() || !resultLow.has_value()) { return std::nullopt; }
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  return rawX;
+}
+
+std::optional<std::int16_t> Accelerometer::getRawAccelerationY()
+{
+  std::optional<std::uint8_t> resultLow  = i2c_.readByte(address_, 0x2A);
+  std::optional<std::uint8_t> resultHigh = i2c_.readByte(address_, 0x2B);
+
+  if (!resultLow.has_value() || !resultLow.has_value()) { return std::nullopt; }
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  return rawX;
+}
+
+std::optional<std::int16_t> Accelerometer::getRawAccelerationZ()
+{
+  std::optional<std::uint8_t> resultLow  = i2c_.readByte(address_, 0x2C);
+  std::optional<std::uint8_t> resultHigh = i2c_.readByte(address_, 0x2D);
+
+  if (!resultLow.has_value() || !resultLow.has_value()) { return std::nullopt; }
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+}
+
+std::optional<std::int16_t> Accelerometer::getRawAccelerationX()
+{
+  std::optional<std::uint8_t> resultLow  = i2c_.readByte(address_, 0x28);
+  std::optional<std::uint8_t> resultHigh = i2c_.readByte(address_, 0x29);
+
+  if (!resultLow.has_value() || !resultLow.has_value()) { return std::nullopt; }
+
+  std::optional<std::int16_t> rawX = (resultHigh.value() << 8) | resultLow.value();
```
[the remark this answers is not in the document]
Platiniom64: I think we still need the optional. Cause the reading of the low and high byte could fail, so we need to know when that happens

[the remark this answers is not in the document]
Platiniom64: changed the variable name

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+std::optional<std::int16_t> Accelerometer::getRawAccelerationY()
+{
+  std::optional<std::uint8_t> resultLow  = i2c_.readByte(address_, 0x2A);
+  std::optional<std::uint8_t> resultHigh = i2c_.readByte(address_, 0x2B);
+
+  if (!resultLow.has_value() || !resultLow.has_value()) { return std::nullopt; }
+
+  std::optional<std::int16_t> rawX = (resultHigh.value() << 8) | resultLow.value();
+
+  return rawX;
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+std::optional<std::int16_t> Accelerometer::getRawAccelerationZ()
+{
+  std::optional<std::uint8_t> resultLow  = i2c_.readByte(address_, 0x2C);
+  std::optional<std::uint8_t> resultHigh = i2c_.readByte(address_, 0x2D);
+
+  if (!resultLow.has_value() || !resultLow.has_value()) { return std::nullopt; }
+
+  std::optional<std::int16_t> rawX = (resultHigh.value() << 8) | resultLow.value();
+
+  return rawX;
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+}
+
+std::optional<core::acceleration_struct> Accelerometer::read()
+{
+  /* check to see if the values are ready to be read */
+  std::optional<std::uint8_t> result = i2c_.readByte(address_, 0x27);
+
+  if (!result.has_value()) { return std::nullopt; }
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  // check if sensor says that data is ready to be read
+  if (result.value() == 0x00) {
+    log_.log(core::LogLevel::kFatal, "acceleration data not ready yet to be read");
+    return std::nullopt;
+  }
+
+  // x axis
+  std::optional<std::int16_t> resultX = getRawAccelerationX();
+  if (!resultX.has_value()) return std::nullopt;
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  std::int16_t XRawAcc
+    = resultX.value() >> 2; /* shifted by 2 as 14bit resolution is used in high performance mode */
+  float XAcceleration = (float)(XRawAcc);
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+  // z axis
+  std::optional<std::int16_t> resultZ = getRawAccelerationZ();
+  if (!resultZ.has_value()) return std::nullopt;
+
+  std::int16_t ZRawAcc = resultZ.value() >> 2;
+  float ZAcceleration  = (float)(ZRawAcc);
+  ZAcceleration        = ZAcceleration / 1000;
+  ZAcceleration        = ZAcceleration * 1.952;
+
+  std::optional<core::acceleration_struct> resultNEW{std::in_place,
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,117 @@
...
+{
+  core::Result result1 = i2c_.writeByteToRegister(address_, CTRL_1, 0x64);
+  if (result1 == core::Result::kFailure) return core::Result::kFailure;
+
+  core::Result result2 = i2c_.writeByteToRegister(address_, CTRL_2, 0xC);
+  if (result2 == core::Result::kFailure) return core::Result::kFailure;
+
+  core::Result result3 = i2c_.writeByteToRegister(address_, CTRL_6, 0x30);
+  if (result3 == core::Result::kFailure) return core::Result::kFailure;
+
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,39 @@
...
+  Accelerometer(const std::uint8_t mux_address, io::I2c &i2c, core::ILogger &log);
+  ~Accelerometer();
+
+  core::Result configure();
+  std::optional<core::acceleration_struct> read();
+
+ private:
+  const std::uint8_t CTRL_1 = 0x20;
+  const std::uint8_t CTRL_2 = 0x21;
+  const std::uint8_t CTRL_6 = 0x25;
```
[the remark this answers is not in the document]
Platiniom64: done

SnickeyX:
also run clang-format before pushing since you're failing the formatting test

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,43 @@
...
+  ~Accelerometer();
+
+  core::Result configure();
+  std::optional<core::acceleration_struct> read();
+
+ private:
+
+  core::ILogger &log_;
+  io::I2c &i2c_;  // I2c object for the accelerometer
+  const std::uint8_t address_;
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include "core/types.hpp"
+#include <core/logger.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+
+static constexpr std::uint8_t CTRL_1 = 0x20;
+static constexpr std::uint8_t CTRL_2 = 0x21;
+static constexpr std::uint8_t CTRL_6 = 0x25;
+
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,43 @@
...
+  Accelerometer(const std::uint8_t mux_address, io::I2c &i2c, core::ILogger &log);
+  ~Accelerometer();
+
+  core::Result configure();
+  std::optional<core::acceleration_struct> read();
+
+ private:
+
+  core::ILogger &log_;
+  io::I2c &i2c_;  // I2c object for the accelerometer
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,43 @@
+#pragma once
+
+#include "i2c_sensors.hpp"
+
+#include <unistd.h>
+
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,156 @@
...
+Accelerometer::~Accelerometer()
+{
+}
+
+std::optional<std::int16_t> Accelerometer::getRawAccelerationX()
+{
+  auto low_byte  = i2c_.readByte(address_, 0x28);
+  if (!low_byte.has_value()) {
+    log_.log(core::LogLevel::kFatal, "could not read register of low value of X axis for accelerometer");
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,156 @@
...
+}
+
+Accelerometer::~Accelerometer()
+{
+}
+
+std::optional<std::int16_t> Accelerometer::getRawAccelerationX()
+{
+  auto low_byte  = i2c_.readByte(address_, 0x28);
+  if (!low_byte.has_value()) {
```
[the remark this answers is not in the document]
Platiniom64: done

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,156 @@
...
+{
+}
+
+Accelerometer::~Accelerometer()
+{
+}
+
+std::optional<std::int16_t> Accelerometer::getRawAccelerationX()
+{
+  auto low_byte  = i2c_.readByte(address_, 0x28);
```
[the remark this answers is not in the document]
Platiniom64: done

Platiniom64:
all done.
Though the sensor doesn't work still, so maybe the issue is with the code. Don't merge the branch yet !!

lib/core/types.hpp
```diff
@@ -40,4 +43,20 @@ using ImuData     = std::array<Float, kNumImus>;
 using EncoderData = std::array<std::uint32_t, kNumEncoders>;
 using KeyenceData = std::array<std::uint32_t, kNumKeyence>;
 
+// data produced by the accelerometer sensor
+struct RawAccelerationData {
+  RawAccelerationData(core::Float x, core::Float y, core::Float z, TimePoint time)
```
mifrandir, line 48: don't declare the constructor in the header; make a cpp file if that's what it takes

also `const`

then again, i don't think we need a constructor here

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+std::optional<std::int16_t> Accelerometer::getRawAccelerationX()
+{
+  const auto low_byte = i2c_.readByte(mux_address_, kXOutL);
+  if (!low_byte) {
+    log_.log(core::LogLevel::kFatal,
+             "Failed to read the low byte for acceleration along the x-axis");
+    return std::nullopt;
+  }
+
```
mifrandir, line 33: empty line; here and elsewhere

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,167 @@
...
+  const core::Result ctrl1Result = i2c_.writeByteToRegister(mux_address_, kCTRL1, 0x64);
+  if (ctrl1Result == core::Result::kFailure) return core::Result::kFailure;
+
+  /* Enable block data update */
+  /* Enable address auto increment */
+  const core::Result ctrl2Result = i2c_.writeByteToRegister(mux_address_, kCTRL2, 0xC);
+  if (ctrl2Result == core::Result::kFailure) return core::Result::kFailure;
+
+  /* Full scale +-16g */
+  /* Filter bandwidth = ODR/2 */
```
mifrandir, line 160: stick with `//` comments here; `/* ... */` is usually more for docs in the header (no clear rule, just preference)

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,167 @@
...
+  const std::optional<core::RawAccelerationData> rawAcceleration{
+    std::in_place, XAcceleration, YAcceleration, ZAcceleration, std::chrono::system_clock::now()};
+
+  return rawAcceleration;
+}
+
+core::Result Accelerometer::configure()
+{
+  // if device id is incorrect, then the address received is wrong
+  const auto deviceID = i2c_.readByte(mux_address_, kDevId);
```
mifrandir, line 137: `deviceId`

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,53 @@
...
+static constexpr std::uint8_t kCTRL2        = 0x21;
+static constexpr std::uint8_t kCTRL6        = 0x25;
+static constexpr std::uint8_t kXOutL        = 0x28;
+static constexpr std::uint8_t kXOutH        = 0x29;
+static constexpr std::uint8_t kYOutL        = 0x2A;
+static constexpr std::uint8_t kYOutH        = 0x2B;
+static constexpr std::uint8_t kZOutL        = 0x2C;
+static constexpr std::uint8_t kZOutH        = 0x2D;
+static constexpr std::uint8_t kDataReady    = 0x27;
+static constexpr std::uint8_t kDevId        = 0x0F;
```
mifrandir, line 27: `kDeviceId`

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,53 @@
...
+static constexpr std::uint8_t expectedDevId = 0x44;
+
+class Accelerometer : II2cMuxSensor<core::RawAccelerationData> {
```
mifrandir, line 30: `class Accelerometer : public ...`

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,53 @@
...
+ public:
+  Accelerometer(const std::uint8_t mux_address,
+                const std::uint8_t channel,
+                io::I2c &i2c,
+                core::ILogger &log);
+  ~Accelerometer();
+
+  core::Result configure();
+  std::optional<core::RawAccelerationData> read();
+  std::uint8_t getChannel();
```
mifrandir, line 40: all these need to be virtual

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+  /* Enable block data update */
+  /* Enable address auto increment */
+  const core::Result ctrl2Result = i2c_.writeByteToRegister(mux_address_, kCTRL2, 0xC);
+  if (ctrl2Result == core::Result::kFailure) return core::Result::kFailure;
+
+  /* Full scale +-16g */
+  /* Filter bandwidth = ODR/2 */
+  const core::Result ctrl3Result = i2c_.writeByteToRegister(mux_address_, kCTRL6, 0x30);
+  if (ctrl3Result == core::Result::kFailure) return core::Result::kFailure;
```
mifrandir, line 162: braces around code blocks, always.

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+  // z axis
+  auto resultZ = getRawAccelerationZ();
+  if (!resultZ) return std::nullopt;
+  const std::int16_t ZRawAcc = resultZ.value() >> 2;
+  core::Float ZAcceleration  = static_cast<core::Float>(ZRawAcc);
+  ZAcceleration              = ZAcceleration / 1000;
+  ZAcceleration              = ZAcceleration * 1.952;
+
+  const std::optional<core::RawAccelerationData> rawAcceleration{
```
mifrandir, line 128: this needn't be optional; just build the struct

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,167 @@
...
+  // y axis
+  auto resultY = getRawAccelerationY();
+  if (!resultY) return std::nullopt;
+  const std::int16_t YRawAcc = resultY.value() >> 2;
+  core::Float YAcceleration  = static_cast<core::Float>(YRawAcc);
+  YAcceleration              = YAcceleration / 1000;
+  YAcceleration              = (YAcceleration * 1.952);
+
+  // z axis
+  auto resultZ = getRawAccelerationZ();
```
mifrandir, line 121: `const`

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,167 @@
...
+  if (!resultZ) return std::nullopt;
+  const std::int16_t ZRawAcc = resultZ.value() >> 2;
```
mifrandir, line 123: variables are lower snake case; i.e. `z_raw_acceleration`; I'd improve this to `raw_z_acceleration`.

mifrandir (changes requested):

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,167 @@
...
+Accelerometer::~Accelerometer()
+{
+}
+
+std::uint8_t Accelerometer::getChannel()
+{
+  return channel_;
+}
+
+std::optional<std::int16_t> Accelerometer::getRawAccelerationX()
```
maxguy2001, line 25: This setup seems very verbose? Unless I'm missing something, the only real difference between getting the raw accelerations for x, y, and z is the register address you read (and the error message but that doesn't really affect functionality)? Therefore would it not make sense to pass that in as a function argument for a single function (e.g. `getRawAccelerationAxis(register_address)`) or similar. Then you would only have to write one function?

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,167 @@
...
+    log_.log(core::LogLevel::kFatal, "acceleration data could not be read");
+    return std::nullopt;
+  }
+  if (dataReady.value() == 0x00) {
+    log_.log(core::LogLevel::kFatal, "acceleration data not ready yet to be read");
+    return std::nullopt;
+  }
+
+  // x axis
+  auto resultX = getRawAccelerationX();
```
maxguy2001, line 103: `const`

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,167 @@
...
+  if (!resultX) return std::nullopt;
+  const std::int16_t XRawAcc
+    = resultX.value() >> 2; /* shifted by 2 as 14bit resolution is used in high performance mode */
+  core::Float XAcceleration = static_cast<core::Float>(XRawAcc);
+  XAcceleration             = XAcceleration / 1000; /* mg to g */
+  XAcceleration = XAcceleration * 1.952; /* Multiply with sensitivity 1.952 in high performance
+                                            mode, 14bit, and full scale +-16g */
+
+  // y axis
+  auto resultY = getRawAccelerationY();
```
maxguy2001, line 113: `const`

maxguy2001 (changes requested):
A few structural changes for now. From nav side, this looks ok.

lib/sensors/accelerometer.cpp
mifrandir: No, the user should definitely not have to pass in the correct register address.

Having a private function like that seems fine, though.

Maybe there's a better approach here? Not that it's too important but one idea I'd have is making an enum for X, Y, and Z and then passing in that value? The function can then convert the enum into register values and do the work? This means the caller isn't required to pass magic register addresses. There are problems with this; perhaps you have better ideas?

maxguy2001: Maybe not a register address then but some way of selecting which axis to read from. Defining 3 separate functions with only trivial changes between them seems quite pointless?

SnickeyX: maybe like a `getRawAccelerationHelper` (naming open to suggestions ofc) taking in two enum values for low and high byte (+1 for axis to use in log messages?)

SnickeyX:
@Platiniom64 similar changes to the temperature sensors here as well, we don't need the `mux_address` pararameter anymore. Just use the default address of the accelerometer sensors instead (another static constexpr) - the address is somewhere in the datasheet.

lib/debug/repl.cpp
```diff
@@ -56,7 +56,15 @@ std::optional<std::unique_ptr<Repl>> Repl::fromFile(const std::string &path)
                 path.c_str());
     return std::nullopt;
   }
-  const auto io = debugger["io"].GetObject();
+  const auto io      = debugger["io"].GetObject();
+
```
mifrandir, line 60: whitespace

lib/debug/repl.cpp
```diff
@@ -123,6 +131,30 @@ std::optional<std::unique_ptr<Repl>> Repl::fromFile(const std::string &path)
       repl->addI2cCommands(bus.GetUint());
     }
   }
+
```
mifrandir, line 134: whitespace; more later

lib/debug/repl.cpp
```diff
@@ -308,4 +340,46 @@ void Repl::addSpiCommands(const std::uint8_t bus)
     addCommand(spi_write_byte_command);
   }
 }
+
+void Repl::addAccelerometerCommands(const std::uint8_t bus, const std::uint8_t device_address)
+{
+  //const auto i2c = std::make_shared<hyped::io::HardwareI2c>(bus, logger_);
```
mifrandir, line 346: ...

lib/sensors/accelerometer.cpp
```diff
@@ -38,7 +38,7 @@ std::optional<std::int16_t> Accelerometer::getRawAccelerationX()
     return std::nullopt;
   }
 
-  const auto raw_acceleration_X = (high_byte.value() << 8) | low_byte.value();
+  const auto raw_acceleration_X = (int16_t) ((high_byte.value() << 8) | low_byte.value());
```
mifrandir, line 41: `static_cast<std::int16_t>`

mifrandir (changes requested):

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+namespace hyped::sensors {
+
+Accelerometer::Accelerometer(const std::uint8_t kDeviceAddr,
+                             const std::uint8_t channel,
+                             io::HardwareI2c &i2c,
+                             core::ILogger &log)
+    : kDeviceAddr(kDeviceAddr),
+      channel_(channel),
+      log_(log),
```
TomLonergan03, line 11: should be `logger_(logger)` to align with the rest of the repo

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  const auto high_byte = i2c_.readByte(kDeviceAddr, kXOutH);
+  if (!high_byte) {
+    log_.log(core::LogLevel::kFatal,
+             "Failed to read the high byte for acceleration along the x-axis");
+    return std::nullopt;
+  }
+
+  const auto raw_acceleration_X = (int16_t) ((high_byte.value() << 8) | low_byte.value());
```
TomLonergan03, line 41: `raw_acceleration_x` i think, similar for y and z later

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include "core/types.hpp"
+#include <core/logger.hpp>
+#include <io/i2c.hpp>
+#include <io/hardware_i2c.hpp>
+
+namespace hyped::sensors {
+
+static constexpr std::uint8_t kCTRL1Addr        = 0x20;
+static constexpr std::uint8_t kCTRL1Value       = 0x64;
```
TomLonergan03, line 19: what is CTRL?

TomLonergan03 (changes requested):

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,159 @@
+#include "accelerometer.hpp"
+
+namespace hyped::sensors {
+
+Accelerometer::Accelerometer(const std::uint8_t kDeviceAddr,
+                             const std::uint8_t channel,
+                             io::HardwareI2c &i2c,
+                             core::ILogger &log)
+    : kDeviceAddr(kDeviceAddr),
```
TomLonergan03, line 9: also this feels weird, defining a constant in the constructor. idk if it goes against the style guide but i haven't seen it elsewhere

lib/sensors/accelerometer.cpp
```diff
@@ -2,11 +2,11 @@
 
 namespace hyped::sensors {
 
-Accelerometer::Accelerometer(const std::uint8_t mux_address,
+Accelerometer::Accelerometer(const std::uint8_t kDeviceAddr,
```
mifrandir, line 5: spell it out, `Address`

lib/sensors/accelerometer.cpp
mifrandir: drop the `Helper` and make it private

lib/sensors/accelerometer.cpp
```diff
@@ -85,37 +85,37 @@ std::optional<std::int16_t> Accelerometer::getRawAccelerationZ()
...
+    log_.log(core::LogLevel::kInfo, "acceleration data not ready yet to be read");
     return std::nullopt;
   }
 
   // x axis
   auto resultX = getRawAccelerationX();
   if (!resultX) return std::nullopt;
   const std::int16_t XRawAcc = resultX.value(); 
-  core::Float XAcceleration = (static_cast<core::Float>(XRawAcc)) * 0.488f;
+  int16_t XAcceleration = (int16_t) ((((int32_t) XRawAcc) * 488) / 1000);
```
mifrandir, line 106: `static_cast` was good

mifrandir: also magic numbers

lib/sensors/accelerometer.hpp
```diff
@@ -15,22 +15,32 @@
...
-static constexpr std::uint8_t kCTRL2        = 0x21;
-static constexpr std::uint8_t kCTRL6        = 0x25;
+static constexpr std::uint8_t kCTRL1Addr        = 0x20;
+static constexpr std::uint8_t kCTRL1Value       = 0x64;
+
+static constexpr std::uint8_t kCTRL2Addr        = 0x21;
+static constexpr std::uint8_t kCTRL2Value        = 0x0;
+
+static constexpr std::uint8_t kCTRL6Addr        = 0x25;
+static constexpr std::uint8_t kCTRL6Value        = 0x30;
```
mifrandir, line 25: `kCtrl6Value`

mifrandir (changes requested):

lib/sensors/accelerometer.hpp
SnickeyX: name of register, @Platiniom64 reckon its worth mentioning that the names are taken directly from the datasheet

lib/sensors/accelerometer.cpp
Platiniom64: turns out the static cast to a float was the whole reason why the accelerometer didn't work. Hence, I am sticking to int here.

SnickeyX:
@Platiniom64 fix the macos break by changing all instances of `sprintf` to `snprintf`, just have to also add the size of the buffer as the second parameter. also merge conflicts!

lib/sensors/accelerometer.cpp
mifrandir: i don't care what you static cast to, but you should static cast!

`static_cast<std::int16_t>(...)`

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,143 @@
...
+    logger_.log(core::LogLevel::kFatal, "acceleration data could not be read");
+    return std::nullopt;
+  }
+  if (dataReady.value() % 2 == 0) {
+    logger_.log(core::LogLevel::kInfo, "acceleration data not ready yet to be read");
+    return std::nullopt;
+  }
+
+  // x axis
+  const auto resultX = getRawAcceleration(Axis::x);
```
mifrandir, line 85: the comment seems very redundant

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,143 @@
...
+  if (!resultX) return std::nullopt;
```
mifrandir, line 86: put braces around

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,143 @@
...
+  const std::optional<core::RawAccelerationData> Acceleration3D{
+    std::in_place, XAcceleration, YAcceleration, ZAcceleration, std::chrono::system_clock::now()};
+
+  return Acceleration3D;
+}
+
+core::Result Accelerometer::configure()
+{
+  // check we are communicating with the correct sensor
+  const auto deviceID = i2c_.readByte(kDeviceAddress, kDevId);
```
mifrandir, line 111: device_id

mifrandir (changes requested):
i won't point out all the style guide violations, especially for local variables

Platiniom64:
accelerometer is back to broken, I still have to fix that...

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,145 @@
+#include "accelerometer.hpp"
+
+namespace hyped::sensors {
+
+Accelerometer::Accelerometer(std::uint8_t channel, io::HardwareI2c &i2c, core::ILogger &logger)
```
TomLonergan03, line 5: `const` for channel

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,145 @@
...
+    = static_cast<std::int16_t>((static_cast<int32_t>(rawAcc) * 488) / 1000);
+
+  return acceleration;
+}
+
+// todolater: check whether the range of values the accelerometer is correct, cause now the noise is
+std::optional<core::RawAccelerationData> Accelerometer::read()
+{
+  // check to see if the values are ready to be read
+  auto data_ready = i2c_.readByte(kDeviceAddress, kDataReady);
```
TomLonergan03, line 75: `const`

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,64 @@
...
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+
+enum Axis { x, y, z };
+static const char *AxisStrings[] = {"x-axis", "y-axis", "z-axis"};
+
+static constexpr std::uint8_t kDeviceAddress = 0x19;
+
+static constexpr std::uint8_t kCtrl1Addr  = 0x20;
```
TomLonergan03, line 23: `kCtrl1Addr`, similar for ctrl2 and 6

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+static constexpr std::uint8_t kYOutL = 0x2A;
+static constexpr std::uint8_t kYOutH = 0x2B;
+
+static constexpr std::uint8_t kZOutL = 0x2C;
+static constexpr std::uint8_t kZOutH = 0x2D;
+
+static constexpr std::uint8_t kDataReady = 0x27;
+
+static constexpr std::uint8_t kDevId        = 0x0F;
```
TomLonergan03, line 43: `kDeviceId`

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,64 @@
...
+static constexpr std::uint8_t expectedDevId = 0x44;
```
TomLonergan03, line 44: `kExpectedDeviceId`

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,64 @@
...
+static constexpr std::uint8_t kCtrl1Value = 0x64;
+
+static constexpr std::uint8_t kCtrl2Addr  = 0x21;
+static constexpr std::uint8_t kCtrl2Value = 0x0;
+
+static constexpr std::uint8_t kCtrl6Addr  = 0x25;
+static constexpr std::uint8_t kCtrl6Value = 0x30;
+
+static constexpr std::uint8_t kXOutL = 0x28;
+static constexpr std::uint8_t kXOutH = 0x29;
```
TomLonergan03, line 33: `kXOutLow` and `kXOutHigh`, same for other axes

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include "core/types.hpp"
+#include <core/logger.hpp>
+#include <io/hardware_i2c.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+
+enum Axis { x, y, z };
+static const char *AxisStrings[] = {"x-axis", "y-axis", "z-axis"};
```
TomLonergan03, line 19: `std::array<std::String, 3>`
no raw pointers please

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,64 @@
+#pragma once
+
+#include "i2c_sensors.hpp"
+
+#include <unistd.h>
```
TomLonergan03, line 5: not used as far as i can see

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,145 @@
...
+    return core::Result::kFailure;
+  }
+
+  // configure the sensor according to what was found in the repo of the manufacturers
+
+  // Sampling rate of 200 Hz
+  // Enable high performance mode
+  const core::Result ctrl1_result
+    = i2c_.writeByteToRegister(kDeviceAddress, kCtrl1Addr, kCtrl1Value);
+  if (ctrl1_result == core::Result::kFailure) return core::Result::kFailure;
```
TomLonergan03, line 129: `{return core::Result::kFailure}` always, here and below

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+  // Enable block data update
+  // Enable address auto increment
+  const core::Result ctrl2_result
+    = i2c_.writeByteToRegister(kDeviceAddress, kCtrl2Addr, kCtrl2Value);
+  if (ctrl2_result == core::Result::kFailure) return core::Result::kFailure;
+
+  // Full scale +-16g
+  // Filter bandwidth = ODR/2
+  const core::Result ctrl3_result
```
TomLonergan03, line 139: Is this meant to be `ctrl6_result`?

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,64 @@
...
+static constexpr std::uint8_t expectedDevId = 0x44;
+
+class Accelerometer : II2cMuxSensor<core::RawAccelerationData> {
+ public:
+  Accelerometer(const std::uint8_t channel, io::HardwareI2c &i2c, core::ILogger &logger);
+  ~Accelerometer();
+
+  virtual core::Result configure();
+  virtual std::optional<core::RawAccelerationData> read();
+  virtual std::uint8_t getChannel();
```
TomLonergan03, line 53: why are these virtual?

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,145 @@
...
+  // check to see if the values are ready to be read
+  auto data_ready = i2c_.readByte(kDeviceAddress, kDataReady);
+  if (!data_ready) {
+    logger_.log(core::LogLevel::kFatal, "acceleration data could not be read");
+    return std::nullopt;
+  }
+  if (data_ready.value() % 2 == 0) {
+    logger_.log(core::LogLevel::kInfo, "acceleration data not ready yet to be read");
+    return std::nullopt;
+  }
```
TomLonergan03, line 83: does it make a difference to the caller which of these events happen? in the first case it is `kFatal` while the second is `kInfo` which makes it seem like it does

if it does then the return values should differ

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,145 @@
...
+  if (!result_y) { return std::nullopt; }
+  const std::int16_t y_raw_acc      = result_y.value();
+  const std::int16_t y_acceleration = getAccelerationFromRaw(y_raw_acc);
+
+  const auto result_z = getRawAcceleration(Axis::z);
+  if (!result_z) { return std::nullopt; }
+  const std::int16_t z_raw_acc      = result_z.value();
+  const std::int16_t z_acceleration = getAccelerationFromRaw(z_raw_acc);
+
+  const std::optional<core::RawAccelerationData> acceleration_3axis{
```
TomLonergan03, line 100: does this need to be optional?

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include "i2c_sensors.hpp"
+
+#include <unistd.h>
+
+#include <cstdint>
+#include <cstdio>
+#include <optional>
+
+#include "core/types.hpp"
```
TomLonergan03, line 11: `<core/types.hpp>` i think, not entirely sure when `""` vs `<>` should be used

TomLonergan03 (changes requested):

lib/sensors/accelerometer.cpp
TomLonergan03: also i believe args are ordered from most to least "standard" ie logger, i2c, channel

lib/sensors/accelerometer.hpp
mifrandir: use `<>` in general, unless it's local (i.e. relative to the current file)

lib/core/types.hpp
```diff
@@ -40,4 +43,21 @@ using AccelerometerData    = std::array<Float, kNumAccelerometers>;
...
+        y(y),
+        z(z),
+        time(time)
+  {
+  }
+
+  std::int16_t x;
+  std::int16_t y;
+  std::int16_t z;
+  TimePoint time;
```
SnickeyX, line 60: `const` for all the variables

lib/debug/repl.cpp
```diff
@@ -58,6 +58,14 @@ std::optional<std::unique_ptr<Repl>> Repl::fromFile(const std::string &path)
...
   const auto io = debugger["io"].GetObject();
 
+  if (!debugger.HasMember("sensors")) {
+    logger_.log(core::LogLevel::kFatal,
+                "Missing required field 'debugger.sensors' in configuration file at %s",
+                path.c_str());
+    return std::nullopt;
+  }
+  const auto sensors = debugger["sensors"].GetObject();
+
```
SnickeyX, line 68: move this to be right above the `accelerometer` check, also remove empty space after

lib/debug/repl.cpp
```diff
@@ -170,6 +178,30 @@ std::optional<std::unique_ptr<Repl>> Repl::fromFile(const std::string &path)
...
     }
   }
+
+  if (!sensors.HasMember("accelerometer")) {
+    logger_.log(core::LogLevel::kFatal,
+                "Missing required field 'sensors.accelerometer' in configuration file");
+    return std::nullopt;
+  }
+  const auto accelerometer = sensors["accelerometer"].GetObject();
+
```
SnickeyX, line 188: there's empty lines like this which we want to avoid here and in `addAccelerometerCommands`, remove them

lib/debug/repl.cpp
```diff
@@ -427,6 +459,55 @@ void Repl::addSpiCommands(const std::uint8_t bus)
...
+      logger_.log(core::LogLevel::kInfo,
+                  "Acceleration in mg: \n x %d \n y %d \n z %d",
+                  result.x,
+                  result.y,
+                  result.z);
+    } else {
+      logger_.log(core::LogLevel::kFatal, "Failed to read accelerometer from bus %d", bus);
+    }
+  };
```
SnickeyX, line 506: to be consistent with the rest of the codebase, check for `!value` first

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,148 @@
+#include "accelerometer.hpp"
+
+namespace hyped::sensors {
+
+Accelerometer::Accelerometer(core::ILogger &logger,
+                             io::HardwareI2c &i2c,
+                             const std::uint8_t channel)
+    : channel_(channel),
+      logger_(logger),
+      i2c_(i2c)
```
SnickeyX, line 10: re-order to the same as the order of the arguements

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,65 @@
...
+  virtual std::optional<core::RawAccelerationData> read();
+  virtual std::uint8_t getChannel();
+
+ private:
+  core::ILogger &logger_;
+  io::HardwareI2c &i2c_;
+  const std::uint8_t channel_;
+
+  std::optional<std::int16_t> getRawAcceleration(Axis axis);
+  std::int16_t getAccelerationFromRaw(std::int16_t rawAcc);
```
SnickeyX, line 62: const for both axis and rawAcc, also change to rawAcceleration

SnickeyX: also it would be nice if these were in a separate `private:` as they are functions - check any of the i/o .hpp files if you are unsure

lib/sensors/accelerometer.hpp
SnickeyX: this ^^

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,65 @@
...
+static constexpr std::uint8_t kExpectedDeviceId = 0x44;
+
+class Accelerometer : II2cMuxSensor<core::RawAccelerationData> {
+ public:
+  Accelerometer(core::ILogger &logger, io::HardwareI2c &i2c, const std::uint8_t channel);
+  ~Accelerometer();
+
+  virtual core::Result configure();
+  virtual std::optional<core::RawAccelerationData> read();
+  virtual std::uint8_t getChannel();
```
SnickeyX, line 54: still virtual..

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,148 @@
...
+  if (!device_id) {
+    logger_.log(core::LogLevel::kFatal, "Failure to read device id of accelerometer");
+    return core::Result::kFailure;
+  }
+  if (device_id.value() != kExpectedDeviceId) {
+    logger_.log(core::LogLevel::kFatal, "Failure: accelerometer didn't give correct device id");
+    return core::Result::kFailure;
+  }
+
+  // configure the sensor according to what was found in the repo of the manufacturers
```
SnickeyX, line 126: this should be above the function in the .hpp file

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,148 @@
...
+      break;
+    case Axis::y:
+      low_byte_address  = kYOutLow;
+      high_byte_address = kYOutHigh;
+      break;
+    case Axis::z:
+      low_byte_address  = kZOutLow;
+      high_byte_address = kZOutHigh;
+      break;
+  }
```
SnickeyX, line 40: make this is a separate private function

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+  const auto raw_acceleration
+    = static_cast<std::int16_t>((high_byte.value() << 8) | low_byte.value());
+
+  return raw_acceleration;
+}
+
+std::int16_t Accelerometer::getAccelerationFromRaw(std::int16_t rawAcc)
+{
+  // these values come from the data sheet. Don't chance them.
```
SnickeyX, line 66: spelling for change

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,148 @@
...
+  std::int16_t acceleration
+    = static_cast<std::int16_t>((static_cast<int32_t>(rawAcc) * 488) / 1000);
+
+  return acceleration;
+}
+
+// todolater: check whether the range of values the accelerometer is correct, cause now the noise is
```
SnickeyX, line 73: TODOLater, also noise is ...?

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,148 @@
...
+std::optional<core::RawAccelerationData> Accelerometer::read()
+{
+  // check to see if the values are ready to be read
+  const auto data_ready = i2c_.readByte(kDeviceAddress, kDataReady);
+  if (!data_ready) {
+    logger_.log(core::LogLevel::kFatal, "acceleration data could not be read");
+    return std::nullopt;
+  }
+  // todolater: here the error is not that bad, hence the return value should indicate that
```
SnickeyX, line 82: TODOLater

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,148 @@
...
+  if (data_ready.value() % 2 == 0) {
+    logger_.log(core::LogLevel::kInfo, "acceleration data not ready yet to be read");
+    return std::nullopt;
+  }
+
+  const auto result_x = getRawAcceleration(Axis::x);
+  if (!result_x) { return std::nullopt; }
+  const std::int16_t x_raw_acc      = result_x.value();
```
SnickeyX, line 90: acc -> acceleration, same for the rest

SnickeyX (commented):
Lots of pedantic but necessary changes. Make sure you remove whitespaces within functions.

lib/core/types.hpp
```diff
@@ -40,4 +43,21 @@ using AccelerometerData    = std::array<Float, kNumAccelerometers>;
 using EncoderData          = std::array<std::uint32_t, kNumEncoders>;
 using KeyenceData          = std::array<std::uint32_t, kNumKeyence>;
 
+// data produced by the accelerometer sensor
+// !! the values are in mg !!
+struct RawAccelerationData {
+  RawAccelerationData(std::int16_t x, std::int16_t y, std::int16_t z, TimePoint time)
```
kshxtij, line 49: `const`s

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,153 @@
...
+    y_acceleration,
+    z_acceleration,
+    std::chrono::system_clock::now()};
+
+  return acceleration_3axis;
+}
+
+core::Result Accelerometer::configure()
+{
+  // check we are communicating with the correct sensor
```
kshxtij, line 122: Is this needed?

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+static constexpr std::uint8_t kDeviceId         = 0x0F;
+static constexpr std::uint8_t kExpectedDeviceId = 0x44;
+
+class Accelerometer : public II2cMuxSensor<core::RawAccelerationData> {
+ public:
+  Accelerometer(core::ILogger &logger, io::HardwareI2c &i2c, const std::uint8_t channel);
+  ~Accelerometer();
+
+  // configure the sensor according to what was found in the repo of the manufacturers
```
kshxtij, line 54: This also seems obvious?

kshxtij (changes requested):

lib/core/types.hpp
```diff
@@ -40,4 +43,24 @@ using AccelerometerData    = std::array<Float, kNumAccelerometers>;
...
+        y(y),
+        z(z),
+        time(time)
+  {
+  }
+
+  const std::int16_t x;
+  const std::int16_t y;
+  const std::int16_t z;
+  const TimePoint time;
```
maxguy2001, line 63: `timepoint`? `time` doesn't really describe it's use?

lib/debug/repl.cpp
```diff
@@ -427,6 +453,46 @@ void Repl::addSpiCommands(const std::uint8_t bus)
...
 
+void Repl::addAccelerometerCommands(const std::uint8_t bus, const std::uint8_t device_address)
+{
+  const auto optional_i2c = io::HardwareI2c::create(logger_, bus);
+  if (!optional_i2c) {
+    logger_.log(core::LogLevel::kFatal, "Failed to create I2C instance on bus %d", bus);
+    return;
+  }
+  const auto i2c = std::make_shared<io::HardwareI2c>(*optional_i2c);
+  if (device_address != 0x19) {
```
maxguy2001, line 464: Where does this number come from? Should it be given a name and stored somewhere?

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,140 @@
...
+  if (!low_byte) {
+    logger_.log(core::LogLevel::kFatal,
+                "Failed to read the low byte for acceleration along the %s",
+                AxisStrings[axis]);
+    return std::nullopt;
+  }
+  const auto high_byte = i2c_.readByte(kDeviceAddress, high_byte_address);
+  if (!high_byte) {
+    logger_.log(core::LogLevel::kFatal,
+                "Failed to read the low byte for acceleration along the %s",
```
maxguy2001, line 37: Wrong message?

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,140 @@
...
+                AxisStrings[axis]);
+    return std::nullopt;
+  }
+  const auto raw_acceleration
```
maxguy2001, line 41: why `const auto` when explicitly casting to `int16_t`?

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,140 @@
...
+      *low_byte_address  = kZOutLow;
+      *high_byte_address = kZOutHigh;
+      break;
+  }
+}
+
+std::int16_t Accelerometer::getAccelerationFromRaw(const std::int16_t rawAcceleration)
+{
+  // these values come from the data sheet. Don't change them.
+  std::int16_t acceleration
```
maxguy2001, line 69: why make an object just to return it. Would it not be better just directly returning the result of the calculation?

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,140 @@
...
+{
+  // check to see if the values are ready to be read
+  const auto data_ready = i2c_.readByte(kDeviceAddress, kDataReady);
+  if (!data_ready) {
+    logger_.log(core::LogLevel::kFatal, "Failed to read acceleration data");
+    return std::nullopt;
+  }
+  // TODOlater: here the error is not that bad, hence the return value should indicate that
+  if (data_ready.value() % 2 == 0) {
+    logger_.log(core::LogLevel::kFatal, "Failed to read acceleration data as it is not ready");
```
maxguy2001, line 86: Does this need to get logged as `kFatal`?

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,140 @@
...
+    return std::nullopt;
+  }
+  const auto result_x = getRawAcceleration(Axis::x);
```
maxguy2001, line 89: log the error here?

lib/sensors/accelerometer.cpp
BenET01, line 37: Should this be high_byte?

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,140 @@
...
+  if (!result_x) { return std::nullopt; }
+  const std::int16_t x_raw_acceleration = result_x.value();
```
maxguy2001, line 91: `const std::int16_t x_acceleration = getAccelerationFromRaw(result_x.value());`? Just seems like an unnecessary step to take.

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,140 @@
...
+  const std::int16_t x_acceleration     = getAccelerationFromRaw(x_raw_acceleration);
+  const auto result_y                   = getRawAcceleration(Axis::y);
+  if (!result_y) { return std::nullopt; }
+  const std::int16_t y_raw_acceleration = result_y.value();
+  const std::int16_t y_acceleration     = getAccelerationFromRaw(y_raw_acceleration);
+  const auto result_z                   = getRawAcceleration(Axis::z);
+  if (!result_z) { return std::nullopt; }
+  const std::int16_t z_raw_acceleration = result_z.value();
+  const std::int16_t z_acceleration     = getAccelerationFromRaw(z_raw_acceleration);
+  const std::optional<core::RawAccelerationData> acceleration_3axis{
```
maxguy2001, line 101: `acceleration_3_axis`?

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,71 @@
+#pragma once
+
+#include "i2c_sensors.hpp"
+
+#include <unistd.h>
+
+#include <cstdint>
+#include <cstdio>
```
maxguy2001, line 8: is this used at all?

maxguy2001 (changes requested):

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,71 @@
...
+  core::Result configure();
+  std::optional<core::RawAccelerationData> read();
+  std::uint8_t getChannel();
+
+ private:
+  core::ILogger &logger_;
+  io::HardwareI2c &i2c_;
+  const std::uint8_t channel_;
+
+ private:
```
BenET01, line 63: Why do we need two privates here?

SnickeyX: more readable if private methods are seaparated from private variables, though @Platiniom64 just flip it so that private methods come before private variables (apologies for being pedantic again)

lib/sensors/accelerometer.cpp
SnickeyX: No, this is an issue I've brought up and want to fix in another PR (this one has been around much too long). Ideally it would be `kDebug` and assuming this runs on a ROS node - the reaction to this "error" should be simply to try to read again till there is data (implementation mileage may vary*)

lib/debug/repl.cpp
SnickeyX: pretty sure you have a` kDeviceAddress` already to check this @Platiniom64 (also since these are global, and we are adding more sensors, the name should be changed to `kDefaultAccelerometerAddress` to avoid confusion later)

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,137 @@
...
+      *high_byte_address = kYOutHigh;
+      break;
+    case Axis::z:
+      *low_byte_address  = kZOutLow;
+      *high_byte_address = kZOutHigh;
+      break;
+  }
+}
+
+std::int16_t Accelerometer::getAccelerationFromRaw(const std::int16_t rawAcceleration)
```
SnickeyX, line 66: `getAccelerationFromRawValue`

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,137 @@
...
+{
+  // these values come from the data sheet. Don't change them.
+  std::int16_t acceleration
+    = static_cast<std::int16_t>((static_cast<int32_t>(rawAcceleration) * 488) / 1000);
```
TomLonergan03, line 70: why are these casting to different sizes of int? it isn't definite that `int32_t*488/1000` will fit into an `int16_t`

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,137 @@
...
+  if (ctrl1_result == core::Result::kFailure) { return core::Result::kFailure; };
+  // Enable block data update
+  // Enable address auto increment
+  const core::Result ctrl2_result
+    = i2c_.writeByteToRegister(kDeviceAddress, kCtrl2Address, kCtrl2Value);
+  if (ctrl2_result == core::Result::kFailure) { return core::Result::kFailure; };
+  // Full scale +-16g
+  // Filter bandwidth = ODR/2
+  const core::Result ctrl6_result
+    = i2c_.writeByteToRegister(kDeviceAddress, kCtrl6Address, kCtrl6Value);
```
TomLonergan03, line 133: should these comments be here or with the values in the header? I'm not sure which is better

TomLonergan03 (changes requested):

lib/sensors/accelerometer.cpp
maxguy2001: Ok, sounds good. Might be a bit pedantic but is would `kDebug` really be the most suitable name for that behavior or would we be better off defining a new log level for this purpose i.e. `kRetry`/`kTempError` (basically something along the lines of "small oopsie but nothing too crashy-burny"? This is something that Navigation and the outlier detection system could potentially use too if we decide to add this? @SnickeyX ?

SnickeyX: I don't see this happening often enough (or ever) but if you can make use of it too then a `kWarn` makes sense since this WARN logs are usually for non-fatal conditions that should not majorly disrupt any processes. I could add this as part of the PR to fix this issue. What do you think @kshxtij?

kshxtij: I agree, feel free to add it

Platiniom64 (approved):
amazing

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,126 @@
...
+    return std::nullopt;
+  }
+  // TODOlater: here the error is not that bad, hence the return value should indicate that
+  if (data_ready.value() % 2 == 0) {
+    logger_.log(core::LogLevel::kFatal, "Failed to read acceleration data as it is not ready");
+    return std::nullopt;
+  }
+  const auto result_x = getRawAcceleration(Axis::x);
+  if (!result_x) { return std::nullopt; }
+  const std::int32_t x_acceleration = getAccelerationFromRawValue(result_x.value());
```
TomLonergan03, line 85: i think the standard way to get the value of optionals is `*result_x`

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include <optional>
+
+#include <core/logger.hpp>
+#include <core/types.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+
+enum class Axis { x = 0, y, z };
+static constexpr char AxisStrings[3][10] = {"x-axis", "y-axis", "z-axis"};
```
TomLonergan03, line 17: why is this 2d char arrays instead of a string array

TomLonergan03 (changes requested):

lib/core/types.hpp
```diff
@@ -40,4 +43,24 @@ using AccelerometerData    = std::array<Float, kNumAccelerometers>;
...
+        y(y),
+        z(z),
+        time_point(time_point)
+  {
+  }
+
+  const std::int32_t x;
+  const std::int32_t y;
+  const std::int32_t z;
+  const TimePoint time_point;
```
mifrandir, line 63: `timestamp`? `measured_at`?

lib/debug/repl.cpp
```diff
@@ -227,10 +253,10 @@ void Repl::addAdcCommands(const std::uint8_t pin)
   adc_read_command.description = description.str();
   adc_read_command.handler     = [this, adc, pin]() {
     const auto value = adc->readValue();
-    if (value) {
-      logger_.log(core::LogLevel::kDebug, "ADC value from pin %d: %d", pin, *value);
-    } else {
+    if (!value) {
       logger_.log(core::LogLevel::kFatal, "Failed to read from ADC pin %d", pin);
+    } else {
+      logger_.log(core::LogLevel::kDebug, "ADC value from pin %d: %d", pin, *value);
```
mifrandir, line 259: why are we not using this pattern here?

```cpp
if (failure) { 
  // handle
  return; 
}
// do more
```

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,126 @@
...
+      high_byte_address_ = kZOutHigh;
+      break;
+  }
+}
+
+std::int32_t Accelerometer::getAccelerationFromRawValue(const std::int16_t rawAcceleration)
+{
+  // these values come from the data sheet. Don't change them.
+  std::int32_t acceleration = (static_cast<int32_t>(rawAcceleration) * 488) / 1000;
+  return acceleration;
```
mifrandir, line 64: just return the value rather than putting into a variable?

lib/sensors/accelerometer.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+core::Result Accelerometer::configure()
+{
+  // check we are communicating with the correct sensor
+  const auto device_id = i2c_.readByte(kDefaultAccelerometerAddress, kDeviceId);
+  if (!device_id) {
+    logger_.log(core::LogLevel::kFatal, "Failure to read device id of accelerometer");
+    return core::Result::kFailure;
+  }
+  if (device_id.value() != kExpectedDeviceId) {
```
mifrandir: dereference (`*`) instead of `.value()`

lib/sensors/accelerometer.hpp
```diff
@@ -0,0 +1,74 @@
...
+static constexpr std::uint8_t kYOutLow  = 0x2A;
+static constexpr std::uint8_t kYOutHigh = 0x2B;
+
+static constexpr std::uint8_t kZOutLow  = 0x2C;
+static constexpr std::uint8_t kZOutHigh = 0x2D;
+
+static constexpr std::uint8_t kDataReady = 0x27;
+
+static constexpr std::uint8_t kDeviceId         = 0x0F;
+static constexpr std::uint8_t kExpectedDeviceId = 0x44;
```
mifrandir, line 50: 1. do these need to live in the header?
2. if so, do we want them to be `static`?
3. what is the difference between `kDeviceId` and `kExpectedDeviceId`? Why are there two of these?

mifrandir (changes requested):

maxguy2001 (changes requested):

TomLonergan03 (approved):
happy with this

maxguy2001 (approved):

BenET01 (approved):
Looks good

kshxtij (approved):

mifrandir (approved):

PR #32, opened 2022-11-03
SNS - Temperature

SnickeyX:
@HTTYDKing 
- [x] implement `read`, `write`
- [x] confirm code works with temperature sensor
- [x] Remove `device_address` and replace with the default address values the temp sns takes


maxguy2001 (approved):
looks good!

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,27 @@
...
+std::optional<uint16_t> Temperature::read()
+{
+  // TODOLater: read from i2c
+  log_.log(hyped::core::LogLevel::kFatal, "Temperature read not implemented");
+  return std::nullopt;
+}
+
+bool Temperature::configure()
+{
+  // TODOLater: configure i2c
```
mifrandir, line 22: So what does this actually implement? Doing the I/O is the only non-trivial bit.

mifrandir: do we have the datasheet?

SnickeyX: yes

mifrandir: then implement it.

SnickeyX: @HTTYDKing

kshxtij: I think that was the intention of this pr anyway? We weren't going to merge this without the temp sensor working

mifrandir: `// TODO` not `// TODOLater`

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,39 @@
...
+namespace hyped::sensors {
+
+Temperature::Temperature(hyped::io::I2c &i2c, hyped::core::ILogger &log) : log_(log), i2c_(i2c)
+{
+}
+
+std::optional<std::uint16_t> Temperature::read()
+{
+  if (i2c_.readByte(TEMPERATURE_1, STATUS) == 0) {
+    auto Temp_High = i2c_.readByte(TEMPERATURE_1, TEMP_H);
```
kshxtij, line 12: readByte returns an optional, you'd want to check if the operation was successful

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,39 @@
...
+    auto Temp_Low  = i2c_.readByte(TEMPERATURE_1, TEMP_L);
```
kshxtij, line 13: same as above

lib/sensors/temperature.cpp
kshxtij: Also const

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,39 @@
...
+    // getting the high and low values from the register
+
+    auto Temp = (TEMP_H << 8) | TEMP_L;
```
kshxtij, line 16: This is just a uint16_t?

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,39 @@
...
+    Temp      = Temp * 0.01;
+
+    return std::optional<std::uint16_t>{Temp};
```
kshxtij, line 19: just `return Temp`

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,39 @@
...
+    // Need to have it do the correct effect and output a log of what occurred
+  } else {
+    return std::nullopt;
+  }
+  // log_.log(hyped::core::LogLevel::kFatal, "Temperature read not implemented");
+}
+
+bool Temperature::configure()
+{
+  i2c_.writeByte(TEMPERATURE_1, CTRL, 1);
```
kshxtij, line 30: We want to know if this was successful or not, writeByte returns `hyped::core::Result`

lib/sensors/temperature.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <cstdint>
+#include <cstdio>
+#include <optional>
+
+#include <core/logger.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+
+const std::uint8_t CTRL          = 0x04;
```
kshxtij, line 14: rename all the constants to start with k

lib/sensors/temperature.hpp
```diff
@@ -0,0 +1,34 @@
...
+const std::uint8_t TEMPERATURE_1 = 0x38;
```
kshxtij, line 15: `kTemperatureAddress1` unclear on this, @SnickeyX thoughts

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,39 @@
...
+    auto Temp = (TEMP_H << 8) | TEMP_L;
+    Temp      = Temp * 0.01;
+
+    return std::optional<std::uint16_t>{Temp};
+    // Reading from the Status register to see if the tempurature is ready to be read
+    // Need to have it do the correct effect and output a log of what occurred
+  } else {
+    return std::nullopt;
+  }
+  // log_.log(hyped::core::LogLevel::kFatal, "Temperature read not implemented");
```
kshxtij, line 25: Delete this

kshxtij (changes requested):

lib/sensors/temperature.cpp
mifrandir: also variable names

lib/sensors/temperature.hpp
mifrandir: and `static constexpr`

lib/sensors/temperature.hpp
SnickeyX: So we don't want any device addresses, since the device addresses of these will simply be the mux's. Instead the constructor of these sensors on the mux should take in the `mux_adderess` as a parameter and call its reads/writes with that.

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace hyped::sensors {
+
+Temperature::Temperature(hyped::io::I2c &i2c, hyped::core::ILogger &log) : log_(log), i2c_(i2c)
+{
+}
+
+std::optional<std::uint16_t> Temperature::read()
+{
+  if (i2c_.readByte(TEMPERATURE_1, STATUS) == 0) {
```
kshxtij, line 11: neither TEMPERATURE_1 nor STATUS are defined anywhere

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,41 @@
...
+    const std::optional<std::uint8_t> Tempurature_High = i2c_.readByte(kMUX, TEMP_H);
```
kshxtij, line 12: you renamed TEMP_H in the header file

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,41 @@
...
+    const std::optional<std::uint8_t> Tempurature_Low  = i2c_.readByte(kMUX, TEMP_L);
```
kshxtij, line 13: same as above

lib/sensors/temperature.cpp
kshxtij: Also temperature_high

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,41 @@
...
+    // getting the high and low values from the register
+
+    if (Tempurature_High = !std::nullopt, Tempurature_Low = !std::nullopt) {
```
kshxtij, line 16: temperature_high.has_value() && temperature_low.has_value()

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,41 @@
...
+      std::optional<std::uint16_t> Tempurature = (Tempurature_High << 8) | Tempurature_Low;
```
kshxtij, line 17: Why is this an optional?

Just `std::uint16_t temperature = ...`

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,41 @@
...
+    }
+    // Reading from the Status register to see if the tempurature is ready to be read
+    // Need to have it do the correct effect and output a log of what occurred
+  }
+  // log_.log(hyped::core::LogLevel::kFatal, "Temperature read not implemented");
+}
+
+bool Temperature::configure()
+{
+  hyped::core::Result::Check = i2c_.writeByte(kMUX, CTRL, 1);
```
kshxtij, line 32: hyped::core::Result is a type 

you want `core::Result check = ....` 

check isn't a good var name either so make it more descriptive

lib/sensors/temperature.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include <core/logger.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+
+static constexpr std::uint8_t kMUX    = 0x70;
+static constexpr std::uint8_t kCTRL   = 0x04;
+static constexpr std::uint8_t kTEMP_H = 0x07;
+static constexpr std::uint8_t kTEMP_L = 0x06;
```
kshxtij, line 17: kTemperatureHigh
kTemperatureLow

kshxtij (changes requested):

lib/sensors/temperature.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include <cstdint>
+#include <cstdio>
+#include <optional>
+
+#include <core/logger.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+
+static constexpr std::uint8_t kMux             = 0x70;
```
SnickeyX, line 14: We don't want this here, this is what the mux class defines.

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,53 @@
+#include "temperature.hpp"
+
+namespace hyped::sensors {
+
+Temperature::Temperature(hyped::io::I2c &i2c, hyped::core::ILogger &log) : log_(log), i2c_(i2c)
```
SnickeyX, line 5: we should take in a `mux_address` from the constructor here and use that as the "device address" when we read/write

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,53 @@
...
+{
+}
+
+std::optional<std::uint16_t> Temperature::read()
+{
+  const auto Status_check = i2c_.readByte(kMux, kStatus);
```
SnickeyX, line 11: i think just change the left to `const std::optional<std::uint8_t> status_check_result`

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+  if (Status_check) {
+    if (Status_check.value() == 0) {
+      const std::optional<std::uint8_t> tempurature_high = i2c_.readByte(kMux, kTempuratureHigh);
+      const std::optional<std::uint8_t> tempurature_low  = i2c_.readByte(kMux, kTempuratureLow);
+      // getting the high and low values from the register
```
SnickeyX, line 17: `temperature_high_byte` and `temperature_low_byte`

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    } else {
+      log_.log(hyped::core::LogLevel::kFatal, "Temperature read not implemented");
+      return std::nullopt;
+    }
+  } else {
+    log_.log(hyped::core::LogLevel::kFatal, "Temperature read not implemented");
+    return std::nullopt;
+  }
+}
```
SnickeyX, line 39: I dont think this is very readable. Start with checking individually if temp_high or temp_low has a value and log + return nullopt if not. Then you can straight away do the bit shifting part since you will have ensured both have a value.

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,53 @@
...
+  const auto Status_check = i2c_.readByte(kMux, kStatus);
+
+  if (Status_check) {
+    if (Status_check.value() == 0) {
+      const std::optional<std::uint8_t> tempurature_high = i2c_.readByte(kMux, kTempuratureHigh);
+      const std::optional<std::uint8_t> tempurature_low  = i2c_.readByte(kMux, kTempuratureLow);
+      // getting the high and low values from the register
+
+      if (tempurature_high.has_value() && tempurature_low.has_value()) {
+        std::uint16_t tempurature = (tempurature_high.value() << 8) | tempurature_low.value();
```
SnickeyX, line 20: Spelling here + const.

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,53 @@
...
+    }
+  } else {
+    log_.log(hyped::core::LogLevel::kFatal, "Temperature read not implemented");
+    return std::nullopt;
+  }
+}
+
+core::Result Temperature::configure()
+{
+  const core::Result write_result = i2c_.writeByteToRegister(kMux, kControl, 1);
```
SnickeyX, line 43: I don't think we want to write a 1 to the CTRL reg since this would activate one_shot_ mode requiring us to re-configure before reading again. We want to write a `0x3C` (0b111100) since this activates continuous mode, allows to to carry out mulitiple reads in succession (from IF_ADD_INC) and sets the output data rate to 200hz (the highest possible). I previously tested with this setting and it gave us expected results.

SnickeyX: also don't directly write `1` or `0x3C`,  make a `static constexpr` and call this `kConfigurationSetting` with a little explanation as to why that number (feel free to use how I justified it above)

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,53 @@
...
+  // We need to know if a begalebone has two registers set to it
```
SnickeyX, line 44: don't need this

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,53 @@
...
+  if (write_result == hyped::core::Result::kFailure) {
+    log_.log(hyped::core::LogLevel::kFatal, "Temperature configure not implemented");
+    return core::Result::kFailure;
+  }
+  // Writing to the Control register 0x04 and setting that to 1 for TEMPERATURE_1
```
SnickeyX, line 49: don't need this

lib/sensors/temperature.hpp
```diff
@@ -0,0 +1,33 @@
...
+static constexpr std::uint8_t kControl         = 0x04;
+static constexpr std::uint8_t kTempuratureHigh = 0x07;
+static constexpr std::uint8_t kTempuratureLow  = 0x06;
```
SnickeyX, line 17: I would prefer these to be of the form `k*insert reg name from datasheet*` with a little comment to the side to explain if there is any ambiguity. Eg, `kData_T_H  // high byte for the temperature value` so it can be very easily referenced with the datasheet. What do you think @kshxtij ?

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,53 @@
...
+}
+
+std::optional<std::uint16_t> Temperature::read()
+{
+  const auto Status_check = i2c_.readByte(kMux, kStatus);
+
+  if (Status_check) {
+    if (Status_check.value() == 0) {
+      const std::optional<std::uint8_t> tempurature_high = i2c_.readByte(kMux, kTempuratureHigh);
+      const std::optional<std::uint8_t> tempurature_low  = i2c_.readByte(kMux, kTempuratureLow);
```
mifrandir, line 16: make these `auto`, too

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,76 @@
...
+    log_.log(hyped::core::LogLevel::kFatal, "Temperature high could not be read");
+    return std::nullopt;
+  }
+
+  if (!temperature_low) {
+    log_.log(hyped::core::LogLevel::kFatal, "Temperature low could not be read");
+    return std::nullopt;
+  }
+
+  std::uint16_t temperature = (temperature_high.value() << 8) | temperature_low.value();
```
SnickeyX, line 49: think we shud keep this to `int16_t` to preserve sign, jic its negative due to a faulty sensor

lib/sensors/temperature.cpp
```diff
@@ -2,45 +2,63 @@
...
-  } else {
-    log_.log(hyped::core::LogLevel::kFatal, "Temperature read not implemented");
+  const auto status_check_result = i2c_.readByte(mux_address_, kStatus);
+
+  const auto temperature_high = i2c_.readByte(mux_address_, kData_T_H);
+  const auto temperature_low  = i2c_.readByte(mux_address_, kData_T_L);
+
+  // getting the temperature_high_byte and temperature_low_byte values from the register
+
+  if (!status_check_result) {
```
mifrandir, line 29: this should move right below the read

you should do the read and then immediately check that it was successful, always.

lib/sensors/temperature.cpp
```diff
@@ -2,45 +2,63 @@
...
-    }
-  } else {
-    log_.log(hyped::core::LogLevel::kFatal, "Temperature read not implemented");
+  const auto status_check_result = i2c_.readByte(mux_address_, kStatus);
+
+  const auto temperature_high = i2c_.readByte(mux_address_, kData_T_H);
+  const auto temperature_low  = i2c_.readByte(mux_address_, kData_T_L);
+
+  // getting the temperature_high_byte and temperature_low_byte values from the register
+
```
mifrandir, line 28: lots of empty lines; get rid of those

lib/sensors/temperature.cpp
```diff
@@ -2,45 +2,63 @@
...
-      return std::nullopt;
-    }
-  } else {
-    log_.log(hyped::core::LogLevel::kFatal, "Temperature read not implemented");
+  const auto status_check_result = i2c_.readByte(mux_address_, kStatus);
+
+  const auto temperature_high = i2c_.readByte(mux_address_, kData_T_H);
+  const auto temperature_low  = i2c_.readByte(mux_address_, kData_T_L);
+
+  // getting the temperature_high_byte and temperature_low_byte values from the register
```
mifrandir, line 27: duh; don't have useless comments

lib/sensors/temperature.cpp
mifrandir: also make this `const` and directly return `temperature` scaled appropriately

lib/sensors/temperature.cpp
```diff
@@ -2,45 +2,63 @@
...
+  }
+
+  if (!temperature_low) {
+    log_.log(hyped::core::LogLevel::kFatal, "Temperature low could not be read");
+    return std::nullopt;
+  }
+
+  std::uint16_t temperature = (temperature_high.value() << 8) | temperature_low.value();
+
+  temperature = temperature * 0.01;
```
mifrandir, line 51: magic constant? here a comment would be helpful...

mifrandir (changes requested):
getting there, good job!

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,62 @@
...
+                         hyped::io::I2c &i2c,
+                         hyped::core::ILogger &log)
+    : log_(log),
+      i2c_(i2c),
+      channel_(channel),
+      device_address_(device_address)
+{
+}
+
+Temperature::~Temperature()
```
mifrandir, line 16: why do we have this deconstructor? this is implicit

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,62 @@
...
+  if (!status_check_result) {
+    log_.log(hyped::core::LogLevel::kFatal, "Temperature could not read status");
+    return std::nullopt;
+  }
+  if (status_check_result.value() == 0) {
+    log_.log(hyped::core::LogLevel::kFatal, "Temperature sensor is not ready to be read from");
+    return std::nullopt;
+  }
+  const auto temperature_high_byte = i2c_.readByte(device_address_, kDataTH);
+  const auto temperature_low_byte  = i2c_.readByte(device_address_, kDataTL);
```
mifrandir, line 32: this should go below the if

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,62 @@
...
+      channel_(channel),
+      device_address_(device_address)
+{
+}
+
+Temperature::~Temperature()
+{
+}
+
+std::optional<std::int16_t> Temperature::read()
```
mifrandir, line 20: sorry to bring this up again: can the sensor correctly measure negative temperatures?

if yes, then this is fine

if a negative temperature always implies that something has gone wrong (without making the assumption that there won't be negative temps outside!) then this should return `std::uint16_t` and `std::nullopt` in case of this error.

lib/sensors/temperature.hpp
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include "i2c_sensors.hpp"
+
+#include <cstdint>
+#include <cstdio>
```
mifrandir, line 6: this isn't needed here

lib/sensors/temperature.hpp
```diff
@@ -0,0 +1,40 @@
...
+static constexpr std::uint8_t kStatus = 0x05;
+// Sets the sensor to continuous mode, sets IF_ADD_INC, and sets sampling rate to 200Hz
+static constexpr std::uint8_t kConfigurationSetting = 0x3c;
+
+class Temperature : public II2cMuxSensor<std::int16_t> {
+ public:
+  Temperature(const std::uint8_t device_address,
+              const std::uint8_t channel,
+              io::I2c &i2c,
+              hyped::core::ILogger &log);
```
mifrandir, line 26: I don't like the ordering here; can we put the logger, then I2c and then everything else? basically the rule should be more common things to the front and more specialised things to the back

mifrandir (changes requested):

lib/sensors/temperature.cpp
SnickeyX: yes it can

lib/sensors/temperature.hpp
SnickeyX: i think this is the case with other places in the codebase too, will fix some in my mux PR

lib/sensors/temperature.cpp
mifrandir: And what is the scaling of the value here? Is it degrees Celsius in ints? That doesn't seem very precise...

lib/sensors/temperature.hpp
mifrandir: fixed by #59

lib/sensors/temperature.cpp
SnickeyX: It is degrees Celcius. We agreed on ints since having decimal precision does not really matter when we are at 80.1 C or -10.5 C if we simply checking if temp is above or below some fixed value.

mifrandir: Hm, but the sensor is more precise? I feel like it's kind of pointless to throw away that precision here. It won't give any measurable acceleration over 32-bit floats and since we're not fussed about precision anyways, the usual float caveats don't apply.

I don't feel super strongly about this, but thought I'd mention it.

SnickeyX: I'll ask the other teams today again if they care at all about temperature precision jic - though from our prespective it invovles a minor change that does not really affect our checks. 

@kshxtij any thoughts?

kshxtij: None of the hardware teams need decimal places, so i don't really think it matters

lib/sensors/temperature.hpp
```diff
@@ -0,0 +1,39 @@
...
```
mifrandir, line 25: order the arguments in the constructor the same way you ordered the class members

lib/sensors/temperature.hpp
SnickeyX: @HTTYDKing quick fix on this + run clang format before commiting

mifrandir (approved):
Modulo formatting.

SnickeyX:
@HTTYDKing make sure to also remove the `device_address` parameter and bring back the `kDefaultAddress` of the temperature sensors themselves - use that address to query, iirc it was 0x38 and 0x3F,, but we will default to 0x38

lib/sensors/temperature.hpp
```diff
@@ -0,0 +1,36 @@
...
+static constexpr std::uint8_t kCtrl        = 0x04;
+static constexpr std::uint8_t kDataTH      = 0x07;
+static constexpr std::uint8_t kDataTL      = 0x06;
+static constexpr std::uint8_t kStatus      = 0x05;
+// Sets the sensor to continuous mode, sets IF_ADD_INC, and sets sampling rate to 200Hz
+static constexpr std::uint8_t kConfigurationSetting = 0x3c;
+
+class Temperature : public II2cMuxSensor<std::int16_t> {
+ public:
+  Temperature(hyped::core::ILogger &log, io::I2c &i2c, const std::uint8_t channel);
```
SnickeyX, line 23: change `log` to `logger`

lib/sensors/temperature.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include <cstdint>
+#include <optional>
+
+#include <core/logger.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+
+static constexpr std::uint8_t kTemperature = 0x38;
```
SnickeyX, line 13: `kTemperatureDefaultAddress`

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,54 @@
...
+      i2c_(i2c),
+      channel_(channel)
+{
+}
+
+std::optional<std::int16_t> Temperature::read()
+{
+  const auto status_check_result = i2c_.readByte(kTemperature, kStatus);
+  if (!status_check_result) {
+    log_.log(hyped::core::LogLevel::kFatal, "Temperature could not read status");
```
SnickeyX, line 16: `logger_log.(...., "Failed to read temerature sensor status at channel %d", channel_); ` -> and similarly for the rest of the logs. Basically just want to standardize logging, so if it's something that went wrong, just start with "Failed to ..." and if something went right start with "Successfully ..." (so you have to think less about the message itself). The `channel_` id in the log just helps identify which exact sensor it was that failed.

SnickeyX (commented):
Mostly just formatting changes, want this to follow the same standard as the I/O library.

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,58 @@
+#include "temperature.hpp"
+
+namespace hyped::sensors {
+
+Temperature::Temperature(hyped::core::ILogger &log, io::I2c &i2c, const std::uint8_t channel)
+    : log_(log),
```
TomLonergan03, line 6: should be `logger_(logger)` for consistency across repo

TomLonergan03 (changes requested):

lib/sensors/temperature.cpp
```diff
@@ -9,6 +9,10 @@ Temperature::Temperature(hyped::core::ILogger &log, io::I2c &i2c, const std::uin
 {
 }
 
+Temperature::~Temperature(){
```
mifrandir: formatting

mifrandir (changes requested):

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,58 @@
...
+}
+
+std::optional<std::int16_t> Temperature::read()
+{
+  const auto status_check_result = i2c_.readByte(kTemperature, kStatus);
+  if (!status_check_result) {
+    log_.log(hyped::core::LogLevel::kFatal, "Temperature could not read status");
+    return std::nullopt;
+  }
+  if (status_check_result.value() == 0) {
```
SnickeyX, line 23: This comparison doesn't make any sense since it does not signify anything from the datasheet. The status reg can check for an invalid read (a temp reading beyond the range of the sensor) and also check if the sensor is still busy coverting the data. For the first case, the value will be 0x06 and 0x01 for the latter case. You should add these two as two separate failure cases.

SnickeyX: In fact, we would actually want to handle the case where data is not ready specially since its not really a failure but rather needs us to check with the sensor again. I will look into this after this PR is merged, but do be sure to add both the cases as failure for now.

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,87 @@
...
+std::optional<std::int16_t> Temperature::read()
+{
+  const auto status_check_result = i2c_.readByte(kTemperatureDefaultAddress, kStatus);
+  if (!status_check_result) {
+    logger_.log(hyped::core::LogLevel::kFatal,
+                "Failed to read temperature sensor status at channel %d",
+                channel_);
+    return std::nullopt;
+  }
+  // Status checks
```
kshxtij, line 25: Needed?

lib/sensors/temperature.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include "i2c_sensors.hpp"
+
+#include <cstdint>
+#include <optional>
+
+#include <core/logger.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+// These numbers are taken from the data sheet
```
kshxtij, line 12: `Values from datasheet`?

kshxtij (changes requested):

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,86 @@
...
+  const auto temperature_low_byte = i2c_.readByte(kTemperatureDefaultAddress, kDataTemperatureLow);
+  if (!temperature_low_byte) {
+    logger_.log(
+      hyped::core::LogLevel::kFatal, "Failed to read temperature low at channel %d", channel_);
+    return std::nullopt;
+  }
+  const auto temperature = ((temperature_high_byte.value() << 8) | temperature_low_byte.value());
+
+  logger_.log(
+    hyped::core::LogLevel::kFatal, "Successfully read temperature sensor at channel %d", channel_);
```
BenET01, line 60: Why is this logging a kFatal?

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,86 @@
+#include "temperature.hpp"
+
+namespace hyped::sensors {
+
+Temperature::Temperature(hyped::core::ILogger &log, io::I2c &i2c, const std::uint8_t channel)
+    : logger_(log),
```
Fayz-S, line 6: Naming convention for logger is now `log_`, needs changed

lib/io/i2c.cpp
```diff
@@ -16,7 +16,7 @@ namespace hyped::io {
 I2c::I2c(const std::uint8_t bus_address, core::ILogger &log) : sensor_address_(0), log_(log)
 {
   char path[13];  // up to "/dev/i2c-2"
-  sprintf(path, "/dev/i2c-%d", bus_address);
+  snprintf(path, sizeof(path), "/dev/i2c-%d", bus_address);
```
maxguy2001, line 19: This has been removed in main branch so just merge master into this branch to resolve merge conflicts

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,86 @@
...
+std::optional<std::int16_t> Temperature::read()
+{
+  const auto status_check_result = i2c_.readByte(kTemperatureDefaultAddress, kStatus);
+  if (!status_check_result) {
+    logger_.log(hyped::core::LogLevel::kFatal,
+                "Failed to read temperature sensor status at channel %d",
+                channel_);
+    return std::nullopt;
+  }
+  if (status_check_result.value() == kBusy) {
```
Fayz-S, line 25: Could this not be a switch instead of a series of if statements to check `status_check_result.value()`? And if they are if statements checking one value there should be else-if statements

maxguy2001, line 25: surely `if (status_check_result.value() == kUnderTemperatureLowLimit` ??

maxguy2001: or change the error message?

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,86 @@
...
+  }
+  if (status_check_result.value() == kOverTemperatureHighLimit) {
+    logger_.log(hyped::core::LogLevel::kFatal,
+                "Failed to read, temperature is above higher limit at channel %d",
+                channel_);
+    return std::nullopt;
+  }
+  if (status_check_result.value() == kUnderTemperatureLowLimit) {
+    logger_.log(hyped::core::LogLevel::kFatal,
+                "Failed to read, temperature sensor is not ready to be read at channel %d",
```
maxguy2001, line 39: I think you might have got a checking if it's busy mixed  up with a temperature check?

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,86 @@
...
+{
+  const core::Result write_result
+    = i2c_.writeByteToRegister(kTemperatureDefaultAddress, kCtrl, kConfigurationSetting);
+  if (write_result == hyped::core::Result::kFailure) {
+    logger_.log(hyped::core::LogLevel::kFatal,
+                "Failed to configure temperature sensor at channel %d",
+                channel_);
+    return core::Result::kFailure;
+  }
+  logger_.log(hyped::core::LogLevel::kFatal,
```
Fayz-S, line 75: kFatal?

Fayz-S (changes requested):

maxguy2001 (changes requested):

lib/sensors/temperature.cpp
SnickeyX: `logger_` I believe

SnickeyX:
as a note: I will be making changes to add this to REPL + using static create (will need to change mux.hpp) in another PR after this and acceleormeter get merged

lib/sensors/temperature.hpp
```diff
@@ -6,34 +6,35 @@
...
 
   core::Result configure();
   std::optional<std::int16_t> read();
   std::uint8_t getChannel();
 
  private:
-  hyped::core::ILogger &logger_;
-  hyped::io::I2c &i2c_;
+  core::ILogger &logger_;
+  io::HardwareI2c &i2c_;
```
mifrandir, line 37: I think this should be `io::II2c`, that's exactly why we have the interface :)

lib/sensors/temperature.cpp
```diff
@@ -0,0 +1,83 @@
...
+    logger_.log(
+      core::LogLevel::kFatal, "Failed to read low byte for temperature at channel %d", channel_);
+    return std::nullopt;
+  }
+  const std::int16_t temperature
+    = ((temperature_high_byte.value() << 8) | temperature_low_byte.value());
+  logger_.log(
+    core::LogLevel::kDebug, "Successfully read from temperature sensor at channel %d", channel_);
+  // Scaling temperature as per the datasheet
+  return temperature * kTemperatureScaleFactor;
```
mifrandir, line 61: this is a float and it's being cast to `std::int16_t`

either avoid the cast or make it obvious

lib/sensors/temperature.hpp
```diff
@@ -0,0 +1,41 @@
...
+static constexpr core::Float kTemperatureScaleFactor = 0.01;
+
+class Temperature : public II2cMuxSensor<std::int16_t> {
+ public:
+  Temperature(core::ILogger &logger, io::HardwareI2c &i2c, const std::uint8_t channel);
+  ~Temperature();
+
+  core::Result configure();
+  std::optional<std::int16_t> read();
+  std::uint8_t getChannel();
```
mifrandir, line 33: this can be `const` on the right

mifrandir: actually, I'm doubtful that we will ever need this getter; the channel is passed on construction and never changed, why do we care here? I'd say omit this until it is required.

mifrandir (changes requested):

lib/sensors/temperature.hpp
SnickeyX: we do need this in the context of the mux, else we would have to pass in the channels used to the mux as an array

mifrandir (approved):

TomLonergan03 (approved):
lgtm

DylanCavers (approved):

kshxtij (approved):

PR #33, opened 2022-11-04
SNS - MUX

SnickeyX:
Goal : Make a general mux class which works with any I2C sensor that needs to be on a mux. 

- [x] Figure out how to take in multiple II2cSensor<T> as a param
- [x] Implement `readAllChannels`, `selectChannel` and `closeAllChannels`


SnickeyX:
Problem: I need to pass the different I2C sensors into the mux by reference since mux should take in abstract class instances. An array (or vector) does not allow for references so pointers it is but was trying with vectors to test first. I've tried using a vector of shared ptrs (didn't work) or even just a vector of pointers (this shud never be used but wanted to get something to work but didnt work either). Need some advice on this @mifrandir.

kshxtij:
Something like `std::vector<std::unique_ptr<II2cSensor<uint16_t>>>`?

lib/sensors/mux.cpp
```diff
@@ -0,0 +1,44 @@
...
+template<typename T>
+MuxWriteResult Mux<T>::write(uint8_t data)
+{
+  // TODOLater: write to mux
+  log_.log(hyped::core::LogLevel::kFatal, "Mux write not implemented");
+  return MuxWriteResult::kFailure;
+}
+
+template<typename T>
+std::optional<std::vector<T>> Mux<T>::readAll()
```
mifrandir, line 37: Vectors here are a bad idea. They allocate on the heap which is not something we want to so at runtime. Use `std::array` instead and add the size to the template.

lib/sensors/mux.hpp
```diff
@@ -0,0 +1,39 @@
...
+  Mux(hyped::io::I2c &i2c,
+      const uint8_t mux_address,
+      std::vector<std::unique_ptr<II2cSensor<T>>> sensors,
+      hyped::core::ILogger &log);
+  ~Mux();
+
+  bool selectMode(Mode io_mode);
+  MuxWriteResult write(uint8_t data);
+  std::optional<std::vector<T>> readAll();
+  bool selecChannel(uint8_t channel);
```
mifrandir, line 31: typo

const

lib/sensors/mux.hpp
```diff
@@ -0,0 +1,39 @@
...
+template<class T>
+class Mux {
+ public:
+  Mux(hyped::io::I2c &i2c,
+      const uint8_t mux_address,
+      std::vector<std::unique_ptr<II2cSensor<T>>> sensors,
+      hyped::core::ILogger &log);
+  ~Mux();
+
+  bool selectMode(Mode io_mode);
```
mifrandir, line 28: const

mifrandir: also: return type bool?

lib/sensors/sensors.hpp
```diff
@@ -17,6 +17,7 @@ class II2cSensor {
    */
   virtual bool configure()        = 0;
   virtual std::optional<T> read() = 0;
+  virtual ~II2cSensor() {}
```
mifrandir, line 20: Unsure whether this is in the right file? sensors.hpp is very vague. Just make one i2c.hpp/cpp?

mifrandir (changes requested):

lib/sensors/mux.cpp
SnickeyX: Will do

lib/sensors/mux.hpp
SnickeyX: Wanted to call selectMode in readAll and write before anything actually happens in those functions, but if selectMode fails for whatever reason (since it also involves writing to the I2C mux) -> wanted to instantly return a `std::nulopt` or `kFailure`

lib/sensors/sensors.hpp
SnickeyX: `i2c_sensors.hpp`?

SnickeyX: just so its not confused with the io i2c files

lib/sensors/mux.hpp
mifrandir: make it a success or failure return type then; not `bool`

SnickeyX:
- Adding `getChannel` to interface and changing it to `II2cMuxSensor` so that we can specify which channel a specific sensor is on to avoid the issue of electroics using only some channels on the mux.  
- Got rid of `selectMode` because its job is essentially being done by io i2c.
- Clearing all channels atm after reading from a channel because 1. We never need to read from more than one  channel at a time and 2. I think having more channels open for no reason may lead to some hardware drawbacks? (confirming with electronics today) but regardless, no harm in closing unused channels. 
- Another issue to discuss with eletronics: how many sensors on a mux can fail before we start panicking (i.e retuning `nullopts` relentlessly). I don't think we should return a `nullopt` if for example 3/4 sensors are working on a mux while being located in the same area on the pod for the same purpose (e.g multiple temperature sensors on the front of the pod). I think the two main considerations here are the number of sensors not working + the type of sensor (will discuss with electronics to figure this out).

mifrandir (changes requested):

lib/sensors/i2c_sensors.hpp
```diff
@@ -6,17 +6,21 @@
 
 namespace hyped::sensors {
 
+enum class I2cConfigureResult { kFailure = 0, kSuccess };
```
maxguy2001, line 9: Do we need a separate enum class for each protocol or could we just define a generic sensor configure result in types.hpp or in a new io/consts.hpp or similar? Surely we would still know what fails since we will be logging it?

lib/sensors/i2c_sensors.hpp
```diff
@@ -6,17 +6,21 @@
...
+
```
maxguy2001, line 10: Idk if this is a standardized thing I'm missing here but what is `T`. Like what would it typically represent if you made a class inheriting from this one?

SnickeyX: Template - so it can be any type. The whole point of this was to be able to return different types when you call `readAllChannels` in mux depending on the sensor.

lib/sensors/mux.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include <optional>
+
+#include <core/logger.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+
+static constexpr std::uint8_t kDefaultMuxAddress = 0x70;
+
+enum class MuxOperationResult { kFailure = 0, kSuccess };
```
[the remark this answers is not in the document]
SnickeyX: Same as before, will make a PR standardizing this if @kshxtij agrees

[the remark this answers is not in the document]
kshxtij: Yeah i like a core::types::Result and `kFailure = 0, kSuccess`, I'll pr it tonight and you can merge master into this branch and update it

lib/sensors/i2c_sensors.hpp
kshxtij: See below

lib/sensors/mux.cpp
```diff
@@ -18,50 +18,49 @@ Mux<T, size>::~Mux()
...
-  for (size_t i = 0; i < size; ++i) {
-    const auto sensor     = sensors_[i];
-    const uint8_t channel = sensor->getChannel();
+  for (std::size_t i = 0; i < size; ++i) {
+    const auto sensor          = sensors_[i];
+    const std::uint8_t channel = sensor->getChannel();
     // First ensure correct channel is selected
-    const MuxOperationResult channel_select_result = selectChannel(channel);
-    if (channel_select_result == MuxOperationResult::kFailure) {
+    const hyped::core::Result channel_select_result = selectChannel(channel);
```
mifrandir, line 56: just `core::Result`

mifrandir: actually, this one can be `auto`

lib/sensors/mux.hpp
```diff
@@ -3,37 +3,31 @@
...
 #include <io/i2c.hpp>
 
 namespace hyped::sensors {
 
-static constexpr uint8_t kDefaultMuxAddress = 0x70;
-
-enum class MuxOperationResult { kFailure = 0, kSuccess };
+static constexpr std::uint8_t kDefaultMuxAddress = 0x70;
 
 template<class T, std::size_t size>
```
mifrandir, line 14: Let's stick with the standard library and call it `std::size_t N`; cf. https://en.cppreference.com/w/cpp/container/array.

mifrandir (changes requested):

kshxtij (approved):

lib/sensors/mux.cpp
```diff
@@ -31,11 +31,11 @@ hyped::core::Result Mux<T, size>::selectChannel(const std::uint8_t channel)
...
-template<typename T, std::size_t size>
-hyped::core::Result Mux<T, size>::closeAllChannels()
+template<typename T, std::size_t N>
+hyped::core::Result Mux<T, N>::closeAllChannels()
 {
-  const std::uint8_t clear_channel_buffer    = 0x00;
-  const hyped::core::Result i2c_write_result = i2c_.writeByte(mux_address_, clear_channel_buffer);
+  const std::uint8_t clear_channel_buffer = 0x00;
+  const auto i2c_write_result             = i2c_.writeByte(mux_address_, clear_channel_buffer);
   if (i2c_write_result == hyped::core::Result::kSuccess) {
```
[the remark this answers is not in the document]
SnickeyX: do you want me to remove it for io as well?

kshxtij:
@mifrandir @maxguy2001 can we get reviews on this?

lib/sensors/mux.cpp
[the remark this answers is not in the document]
mifrandir: Of course! I still see plenty of `hyped::` by the way. If it's not required, don't put it there.

lib/sensors/mux.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+template<typename T, std::size_t N>
+Mux<T, N>::~Mux()
+{
+}
+
+template<typename T, std::size_t N>
+core::Result Mux<T, N>::selectChannel(const std::uint8_t channel)
+{
+  const std::uint8_t channel_buffer = 1 << channel;
```
mifrandir, line 23: this may overflow; make sure the argument passed here is correct

lib/sensors/mux.cpp
```diff
@@ -0,0 +1,84 @@
...
+               core::ILogger &log)
+    : log_(log),
+      i2c_(i2c),
+      sensors_(std::move(sensors))
+{
+}
+
+template<typename T, std::size_t N>
+Mux<T, N>::~Mux()
+{
```
mifrandir, line 17: in light of the other comment, you should static assert here that `N < 8`

lib/sensors/mux.cpp
```diff
@@ -0,0 +1,84 @@
...
+               const std::uint8_t mux_address,
+               const std::array<std::unique_ptr<II2cMuxSensor<T>>, N> sensors,
+               core::ILogger &log)
+    : log_(log),
+      i2c_(i2c),
+      sensors_(std::move(sensors))
+{
+}
+
+template<typename T, std::size_t N>
```
mifrandir, line 15: moreover, this can then just be `std::uint8_t`

mifrandir (changes requested):

lib/sensors/mux.cpp
SnickeyX: in the deconstructor and not the constructor?

SnickeyX: got it, pushing in a sec

lib/sensors/mux.cpp
```diff
@@ -10,14 +10,15 @@ Mux<T, N>::Mux(io::I2c &i2c,
       i2c_(i2c),
       sensors_(std::move(sensors))
 {
+  static_assert(N <= 8, "Mux can only have up to 8 channels");
```
mifrandir, line 13: But `static_cast<std::uint8_t>(1 << 8) == 0`?

SnickeyX: whoops, that equals to was copilot

mifrandir: but you can't have more sensors than you have channels

lib/sensors/mux.hpp
```diff
@@ -10,6 +10,7 @@
 namespace hyped::sensors {
 
 static constexpr std::uint8_t kDefaultMuxAddress = 0x70;
+static constexpr std::uint8_t kMaxChannelNumber  = 8;
```
mifrandir, line 13: To me `kMaxChannelNumber` should be `7` (based on the name)

However, I agree with the choice of 8. Rename to `kMaxNumChannels`?

mifrandir (changes requested):
Almost there

lib/sensors/mux.cpp
```diff
@@ -21,6 +21,10 @@ Mux<T, N>::~Mux()
 template<typename T, std::uint8_t N>
 core::Result Mux<T, N>::selectChannel(const std::uint8_t channel)
 {
+  if (channel >= kMaxChannelNumber) {
+    log_.log(core::LogLevel::kFatal, "Mux : inputted channel out of range");
```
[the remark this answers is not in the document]
SnickeyX: have updated it to suit our style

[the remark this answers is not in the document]
kshxtij: You don't need to have `Mux: ` within the error message as the logger formats output as 
`timestamp LEVEL[Logger_name] log msg`.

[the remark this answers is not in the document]
SnickeyX: changed, please check now @kshxtij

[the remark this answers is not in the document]
kshxtij: LGTM, we can talk about standardising error and log messages in the meeting at some point

lib/sensors/mux.hpp
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
+#include "i2c_sensors.hpp"
+
+#include <array>
+#include <memory>
```
maxguy2001, line 6: #include <cstdint>?

maxguy2001 (changes requested):

maxguy2001 (approved):

mifrandir (approved):

PR #38, opened 2022-11-07
Nav - Wheel Encoder Preprocessing 

aryansharma19:
implementing preprocessing of encoder's data


lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,5 +1,9 @@
 #include "preprocess_encoders.hpp"
-
+#include "core/types.hpp"
+#include <iterator>
+#include <algorithm>
+#include <array>
+#include <vector>
```
maxguy2001, line 6: No vectors! Everything should have a statically defined size!

lib/navigation/preprocess_encoders.cpp
```diff
@@ -34,7 +38,38 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
...
   documentation as appropriate
   */
-  return {0, 0, 0, 0};
+ core::EncoderData encoder_data_copy = std::copy(encoder_data.begin(), encoder_data.end(), std::back_inserter(encoder_data));
+ std::sort(encoder_data_copy.begin(), encoder_data_copy.end());
+ core::Float median = 0;
+ core::Float q1 = 0.0;
+ core::Float q3 = 0.0;
+ if(encoder_data_copy.size() % 2 == 0){
+  median = (encoder_data_copy[(std::size(encoder_data_copy))/2] + encoder_data_copy[((std::size(encoder_data_copy))/2) + 1])/2.0;
```
maxguy2001, line 47: use `.at(index)` instead of `[index]`

maxguy2001, line 47: Implicit typecasting (hint: look at `static_cast`)

lib/navigation/preprocess_encoders.cpp
```diff
@@ -34,7 +38,38 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
...
+ }
+ else{
+  median = encoder_data_copy[(std::size(encoder_data_copy) + 1)/2];
+ }
+ core::EncoderData array_for_q1 ;
```
maxguy2001, line 52: not informative name. What is the point of this?

lib/navigation/preprocess_encoders.cpp
```diff
@@ -34,7 +38,38 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
...
+ std::copy_if(encoder_data_copy.begin(),encoder_data_copy.end(),array_for_q1.begin(),[](uint32_t i){return i< median;});
```
maxguy2001, line 53: `static_cast` for typecasting. That's C-style, not C++ style

lib/navigation/preprocess_encoders.cpp
```diff
@@ -34,7 +38,38 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
...
+ core::EncoderData array_for_q2;
+ std::copy_if(encoder_data_copy.begin(),encoder_data_copy.end(),array_for_q2.begin(),[](uint32_t i){return i> median;});
+ if(array_for_q1.size() % 2 == 0){
+  q1 = (array_for_q1[(array_for_q1.size())/2] +array_for_q1[((array_for_q1.size())/2) + 1] )/2.0;
```
maxguy2001, line 57: lots of if statements and error handling here just to get upper and lower bounds dependent on if we have an unreliable encoder. Consider defining a function that takes in a copy of the encoder data and returns the median and the upper and lower bounds?

maxguy2001 (changes requested):
Good progress so far!

lib/navigation/preprocess_encoders.cpp
maxguy2001: This will help to avoid writing single functions that are too long. A function in general should not exceed 60 lines of code.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -34,7 +38,38 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
   -also figure out return type/ what we update and update
   documentation as appropriate
   */
-  return {0, 0, 0, 0};
+ core::EncoderData encoder_data_copy = std::copy(encoder_data.begin(), encoder_data.end(), std::back_inserter(encoder_data));
```
maxguy2001, line 41: Looks like this needs to be split into 2 lines. You first have to initialize a container for the new data and then assign it:
`core::EncoderData encoder_data_copy;`
`std::copy(encoder_data.begin(), encoder_data.end(), std::back_inserter(encoder_data_copy);`

lib/navigation/preprocess_encoders.cpp
```diff
@@ -34,9 +38,29 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
   -also figure out return type/ what we update and update
   documentation as appropriate
   */
-  return {0, 0, 0, 0};
+ //core::EncoderData encoder_data_copy = std::copy(encoder_data.begin(), encoder_data.end(), std::back_inserter(encoder_data));
+ //core::EncoderData encoder_data_copy = std::copy(encoder_data.begin(), encoder_data.end())
+ core::EncoderData encoder_data_copy;
+ for(int i = 0;i<core::kNumEncoders;i++){
```
maxguy2001, line 44: std::size_t and ++i

lib/navigation/preprocess_encoders.cpp
```diff
@@ -34,9 +38,29 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
...
+  encoder_data_copy.at(i) = encoder_data.at(i);
+ }
+ core::Float median = calulate_value(encoder_data_copy);
```
maxguy2001, line 47: what is `calculate_value`? Function names should be informative

lib/navigation/preprocess_encoders.cpp
```diff
@@ -34,9 +38,29 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
...
+ if(encoder_data.size() % 2 == 0){
+    median =  ((encoder_data.at(encoder_data.size()/2) + encoder_data.at((encoder_data.size()/2) + 1))/2.0);
+  }
+  else{
+    median = encoder_data.at(encoder_data.size() + 1)/2.0;
+  }
+ core::Float q1 = encoder_data_copy.at(((encoder_data_copy.size() + 1)/4) -1);
```
maxguy2001, line 54: where are these numbers coming from? if the 4 is number of encoders then `core::kNumEncoders`

lib/navigation/preprocess_encoders.hpp
```diff
@@ -15,11 +15,15 @@ class EncodersPreprocessor {
 
   void checkReliable(const core::EncoderData &encoder_data);
 
+  core::Float EncodersPreprocessor::calculating_median(const core::EncoderData encoder_data);
```
maxguy2001, line 18: You don't define class methods with `EncodersPreprocessor::`. You're already in the class. You only have to use that in the cpp file.

maxguy2001 (changes requested):

lib/navigation/preprocess_encoders.cpp
```diff
@@ -34,8 +38,25 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
...
-}
+ //core::EncoderData encoder_data_copy = std::copy(encoder_data.begin(), encoder_data.end(), std::back_inserter(encoder_data));
+ //core::EncoderData encoder_data_copy = std::copy(encoder_data.begin(), encoder_data.end())
+ core::EncoderData encoder_data_copy;
+ core::Float median = 0;
+ for(int i = 0;i<core::kNumEncoders;i++){
+  encoder_data_copy.at(i) = encoder_data.at(i);
+ }
+ if(encoder_data.size() % 2 == 0){
+    median =  ((encoder_data.at(encoder_data.size()/2) + encoder_data.at((encoder_data.size()/2) + 1))/2.0);
```
maxguy2001, line 49: Also be careful here,  it looks like you're doing integer division.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,4 +1,8 @@
 #include "preprocess_encoders.hpp"
+#include "core/types.hpp"
+#include <iterator>
```
maxguy2001, line 3: Don't need iterator

lib/navigation/preprocess_encoders.cpp
maxguy2001: also shouldn't be a loop. Can use std::copy but can help with that on monday

lib/navigation/preprocess_encoders.cpp
```diff
@@ -34,8 +38,34 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
...
+
+ core::EncoderData num_outliers_per_encoder;
+ for(int i =0;i<core::kNumEncoders;i++){
+  if(encoder_data_copy.at(i) > q3 || encoder_data_copy.at(i) < q1){
+    num_outliers_per_encoder.at(i) = num_outliers_per_encoder.at(i) + 1;
+    encoder_data_copy.at(i) = median;   // updating the value of outlier to the median
+  }
+ }
+
+ return {0, 0, 0, 0}; // might be a good idea to create an array of size 2 such that index 0 contains our updated encoder_data and the element at index 1 is an array of the number of outliers per encoder
```
maxguy2001, line 66: No, we have a class member for this. This is better as it is a more efficient way to store data between function calls. Also it would have to be an array of arrays and without defining a struct, it may not be clear where data is coming from. However, it does not seem sensible to define a struct purely for this purpose.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -34,8 +38,34 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
...
+ core::EncoderData encoder_data_copy;
+ core::Float median = 0;
+ for(int i = 0;i<core::kNumEncoders;i++){
+  encoder_data_copy.at(i) = encoder_data.at(i);
+ }
+ if(encoder_data.size() % 2 == 0){
+    median =  ((encoder_data.at(encoder_data.size()/2) + encoder_data.at((encoder_data.size()/2) + 1))/2.0);  //calculating the median of encoder data if the length of the array is even
+                                                                                                               
+  }
+  else{
```
maxguy2001, line 52: repeated functionality. Define a helper function.

maxguy2001: Hint: Look at @BenET01's pr. There is a helper function to get the quartiles and the use of the `Quartiles` struct as a return type.

lib/navigation/preprocess_encoders.hpp
```diff
@@ -15,11 +15,13 @@ class EncodersPreprocessor {
 
   void checkReliable(const core::EncoderData &encoder_data);
 
+
+
   // initialised as {0, 0, 0, 0}, count of consecutive outliers
   std::array<uint16_t, core::kNumEncoders> num_outliers_per_encoder_;
```
maxguy2001, line 21: `std::uint16_t`

maxguy2001 (changes requested):

lib/navigation/preprocess_encoders.cpp
```diff
@@ -19,6 +34,7 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
   return {0, 0, 0, 0};
 }
 
+/*
 core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData encoder_data)
 {
   /*
```
maxguy2001, line 40: this is why build is failing. you've opened up a new comment block within a comment block and the compiler is being incredibly pedantic

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,6 +1,21 @@
...
+#include <algorithm>
+#include <array>
+#include <cmath>  
+#include <numeric>
 
 namespace hyped::navigation {
+  //core::EncoderData reliability_of_encoders;
+  /*Initialising the array with 1(assuming that all the encoders are initially reliable )
+  a value of 0 corresponding at the ith position means that the encoder is unreliable 
+  */
```
maxguy2001, line 12: comment block within a comment block...

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,6 +1,21 @@
 #include "preprocess_encoders.hpp"
+#include "core/types.hpp"
+#include <algorithm>
```
maxguy2001, line 3: surely a lot of these are needed in the header?

lib/navigation/preprocess_encoders.cpp
```diff
@@ -33,10 +49,86 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
   -also has to be able to handle 1 unreliable sensor
   -also figure out return type/ what we update and update
   documentation as appropriate
+  }
   */
-  return {0, 0, 0, 0};
+template<typename T>
```
maxguy2001, line 54: not quite:
```
template<std::size_t N>
std::array<core::Float,3> EncodersPreprocessor::quartiles(std::array<std::uint32_t , N> & encoder_data){ ...
```

maxguy2001: This is I think a better use of templates and then you can make the code adaptable to it taking an array of length 3 or 4 (in code knumiums or knumimus-1 because no hard coding please)

lib/navigation/preprocess_encoders.cpp
```diff
@@ -33,10 +49,86 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
...
+std::array<core::Float,3> EncodersPreprocessor::quartiles(T encoder_data){
+  core::Float median;
```
maxguy2001, line 56: This code is quite messy and repetitive. Consider using a helper function.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -33,10 +49,86 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
...
+  int q3_low = static_cast<int> (std::ceil((3*(encoder_data.size()+1))/4.0));
+  q1 = encoder_data.at(q1_low-1) + (((encoder_data.size()+1)/4.0) - q1_low)*(encoder_data.at(q1_high-1) - encoder_data.at(q1_low-1));
+  q3 = encoder_data.at(q3_high-1) +(((3*(encoder_data.size()+1))/4.0) - q3_low)*(encoder_data.at(q3_high-1) - encoder_data.at(q3_low-1));
+  if(encoder_data.size() % 2 == 0){
+    median = (encoder_data.at((encoder_data.size()/2)-1) + encoder_data.at(((encoder_data.size()/2) + 1))-1)/2.0;
+  }
+  else{
+   median = encoder_data.at(((encoder_data.size()+1)/2)-1);
+  }
+  iqr = q3-q1;
```
maxguy2001, line 75: consts.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -33,10 +49,86 @@ core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData e
...
+  output.at(1) = upper_bound;
+  output.at(2) = lower_bound;
+
+  return output; // return an output of the sequence median,upper_bound,lower_bound
+}
+
+
+core::EncoderData EncodersPreprocessor::detectOutliers(const core::EncoderData encoder_data){
+  const uint8_t num_reliable_encoders = std::accumulate(are_encoders_reliable_.begin(),are_encoders_reliable_.end(),0);
+  core::Float median;
```
maxguy2001, line 88: I feel like you've defined lots of medians in lots of different functions. The median should maybe be calculated once and then used to construct upper and lower bounds, nothing more.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -50,7 +142,18 @@ void EncodersPreprocessor::checkReliable(const core::EncoderData &encoder_data)
 
   -also figure out return type/ what we update and update
   documentation as appropriate
-  */
+  
+
 }
+*/
+
+void EncodersPreprocessor::checkReliable(const core::EncoderData num_outliers_per_encoder){
```
maxguy2001, line 150: Also include check for if number of unreliable encoders > 1 and add a //TODOLater comment to write a fail state. The high level design of this class will be changing marginally to better represent our ability to fail when necessary but this can be implemented slightly later. For now, have a look at #39, its use of templates and some of the comments I left there to try and remove some of the boilerplate code you have repeated here (it's not technically "wrong" but can be more succinct). If you have any questions on what's happening here, you're always welcome to shoot me a message or ask at meetings.

lib/navigation/preprocess_encoders.hpp
```diff
@@ -11,9 +11,17 @@ class EncodersPreprocessor {
...
  private:
+ //extern core::EncoderData reliability_of_encoders;      // some addition
+ //extern core::EncoderData num_outliers_per_encoder;
   core::EncoderData detectOutliers(const core::EncoderData encoder_data);
+  
+  //void checkReliable(const core::EncoderData &encoder_data);
+  void checkReliable(const core::EncoderData num_outliers_per_encoder);
+
+  template<typename T>
+  std::array<core::Float,3> quartiles(T encoder_data);
```
maxguy2001, line 22: Look at how I suggested it be defined in cpp file.

maxguy2001 (changes requested):
A few further changes for now. I think we need to look at using templates and removing repetitive code with helper functions here but good progress.

maxguy2001 (changes requested):

lib/navigation/preprocess_encoders.hpp
```diff
@@ -11,9 +13,21 @@ class EncodersPreprocessor {
...
  private:
+  // extern core::EncoderData reliability_of_encoders;      // some addition
+  // extern core::EncoderData num_outliers_per_encoder;
   core::EncoderData detectOutliers(const core::EncoderData encoder_data);
 
-  void checkReliable(const core::EncoderData &encoder_data);
+  // void checkReliable(const core::EncoderData &encoder_data);
+  void checkReliable(const core::EncoderData &num_outliers_per_encoder);
+
+  // template<typename T>
```
maxguy2001, line 23: delete comments to tidy up code.

maxguy2001 (changes requested):

maxguy2001 (changes requested):

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,13 +1,21 @@
 #include "preprocess_encoders.hpp"
 
+#include <cmath>
+
+#include <algorithm>
+#include <array>
```
maxguy2001, line 6: Array already included in header

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,13 +1,21 @@
...
+#include <numeric>
+
+#include "core/types.hpp"
+
 namespace hyped::navigation {
 
 EncodersPreprocessor::EncodersPreprocessor()
```
maxguy2001, line 13: Can you implement the logger here too pls?

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,13 +1,21 @@
...
 {
   // TODOLater: implement
 }
 
-core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
+core::EncoderData EncodersPreprocessor::processData(const core::EncoderData encoder_data)
```
maxguy2001, line 18: this needs to be able to notify of failure so return type should be optional.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +24,91 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
-  - check how many times an individual encoder
-  has been an outlier in a row (outlier encoders)
-  -if num_outliers_per_encoder_[i] > n (tbd), mark encoder as unreliable
-  in reliable encoders.
-  -if number unreliable in reliable_encoders > 1, fail state
-
-  -also figure out return type/ what we update and update
-  documentation as appropriate
-  */
+  for (int i = 0; i < core::kNumEncoders; i++) {
```
maxguy2001, line 107: for loop style...

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +24,91 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
+    if (num_outliers_per_encoder_.at(i) > max_consecutives) {  // for now assuming n to be 10
```
maxguy2001, line 108: naming

lib/navigation/preprocess_encoders.hpp
```diff
@@ -12,8 +14,15 @@ class EncodersPreprocessor {
 
  private:
   core::EncoderData detectOutliers(const core::EncoderData encoder_data);
+  std::uint8_t max_consecutives = 10;
```
maxguy2001, line 17: docstrings and comments in this file

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +24,91 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
+  core::EncoderData encoder_data_copy;
+  std::copy(encoder_data.begin(), encoder_data.end(), encoder_data_copy.begin());
+  Quartile quartiles;
+  if (num_reliable_encoders_ == core::kNumEncoders - 1) {
+    std::array<uint32_t, core::kNumEncoders - 1> reliable_sensors_data;
+    std::size_t counter = 0;
+    for (std::size_t i = 0; i < encoder_data.size(); ++i) {
+      if (are_encoders_reliable_.at(i) == true) {
+        reliable_sensors_data.at(counter) = encoder_data.at(i);
+        counter++;
```
maxguy2001, line 72: pre incrementing

maxguy2001 (changes requested):
Good progress! Not quite ready to be converted from draft but starting to look good!

lib/navigation/preprocess_encoders.hpp
```diff
@@ -1,3 +1,5 @@
+#include "consts.hpp"
+
```
maxguy2001, line 2: If you're using optional return types, you'll need to add `#include <optional>` at the top of the file here

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +31,89 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
-
-  -also has to be able to handle 1 unreliable sensor
-  -also figure out return type/ what we update and update
-  documentation as appropriate
-  */
-  return {0, 0, 0, 0};
+  core::EncoderData encoder_data_copy;
+  std::copy(encoder_data.begin(), encoder_data.end(), encoder_data_copy.begin());
+  Quartile quartiles;
+  if (num_reliable_encoders_ == core::kNumEncoders - 1) {
```
maxguy2001, line 67: This does not cover the case if there is more than 1 unreliable encoder (hint: use an `if`, `else if`, `else` statement).

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +31,89 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
+      }
+    }
+    quartiles = getQuartiles(reliable_data);
+  } else {
+    core::EncoderData reliable_data;
+    std::copy(
+      encoder_data.begin(), encoder_data.end(), reliable_data.begin());
+    quartiles = getQuartiles(reliable_data); 
+  }
+  const core::Float iqr         = quartiles.q3 - quartiles.q1;
```
maxguy2001, line 83: I know its pedantic but maybe `interquartile_range` or a comment explaining what it is for someone reading this without context.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +31,89 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
-  - check how many times an individual encoder
-  has been an outlier in a row (outlier encoders)
-  -if num_outliers_per_encoder_[i] > n (tbd), mark encoder as unreliable
-  in reliable encoders.
-  -if number unreliable in reliable_encoders > 1, fail state
-
-  -also figure out return type/ what we update and update
-  documentation as appropriate
-  */
+  for (int i = 0; i < core::kNumEncoders; ++i) {
```
maxguy2001, line 106: `std::size_t`

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +31,89 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
+    if (encoder_outliers.at(i) > max_consecutives) {  // for now assuming n to be 10
```
maxguy2001, line 107: maybe rename `max_consecutives` for clarity

maxguy2001: Also `max_consecutives` is a class member so needs an underscore after it (just a naming convention).

maxguy2001 (changes requested):
Good work, definitely moving in the right direction! Currently doesn't build on mac (hint: look at the log on the github task and fix it in the codebase!). Still need docstrings on class functions you have defined and tidying up comments etc., but almost there! Should be more or less on track for getting merged by the end of next week hopefully.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,13 +1,28 @@
 #include "preprocess_encoders.hpp"
 
+#include <cmath>
+
+#include <algorithm>
+#include <numeric>
```
maxguy2001, line 6: not used

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,13 +1,28 @@
...
+#include <optional>
+
+#include "core/types.hpp"
+
 namespace hyped::navigation {
 
-EncodersPreprocessor::EncodersPreprocessor()
-{
+//EncodersPreprocessor::EncodersPreprocessor()
```
maxguy2001, line 13: You can tidy up the comments (everywhere else too).

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +31,101 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
+  const auto encoder_data = detectOutliers(raw_encoder_data);
+  SensorChecks sensors  = checkReliable();
+  if(!encoder_data.has_value() || sensors == SensorChecks::kUnacceptable){
+    return std::nullopt;
+  } else{
+    return encoder_data;
+  }
+}
+
+template<std::size_t N>
```
maxguy2001, line 43: Templated functions have to be moved into the header file.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +31,101 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
+core::Float EncodersPreprocessor::getSpecificQuartile(const std::array<std::uint32_t, N> &reliable_data, const core::Float quartile_percent){
+
+  const core::Float quartile_index = (num_reliable_encoders_ - 1)*quartile_percent;
+  const std::uint8_t quartile_high = static_cast<int>(std::ceil(quartile_index));
```
maxguy2001, line 47: `static_cast<std::uint8_t>`

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +31,101 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
+  const std::uint8_t quartile_low = static_cast<int>(std::floor(quartile_index));
```
maxguy2001, line 48: `static_cast<std::uint8_t>`

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +31,101 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
+  const core::Float quartile = (reliable_data.at(quartile_high) + reliable_data.at(quartile_low))/2.0;
+  return quartile;
+}
+
+template<std::size_t N>
```
maxguy2001, line 53: Templated function, needs to be moved to header file.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +31,101 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
+        ++counter;
+      }
+    }
+    quartiles = getQuartiles(reliable_data);
+    interquartile_range        = quartiles.q3 - quartiles.q1;
+    upper_bound = quartiles.median + 1.2 * interquartile_range ;
+    lower_bound = quartiles.median - 1.2 * interquartile_range ;
+  }
+  else{
+    logger_.log(core::LogLevel::kFatal, "Sensors have become unreliable");
```
maxguy2001, line 95: What sensors? Error comments have to be as informative as possible

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +31,101 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
+    return std::nullopt;
+  }
+
+  for (std::size_t i = 0; i < encoder_data_copy.size(); ++i) {
+    if (encoder_data_copy.at(i) > upper_bound || encoder_data_copy.at(i) < lower_bound || encoders_reliable_.at(i) == false) {
+      encoder_data_copy.at(i)         = quartiles.median;
+
+      if(encoders_reliable_.at(i)){
+        encoder_outliers_.at(i) = encoder_outliers_.at(i) + 1;
```
maxguy2001, line 104: either `encoder_outliers_.at(i) += 1` or `++encoder_outliers_.at(i)`.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +31,101 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
-  has been an outlier in a row (outlier encoders)
-  -if num_outliers_per_encoder_[i] > n (tbd), mark encoder as unreliable
-  in reliable encoders.
-  -if number unreliable in reliable_encoders > 1, fail state
-
-  -also figure out return type/ what we update and update
-  documentation as appropriate
-  */
+  for (std::size_t i = 0; i < core::kNumEncoders; ++i) {
+    if (encoder_outliers_.at(i) > max_consecutive_outliers_) {  // for now assuming n to be 10
```
maxguy2001, line 119: You need to turn this into an and statement to also check if it has already been classified as unreliable. Otherwise you will set an unreliable sensor as still false but then still subtract 1 from the number of reliable imus which will cause critical failure every time a single sensor fails.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -16,41 +31,101 @@ core::EncoderData EncodersPreprocessor::processData(const core::EncoderData)
...
+      encoders_reliable_.at(i) = false;     // the encoder is now unrealiable
+      --num_reliable_encoders_;
+    }
+  }
+  if(num_reliable_encoders_ < core::kNumEncoders - 1){
+    logger_.log(core::LogLevel::kFatal, "Sensors have become unreliable");
```
maxguy2001, line 125: informative error messages. If the pod crashes, burns and blows up we wanna know why...

lib/navigation/preprocess_encoders.hpp
```diff
@@ -1,25 +1,42 @@
...
-  EncodersPreprocessor();
+  //EncodersPreprocessor();
+  EncodersPreprocessor(core::ILogger &logger);
 
-  core::EncoderData processData(const core::EncoderData encoder_data);
+  std::optional<core::EncoderData> processData(const core::EncoderData encoder_data);
 
  private:
-  core::EncoderData detectOutliers(const core::EncoderData encoder_data);
+  std::optional<core::EncoderData> detectOutliers(const core::EncoderData encoder_data);
```
maxguy2001, line 18: Docstrings here pls. It's important that future navigation members know what's going on!

maxguy2001 (changes requested):
A few more changes to be implemented. Looking great now though!

lib/navigation/preprocess_encoders.hpp
```diff
@@ -1,25 +1,42 @@
...
   // initialised as {0, 0, 0, 0}, count of consecutive outliers
-  std::array<uint16_t, core::kNumEncoders> num_outliers_per_encoder_;
+  std::array<uint16_t, core::kNumEncoders> encoder_outliers_;
 
   // initialised as all true, bool mask of reliable sensors
-  std::array<bool, core::kNumEncoders> are_encoders_reliable_;
+  std::array<bool, core::kNumEncoders> encoders_reliable_;
+
+  core::ILogger &logger_;
+
```
mifrandir, line 39: whitespace

lib/navigation/preprocess_encoders.hpp
```diff
@@ -1,25 +1,42 @@
+#include "consts.hpp"
+
 #include <array>
 #include <cstdint>
 
 #include "core/types.hpp"
+#include "core/logger.hpp"
 
 namespace hyped::navigation {
 class EncodersPreprocessor {
```
mifrandir, line 10: tidy up the ordering in this class a little

we usually go with public (constructors, methods, members) and then private (constructors, methods, members)

lib/navigation/preprocess_encoders.hpp
```diff
@@ -1,25 +1,63 @@
...
+    const core::Float quartile_index = (num_reliable_encoders_ - 1)*quartile_percent;
+    const std::uint8_t quartile_high = static_cast<std::uint8_t>(std::ceil(quartile_index));
+    const std::uint8_t quartile_low = static_cast<std::uint8_t>(std::floor(quartile_index));
+    const core::Float quartile = (reliable_data.at(quartile_high) + reliable_data.at(quartile_low))/2.0;
+    return quartile;
+}
+
+ template<std::size_t N>
+ Quartile getQuartiles(std::array<std::uint32_t, N> &reliable_data){
+    std::sort(reliable_data.begin(), reliable_data.end());
```
maxguy2001, line 43: Look at the logs and try figure out why it doesn't build any more. Hint: Look at this line and then look at the included headers at the top of this file to see what you might have been missing...

maxguy2001 (changes requested):

lib/io/i2c.cpp
```diff
@@ -4,6 +4,7 @@
 #include <unistd.h>
 
 #include <sys/ioctl.h>
+#include <stdio.h>
```
maxguy2001, line 7: `cstdio` is already included in the header for this file and also contains a definition of `snprintf` so this include is not necessary.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,13 +1,29 @@
 #include "preprocess_encoders.hpp"
 
+//#include <cmath>
+
+#include <algorithm>
+#include <numeric>
+#include <optional>
```
maxguy2001, line 7: Look at your includes here. Anything included in the cpp file does not need to be included in the hpp file

maxguy2001 (changes requested):

maxguy2001 (approved):
Looks good!

test/navigation/preprocess_encoders.cpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+TEST(Encoder, equal_data)
+{
+  utils::ManualTime manual_time;
+  core::Logger logger("test", core::LogLevel::kFatal, manual_time);
+  navigation::EncodersPreprocessor encoder_processer(logger);
+  core::EncoderData data   = {1, 1, 1, 1};
+  core::EncoderData answer = {1, 1, 1, 1};
+  auto final_data          = encoder_processer.processData(data);
```
mifrandir, line 40: `const`s

test/navigation/preprocess_encoders.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <gtest/gtest.h>
+
+#include <core/logger.hpp>
+#include <core/types.hpp>
+#include <navigation/preprocess_encoders.hpp>
+#include <utils/manual_time.hpp>
+
+namespace hyped::test {
+
+core::Float epsilon = 1e-5;
```
mifrandir, line 14: `static constexpr core::Float kEpsilon = ...`

move this into a header such as `test/navigation/utils.hpp` and have other files include it, as required

mifrandir (changes requested):

lib/navigation/preprocess_encoders.cpp
```diff
@@ -2,55 +2,99 @@
...
+    upper_bound         = quartiles.median + 1.2 * interquartile_range;
+    lower_bound         = quartiles.median - 1.2 * interquartile_range;
+  }
+  // Case 3: More than one sensors have become unreliable, stopping everything and entering into
+  // fail State.
+  else {
+    logger_.log(core::LogLevel::kFatal,
+                "Number of unreliable encoder sensors have exceeded the threshold");
+    return std::nullopt;
+  }
```
TomLonergan03, line 65: i would move this to a guard clause before the if statements, then `else if` on line 45 can just be `if`

lib/navigation/preprocess_encoders.cpp
```diff
@@ -2,55 +2,99 @@
...
+  for (std::size_t i = 0; i < encoder_data_copy.size(); ++i) {
+    // replacing the ouliers or data from faulty sensors with the median of the dataset.
+    if (encoder_data_copy.at(i) > upper_bound || encoder_data_copy.at(i) < lower_bound
+        || encoders_reliable_.at(i) == false) {
+      encoder_data_copy.at(i) = quartiles.median;
+
+      if (encoders_reliable_.at(i)) { ++encoder_outliers_.at(i); }
+    } else {
+      encoder_outliers_.at(i) = 0;
+    }
```
TomLonergan03, line 76: this could probably be structured better, seems weird to have one outer condition being `encoders_reliable_.at(i) == false` then also checking within that condition if `encoders_reliable_.at(i) == true`

lib/navigation/preprocess_encoders.cpp
```diff
@@ -2,55 +2,99 @@
 
 namespace hyped::navigation {
 
-EncodersPreprocessor::EncodersPreprocessor()
+EncodersPreprocessor::EncodersPreprocessor(core::ILogger &logger)
+    : logger_(logger),
+      encoder_outliers_({0, 0, 0, 0}),
+      encoders_reliable_({true, true, true, true}),
+      num_reliable_encoders_(core::kNumImus)
```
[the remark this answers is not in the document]
TomLonergan03: looking at the rest of the PR this should be `core::kNumEncoders`

TomLonergan03 (changes requested):

lib/navigation/preprocess_encoders.cpp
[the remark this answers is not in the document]
maxguy2001: Good spot! The dangers of copy/pasting code @aryansharma19!!!

lib/navigation/preprocess_encoders.cpp
```diff
@@ -2,55 +2,99 @@
...
-  - call detectOutliers. Return of that function is return of this function
-  - call checkRelaible
-
-  */
-  return {0, 0, 0, 0};
+  const auto encoder_data = detectOutliers(raw_encoder_data);
+  SensorChecks sensors    = checkReliable();
+  if (!encoder_data.has_value() || sensors == SensorChecks::kUnacceptable) {
+    return std::nullopt;
+  } else {
```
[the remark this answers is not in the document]
maxguy2001: This was just done for clarity to show it was the default case.

lib/navigation/preprocess_encoders.cpp
maxguy2001: No because it directly assigns values to an instantiated object depending on the conditions. Surely this is exactly the use case for `if`, `else if`, `else`?

lib/navigation/preprocess_encoders.cpp
maxguy2001: This extra statement is necessary because otherwise if we get one failure, the system will keep counting that failure on every following iteration, leading to an unnecessary critical failure...

lib/navigation/preprocess_encoders.cpp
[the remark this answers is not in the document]
maxguy2001: This is why we want lots of eyes on code!

lib/navigation/preprocess_encoders.cpp
TomLonergan03: idk about `if`, `else if` but there is definitely no need to set up all the variables and copy the encoder data around if you can already determine `num_reliable_encoders_ < core::kNumEncoders - 1` as soon as the function is called

maxguy2001: I mean we could check this closer to the start but the reason we put it here was to give a default case of failure in case of bugs, etc. This is just an additional precaution to avoid undefined behavior in the logic of this function (we should never reach this if statement but its good to have it surely?). As for the copies, this is because the thresholds should be different depending on if we have unreliable sensors already and the readings need to be stored in different sized arrays since we don't want to be using anything that does not have a statically defined size.

TomLonergan03: it makes sense to have, but it should be the first thing in the function. i understand why the copy of the data is made, but in the case it is already provable at the start of the function that `num_reliable_encoders_ < core::kNumEncoders - 1` and therefore the function should exit there. making the copy means allocating memory in a case where we know it is going to do nothing but be deallocated immediately after, which is pointless

maxguy2001: The container which we copy the array to is also an array with a statically defined size. Therefore the space is already allocated to stack memory so we're not actually allocating memory, just reassigning the values stored at an address (I think? @Fayz-S?). I fully see what you mean but also I feel like it's just removing a safeguard in the form of a default fail state?

lib/navigation/preprocess_encoders.cpp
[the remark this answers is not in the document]
mifrandir: i agree with Tom

test/navigation/preprocess_encoders.cpp
kshxtij: Isn't this already in core::types::kEpsilon?

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,103 @@
...
 
+#include "core/types.hpp"
+
 namespace hyped::navigation {
 
-EncodersPreprocessor::EncodersPreprocessor()
+EncodersPreprocessor::EncodersPreprocessor(core::ILogger &logger)
+    : logger_(logger),
+      num_consecutive_outliers_per_encoder_({0, 0, 0, 0}),
+      is_reliable_per_encoder_({true, true, true, true}),
```
mifrandir, line 11: Actually, the name the other PR went with is different so let's change this for consistency.

`are_encoders_reliable_`

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,103 @@
...
-  -set outlier points as median
-
-  -also has to be able to handle 1 unreliable sensor
-  -also figure out return type/ what we update and update
-  documentation as appropriate
-  */
-  return {0, 0, 0, 0};
+  const auto statistics = getStatistics(encoder_data);
+  if (!statistics) {
+    logger_.log(core::LogLevel::kFatal, "Failed to obtain statistics for measurement");
```
TomLonergan03, line 68: this should return `nullopt`

TomLonergan03 (changes requested):

TomLonergan03 (approved):

kshxtij (approved):

lib/navigation/preprocess_encoders.hpp
```diff
@@ -1,27 +1,104 @@
...
+    core::Float lower_bound;
+  };
+
+  /**
+   * @brief calculate the upper and lower bounds as well as the median for
+   * the input data array
+   *
+   * @param encoder_data
+   * @return std::optional<Statistics>
+   */
```
mifrandir, line 41: remove, no valuable information is contained in this

just from reading `struct Statistics { ... }` and the signature all of these things can be inferred

lib/navigation/preprocess_encoders.hpp
```diff
@@ -1,27 +1,104 @@
...
+  std::optional<Statistics> getStatistics(const core::EncoderData &encoder_data) const;
+
+  /**
+   * @brief tidies up the received data by detecting the outliers or data received from faulty
+     sensors and replaces them with median value of the dataset; if there are too many
+     outliers or faulty sensors then this function fails
```
mifrandir, line 47: be brief, people don't want to read a novel; keep the non-obvious parts and remove everything else

"detects new outliers and replaces values from faulty sensors with the median; fails if not enough reliable data is available"

if people want to know how exactly it works, they'll go read the implementation (which I would advise everyone to do anyways)

mifrandir (changes requested):
didn't read the tests; just small changes otherwise

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,104 @@
...
+        || encoder_data.at(i) < statistics->lower_bound) {
+      sanitised_data.at(i) = statistics->median;
+      ++num_consecutive_outliers_per_encoder_.at(i);
+    } else {
+      num_consecutive_outliers_per_encoder_.at(i) = 0;
+    }
+    if (!are_encoders_reliable_.at(i)) { sanitised_data.at(i) = statistics->median; }
+  }
+  if (checkReliable() == SensorChecks::kUnacceptable) { return std::nullopt; }
+  return sanitised_data;
```
mifrandir, line 83: Reading this again, I think we can do better:

1. Update `num_consecutive_outliers_per_encoder`
2. `checkReliable`
3. Update `sanitised_data` basedo n the new values in `are_encoders_reliable`

mifrandir (changes requested):

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,109 @@
...
+    if(sanitised_data.at(i) > statistics->upper_bound
+        || sanitised_data.at(i) < statistics->lower_bound){
+          ++num_consecutive_outliers_per_encoder_.at(i);
+        }
+    else{
+      num_consecutive_outliers_per_encoder_.at(i) = 0;
+    }
+  }
+  if (checkReliable() == SensorChecks::kUnacceptable) { return std::nullopt; }
+  return sanitised_data;
```
mifrandir, line 82: everything below this is dead

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,109 @@
...
-  -also figure out return type/ what we update and update
-  documentation as appropriate
-  */
-  return {0, 0, 0, 0};
+  const auto statistics = getStatistics(encoder_data);
+  if (!statistics) {
+    logger_.log(core::LogLevel::kFatal, "Failed to obtain statistics for measurement");
+    return std::nullopt;
+  }
+  auto sanitised_data = encoder_data;
```
mifrandir, line 71: i think this wants to go below? for now we are fine just dealing with `encoder_data`, i believe

mifrandir (changes requested):

lib/navigation/preprocess_encoders.cpp
aryansharma19: won't that be only true if the "if" condition is satisfied

mifrandir: no, inside the if you return `std::nullopt`

outside the if you return `sanitised_data`

either way, you return

mifrandir: honestly, i'm shocked that this compiles even with `-Wpedantic` on... time for clang-analyzer or cppcheck @kshxtij

lib/navigation/preprocess_encoders.cpp
maxguy2001: It's used directly below?

lib/navigation/preprocess_encoders.cpp
maxguy2001: Surely checkReliable should be the last thing we do as the decisions made in checkRelaible are made based on data updated in sanitise?

lib/navigation/preprocess_encoders.hpp
```diff
@@ -1,27 +1,96 @@
...
+   * @return the quantile that we require
+   */
+  template<std::size_t N>
+  core::Float getSpecificQuantile(const std::array<std::uint32_t, N> &reliable_data,
+                                  const core::Float fraction) const
+  {
+    const core::Float theoretical_index = (num_reliable_encoders_ - 1) * fraction;
+    const std::size_t low_index         = static_cast<std::size_t>(std::floor(theoretical_index));
+    const std::size_t high_index        = static_cast<std::size_t>(std::ceil(theoretical_index));
+    const auto low                      = static_cast<core::Float>(reliable_data.at(low_index));
```
BenET01, line 69: why use auto when static casting to a type?

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,106 @@
...
+    return std::nullopt;
+  }
+  if (num_reliable_encoders_ == core::kNumEncoders) {
+    auto reliable_data                    = encoder_data;
+    const Quartile quartiles              = getQuartiles(reliable_data);
+    const core::Float interquartile_range = quartiles.q3 - quartiles.q1;
+    return {
+      {.median      = quartiles.median,
+       .upper_bound = quartiles.median + static_cast<core::Float>(1.5) * interquartile_range,
+       .lower_bound = quartiles.median - static_cast<core::Float>(1.5) * interquartile_range}};
```
SnickeyX, line 44: not sure if these static_casts are necessary?

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,106 @@
...
+  } else {
+    std::array<uint32_t, core::kNumEncoders - 1> reliable_data;
```
SnickeyX, line 46: `std::uint32_t`

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,106 @@
...
+        reliable_data.at(j) = encoder_data.at(i);
+        ++j;
+      }
+    }
+    const Quartile quartiles              = getQuartiles(reliable_data);
+    const core::Float interquartile_range = quartiles.q3 - quartiles.q1;
+    return {
+      {.median      = quartiles.median,
+       .upper_bound = quartiles.median + static_cast<core::Float>(1.2) * interquartile_range,
+       .lower_bound = quartiles.median - static_cast<core::Float>(1.2) * interquartile_range}};
```
SnickeyX, line 59: ^

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,106 @@
...
-  num_outliers_per_encoder_[i] += 1
-  -set outlier points as median
-
-  -also has to be able to handle 1 unreliable sensor
-  -also figure out return type/ what we update and update
-  documentation as appropriate
-  */
-  return {0, 0, 0, 0};
+  const auto statistics = getStatistics(encoder_data);
+  if (!statistics) {
```
SnickeyX, line 67: maybe `measurement_statistics` ?

test/navigation/preprocess_encoders.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+namespace hyped::test {
+
+bool checkArrayEquality(core::EncoderData &encoder_data_a, core::EncoderData &encoder_data_b)
+{
+  if (encoder_data_a.size() != encoder_data_b.size()) { return false; }
+  for (std::size_t i = 0; i < encoder_data_a.size(); ++i) {
+    if (std::fabs(encoder_data_a.at(i) - encoder_data_b.at(i)) > core::kEpsilon) { return false; }
+  }
+
```
SnickeyX, line 20: empty line

SnickeyX (changes requested):

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,106 @@
...
+
 namespace hyped::navigation {
 
-EncodersPreprocessor::EncodersPreprocessor()
+EncodersPreprocessor::EncodersPreprocessor(core::ILogger &logger)
+    : logger_(logger),
+      num_consecutive_outliers_per_encoder_({0, 0, 0, 0}),
+      are_encoders_reliable_({true, true, true, true}),
+      num_reliable_encoders_(core::kNumEncoders),
+      max_consecutive_outliers_(10)
```
TomLonergan03, line 13: this should be a `constexpr` instead of a `const` class member

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,106 @@
...
+  }
+  if (num_reliable_encoders_ == core::kNumEncoders) {
+    auto reliable_data                    = encoder_data;
+    const Quartile quartiles              = getQuartiles(reliable_data);
+    const core::Float interquartile_range = quartiles.q3 - quartiles.q1;
+    return {
+      {.median      = quartiles.median,
+       .upper_bound = quartiles.median + static_cast<core::Float>(1.5) * interquartile_range,
+       .lower_bound = quartiles.median - static_cast<core::Float>(1.5) * interquartile_range}};
+  } else {
```
TomLonergan03, line 45: don't need the `else`

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,106 @@
...
-  */
-  return {0, 0, 0, 0};
+  const auto statistics = getStatistics(encoder_data);
+  if (!statistics) {
+    logger_.log(core::LogLevel::kFatal, "Failed to obtain statistics for measurement");
+    return std::nullopt;
+  }
+  auto sanitised_data = encoder_data;
+  for (std::size_t i = 0; i < sanitised_data.size(); ++i) {
+    if (sanitised_data.at(i) > statistics->upper_bound
```
TomLonergan03, line 73: dereference `statistics` into a variable once it's confirmed not `std::nullopt` instead of using `->`

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,106 @@
...
+  }
+  if (checkReliable() == SensorChecks::kUnacceptable) { return std::nullopt; }
+  for (std::size_t i = 0; i < are_encoders_reliable_.size(); ++i) {
+    if (!are_encoders_reliable_.at(i)) { sanitised_data.at(i) = statistics->median; }
+  }
+  return sanitised_data;
 }
 
-void EncodersPreprocessor::checkReliable(const core::EncoderData &encoder_data)
+SensorChecks EncodersPreprocessor::checkReliable()
```
TomLonergan03, line 88: is there a reason for `SensorChecks` instead of `Result`? this could be argued either way tbh

lib/navigation/preprocess_encoders.hpp
```diff
@@ -1,27 +1,96 @@
...
 
+#include <cmath>
+
+#include <algorithm>
 #include <array>
 #include <cstdint>
+#include <optional>
 
+#include "core/logger.hpp"
 #include "core/types.hpp"
```
TomLonergan03, line 12: `<core/logger.hpp>` and `<core/types.hpp>` as these aren't relative addresses

lib/navigation/preprocess_encoders.hpp
```diff
@@ -1,27 +1,96 @@
...
+   * @param fraction in [0, 1] corresponding to the quantile (e.g. 0.25 for the 25th percentile)
+   * @return the quantile that we require
+   */
+  template<std::size_t N>
+  core::Float getSpecificQuantile(const std::array<std::uint32_t, N> &reliable_data,
+                                  const core::Float fraction) const
+  {
+    const core::Float theoretical_index = (num_reliable_encoders_ - 1) * fraction;
+    const std::size_t low_index         = static_cast<std::size_t>(std::floor(theoretical_index));
+    const std::size_t high_index        = static_cast<std::size_t>(std::ceil(theoretical_index));
```
TomLonergan03, line 68: `auto` for these 2

TomLonergan03 (changes requested):

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,106 @@
...
+{
+}
+
+std::optional<core::EncoderData> EncodersPreprocessor::processData(
+  const core::EncoderData &raw_encoder_data)
 {
-  // TODOLater: implement
+  const auto encoder_data = sanitise(raw_encoder_data);
+  const auto sensors      = checkReliable();
+  if (sensors == SensorChecks::kUnacceptable) { return std::nullopt; }
```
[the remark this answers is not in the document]
aryansharma19: I guess the checkReliable() function takes care of that.

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,56 +1,106 @@
...
-  /*
-  TODOLater: implement
-  basic plan:
-  - call detectOutliers. Return of that function is return of this function
-  - call checkRelaible
-
-  */
-  return {0, 0, 0, 0};
+  if (num_reliable_encoders_ > core::kNumEncoders
+      || num_reliable_encoders_ < core::kNumEncoders - 1) {
```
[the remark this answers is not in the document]
aryansharma19: I don't think so, because kNumEncoders represents the number of encoder sensors that we have initially and num_reliable_encoders represents the number of encoder sensors that are working at the moment. As we go on the sensors can either keep working properly or stop working. That's why num_reliable_encoders is only decremented and can never be greater than kNumEncoders

lib/navigation/preprocess_encoders.cpp
aryansharma19: what about statistic_values ?

lib/navigation/preprocess_encoders.hpp
aryansharma19: why do we need auto?

TomLonergan03: it's clear from the `static_cast<std::size_t>` that the type is `std::size_t` so there is no need to duplicate it right next to it

lib/navigation/preprocess_encoders.cpp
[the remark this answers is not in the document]
SnickeyX: exactly, but you have a check to see if it is greater which is then redundant?

lib/navigation/preprocess_encoders.cpp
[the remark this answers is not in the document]
SnickeyX: i think it would log when more than one encoder has failed, but I'd think we want to know even when a single encoder has failed somewhere (it looks like you could log this right after you decrement the `num_reliable_encoders` actually)

lib/navigation/preprocess_encoders.cpp
SnickeyX: actually `encoder_statistics`  would be clear and descriptive

lib/navigation/consts.hpp
```diff
@@ -19,6 +19,12 @@ enum class SensorChecks { kUnacceptable = 0, kAcceptable };
 
 inline core::Trajectory zero_trajectory = {0, 0, 0};
 
+struct Quartile {
+  core::Float q1;
+  core::Float median;
+  core::Float q3;
```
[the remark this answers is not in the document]
maxguy2001: second quantile is median but yeah we should probably use the full names for first and third quartiles @aryansharma19.

lib/navigation/preprocess_encoders.cpp
maxguy2001: We initialized like this for the sake of consistency against other values but could do this alternatively if this is the preferred method?

lib/navigation/preprocess_encoders.cpp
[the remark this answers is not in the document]
maxguy2001: Yeah don't be afraid of using the log. Probably use the `kWarn` level for if just one sensor becomes unreliable, `kCritical` for any fail states and `kDebug` otherwise I guess?

lib/navigation/preprocess_encoders.cpp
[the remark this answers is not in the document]
maxguy2001: It's just there to catch any bugs I believe, I'm pretty sure this was implemented by @mifrandir? The number of unreliable encoders is only ever reduced but we don't explicitly check for any other values elsewhere in the code so would lead to some dangerous behavior if this isn't caught.

lib/navigation/preprocess_encoders.cpp
maxguy2001: Yeah just the naming is much more appropriate for what this does because we can have a sensor already failed which is "acceptable" but not really a resounding "success"...

lib/navigation/preprocess_encoders.hpp
maxguy2001: But if you directly static cast to a type, why then just call it `auto`? Surely you're as well using the type you just explicitly casted to?

TomLonergan03: I think there is a discussion to be had on Monday about what exactly gets `auto`ed + various other minor standards. This is fine as it is for now

lib/navigation/preprocess_encoders.cpp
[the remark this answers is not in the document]
SnickeyX: since that is completely undefined behavior I think it should be dealt with separately (like a different kFatal log to emphasize this) than just logging that its an unsuitable amount of encoders

lib/navigation/preprocess_encoders.cpp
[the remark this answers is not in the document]
SnickeyX: `kWarn` coming from #91 soon™

lib/navigation/preprocess_encoders.cpp
SnickeyX: this would lead to a runtime error if its not dereferenced

lib/navigation/preprocess_encoders.cpp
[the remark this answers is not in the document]
maxguy2001: So once #91 is merged into main, merge main into this branch @aryansharma19.

lib/navigation/preprocess_encoders.cpp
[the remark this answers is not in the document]
maxguy2001: Fair point. @aryansharma19?

lib/navigation/preprocess_encoders.cpp
TomLonergan03: yea that's fair

TomLonergan03 (approved):
i'm happy with this

lib/navigation/preprocess_encoders.cpp
```diff
@@ -2,55 +2,104 @@
...
-  -if number unreliable in reliable_encoders > 1, fail state
+  const auto encoder_statistics_optional = getStatistics(encoder_data);
+  if (!encoder_statistics_optional) {
+    logger_.log(core::LogLevel::kFatal, "Failed to obtain statistics for measurement");
+    return std::nullopt;
+  }
+  const auto encoder_statistics = *encoder_statistics_optional;
+  auto sanitised_data           = encoder_data;
+  for (std::size_t i = 0; i < sanitised_data.size(); ++i) {
+    if (sanitised_data.at(i) > encoder_statistics.upper_bound
```
maxguy2001, line 67:
```
if(sanitised_data.at(i) > encoder_statistics.upper_bound && are_encoders_reliable.at(i) ||  sanitised_data.at(i) < encoder_statistics.lower_bound && are_encoders_reliable.at(i))
```
?
Currently if `encoders.at(i)` are unreliable, we keep adding to `num_consecutive_outliers_per_encoder.at(i)` which has the potential to get too big for a `uint16_t` to contain it on an extended run.

maxguy2001: Or break it up or apply this logic more appropriately.

maxguy2001 (changes requested):

lib/navigation/preprocess_encoders.cpp
```diff
@@ -2,55 +2,106 @@
...
+      sanitised_data.at(i) = encoder_statistics.median;
+    } else if (sanitised_data.at(i) > encoder_statistics.upper_bound
+               || sanitised_data.at(i) < encoder_statistics.lower_bound) {
+      ++num_consecutive_outliers_per_encoder_.at(i);
+      sanitised_data.at(i) = encoder_statistics.median;
+    } else {
+      num_consecutive_outliers_per_encoder_.at(i) = 0;
+    }
+  }
+  if (checkReliable() == SensorChecks::kUnacceptable) { return std::nullopt; }
```
SnickeyX, line 77: why are we doing this here as well? looking at `processData`, the call to `sanitise` is followed by `checkReliable` already. Either remove it from here or from processData

lib/navigation/preprocess_encoders.cpp
```diff
@@ -2,55 +2,106 @@
...
+}
 
-  -also figure out return type/ what we update and update
-  documentation as appropriate
-  */
+SensorChecks EncodersPreprocessor::checkReliable()
+{
+  if (num_reliable_encoders_ > core::kNumEncoders) {
+    logger_.log(core::LogLevel::kFatal,
+                "Number of reliable encoders is greater than number of encoders");
```
SnickeyX, line 88: "...than the total number of encoders"

lib/navigation/preprocess_encoders.cpp
```diff
@@ -2,55 +2,106 @@
...
+  }
+  for (std::size_t i = 0; i < core::kNumEncoders; ++i) {
+    // changes reliable sensor to false if max consecutive outliers are reached
+    if (num_consecutive_outliers_per_encoder_.at(i) > max_consecutive_outliers_
+        && are_encoders_reliable_.at(i)) {
+      are_encoders_reliable_.at(i) = false;  // the encoder is now unrealiable
+      logger_.log(core::LogLevel::kWarn, "Sensor %d is now labelled as unreliable", i);
```
SnickeyX, line 95: "Encoder %d....", sensor is very vague

test/navigation/preprocess_encoders.cpp
```diff
@@ -0,0 +1,65 @@
...
+  navigation::EncodersPreprocessor encoder_processer(logger);
+  const core::EncoderData data = {6, 1, 1, 1};
+  for (std::size_t i; i < 12; ++i) {
+    encoder_processer.processData(data);
+  }
+  core::EncoderData answer = {1, 1, 1, 1};
+  auto final_data          = encoder_processer.processData(data);
+  ASSERT_TRUE(checkArrayEquality(*final_data, answer));
+}
+}  // namespace hyped::test
```
SnickeyX, line 65: space before final bracket

lib/navigation/preprocess_encoders.hpp
```diff
@@ -1,27 +1,96 @@
...
-  // initialised as {0, 0, 0, 0}, count of consecutive outliers
-  std::array<std::uint16_t, core::kNumEncoders> num_outliers_per_encoder_;
+  /**
+   * @brief cacluates the quartiles (q1,median,q3)
+   *
+   * @param array containing filtered values(i.e with no values from faulty sensors)
+   * @return a quartile object which contains q1, q3 and the median value
+   */
+  template<std::size_t N>
+  Quartile getQuartiles(std::array<std::uint32_t, N> &reliable_data) const
```
SnickeyX, line 81: since we know this will only run when N=4 or when N=3, why not explicitly instantiate this in the .hpp for quicker compilation. (look [here](https://stackoverflow.com/questions/13068993/when-would-you-use-template-explicit-instantiation#:~:text=func%3Cdouble%3E()%3B%20//%20OK%0A%20%20%20%20func%3Cint%3E()%3B%20//%20OK) - don't follow the .cpp definition part here but the explicit instantiation is pretty useful)

SnickeyX (changes requested):

lib/navigation/preprocess_encoders.cpp
```diff
@@ -2,55 +2,106 @@
...
-  TODOLater: implement
-  rough process:
-  - check how many times an individual encoder
-  has been an outlier in a row (outlier encoders)
-  -if num_outliers_per_encoder_[i] > n (tbd), mark encoder as unreliable
-  in reliable encoders.
-  -if number unreliable in reliable_encoders > 1, fail state
+  const auto encoder_statistics_optional = getStatistics(encoder_data);
+  if (!encoder_statistics_optional) {
+    logger_.log(core::LogLevel::kFatal, "Failed to obtain statistics for measurement");
```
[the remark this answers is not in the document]
SnickeyX: also statistics for what? (ik its annoying but it would be  confusing for those reading this message who aren't in Nav)

lib/navigation/preprocess_encoders.cpp
```diff
@@ -2,55 +2,106 @@
...
-  return {0, 0, 0, 0};
+  if (num_reliable_encoders_ < core::kNumEncoders - 1) {
+    logger_.log(core::LogLevel::kFatal,
+                "Unsuitable number of reliable encoders (%d of %d)",
+                num_reliable_encoders_,
+                core::kNumEncoders);
+    return std::nullopt;
+  }
+  if (num_reliable_encoders_ == core::kNumEncoders) {
+    auto reliable_data                    = encoder_data;
```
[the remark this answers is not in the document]
BenET01: We'll need to create the reliable_data variable anyway so that we can change the values, so I think it doesn't matter what way we do that

lib/navigation/preprocess_encoders.cpp
```diff
@@ -2,55 +2,106 @@
...
+    if (are_encoders_reliable_.at(i)) {
+      reliable_data.at(j) = encoder_data.at(i);
+      ++j;
+    }
+  }
+  const Quartile quartiles              = getQuartiles(reliable_data);
+  const core::Float interquartile_range = quartiles.third_quartile - quartiles.first_quartile;
+  return {{.median      = quartiles.median,
+           .upper_bound = quartiles.median + 1.2F * interquartile_range,
+           .lower_bound = quartiles.median - 1.2F * interquartile_range}};
```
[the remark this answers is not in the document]
BenET01: I'm not actually too sure, @maxguy2001 should be able to answer

lib/navigation/preprocess_encoders.cpp
```diff
@@ -2,55 +2,106 @@
...
-  -if num_outliers_per_encoder_[i] > n (tbd), mark encoder as unreliable
-  in reliable encoders.
-  -if number unreliable in reliable_encoders > 1, fail state
+  const auto encoder_statistics_optional = getStatistics(encoder_data);
+  if (!encoder_statistics_optional) {
+    logger_.log(core::LogLevel::kFatal, "Failed to obtain statistics for measurement");
+    return std::nullopt;
+  }
+  const auto encoder_statistics = *encoder_statistics_optional;
+  auto sanitised_data           = encoder_data;
```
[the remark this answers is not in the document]
BenET01: We need to create a copy of the data so we can change outliers later

lib/navigation/preprocess_encoders.hpp
BenET01: We were looking to make this as general as possible for future teams, so this function could run with any number of sensors

test/navigation/preprocess_encoders.cpp
BenET01: formatting

lib/navigation/preprocess_encoders.cpp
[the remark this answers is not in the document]
SnickeyX: I see, think this should be fine

lib/navigation/preprocess_encoders.cpp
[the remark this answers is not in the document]
maxguy2001: Since the number of datapoints decreases if we lose a reliable sensor, we need to decrease the threshold in response.

maxguy2001 (approved):

PR #39, opened 2022-11-07
Nav - IMU Preprocessing

BenET01:
Implementing pre-processing for IMU data


kshxtij:
For the sake of consistency, we should rename anything named IMU to Accelerometer to match #31

lib/navigation/preprocess_imu.cpp
```diff
@@ -44,16 +46,53 @@ namespace hyped::navigation
...
-    if (num_reliable_imus == 4)
-    {
-      std::sort(imu_data.start(), imu_data.end());
-      const core::Float q1 = (imu_data.at(0) + imu_data.at(1)) / 2.0;
-      const core::Float median = (imu_data.at(1) + imu_data.at(2)) / 2.0;
-      const core::Float q3 = (imu_data.at(2) + imu_data.at(3)) / 2.0;
+    const uint8_t num_reliable_accelerometers = std::accumulate(are_imus_reliable_.begin(), are_imus_reliable_.end(), 0);
+    if (num_reliable_accelerometers == 4)
+    { 
+      core::ImuData clean_accelerometer_data_copy = std::copy(clean_accelerometer_data);
```
maxguy2001, line 54: Look at the arguments for `std::copy()`. C++ needs to know where to look and where to put things!

lib/navigation/preprocess_imu.cpp
```diff
@@ -44,16 +46,53 @@ namespace hyped::navigation
...
+      std::sort(clean_accelerometer_data_copy.begin(), clean_accelerometer_data_copy.end());
+      const core::Float q1 = (clean_accelerometer_data_copy.at(0) + clean_accelerometer_data_copy.at(1)) / 2.0;
+      const core::Float median = (clean_accelerometer_data_copy.at(1) + clean_accelerometer_data_copy.at(2)) / 2.0;
+      const core::Float q3 = (clean_accelerometer_data_copy.at(2) + clean_accelerometer_data_copy.at(3)) / 2.0;
+      const core::Float iqr = q3 - q1;
+
```
maxguy2001, line 60: probably easier to define upper and lower bounds here too and just compare data to those?

lib/navigation/preprocess_imu.cpp
```diff
@@ -44,16 +46,53 @@ namespace hyped::navigation
...
+      for (int i = 0; i < core::kNumImus; ++i) {
```
maxguy2001, line 61: `size_t i = 0`

lib/navigation/preprocess_imu.cpp
```diff
@@ -44,16 +46,53 @@ namespace hyped::navigation
...
+        if (clean_accelerometer_data.at(i) > q3 + 1.5*iqr || clean_accelerometer_data.at(i) < q1 - 1.5*iqr) {
+            clean_accelerometer_data.at(i) = median;
+            num_outliers_per_imu_.at(i) += 1;
+        } else {
+          num_outliers_per_imu_.at(i) = 0;
+        }
+      }
+    } else if (num_reliable_accelerometers == 3) {
+      core::ImuData clean_accelerometer_data_copy = std::copy(clean_accelerometer_data);
```
maxguy2001, line 70: again, look at how `std::copy()` works

maxguy2001: Also surely you would want to define an array fo length `core::kNumImus -1` here?

lib/navigation/preprocess_imu.cpp
```diff
@@ -44,16 +46,53 @@ namespace hyped::navigation
...
+      for (int i = 0; i < core::kNumImus; ++i) {
+        if (are_imus_reliable_.at(i) == false) {
+          clean_accelerometer_data_copy.at(i) = -1;
```
maxguy2001, line 73: I'm not  fan of this method. Think of a way to do it without needing to do this.

lib/navigation/preprocess_imu.cpp
```diff
@@ -44,16 +46,53 @@ namespace hyped::navigation
...
+      const core::Float q1 = (clean_accelerometer_data_copy.at(1) + clean_accelerometer_data_copy.at(2)) / 2.0;
+      const core::Float median = clean_accelerator_data_copy.at(2);
+      const core::Float q3 = (clean_accelerometer_data_copy.at(2) + clean_accelerometer_data_copy.at(3)) / 2.0;
+      const core::Float iqr = q3 - q1;
+      
+      for (int i = 0; i < core::kNumImus; ++i) {
+        if (are_imus_reliable_.at(i) == false) {
+          clean_accelerometer_data.at(i) = median;
+        } else {
+          if (clean_accelerometer_data.at(i) > q3 + 1.5*iqr || clean_accelerometer_data.at(i) < q1 - 1.5*iqr) {
```
maxguy2001, line 86: This won't be 1.5, it will probably be something like 1.2 but play around with numbers and decide what you think this value should be?

maxguy2001 (changes requested):

lib/navigation/preprocess_imu.cpp
kshxtij: `std::size_t`

lib/navigation/preprocess_imu.cpp
BenET01: How would that work with the return type, would the unreliable imu just be converted to median as well afterwards or would we return a different size array?

maxguy2001: We would replace the unreliable value with median. Return type has to remain constant (unless we want to start making return type optional but I don't think that's necessary here).

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,120 @@
...
+    - if any datapoint outwith this, set
+    outlier_imus_[i] += 1
+    -set outlier points as median
+
+    -also has to be able to handle 1 unreliable sensor
+    -also figure out return type/ what we update and update
+    documentation as appropriate
+    */
+    core::ImuData clean_accelerometer_data;
+    core::Float magnitude;
```
maxguy2001, line 41: Also, now that I think about it, we're going to have some issues with using this definition of magnitude since we will have to keep everything constant during deceleration. Maybe not a job for now but we should maybe include some hardware calibration to decide what we roughly want to be as the positive direction and what will be the negative.

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,120 @@
...
+      {
+        magnitude += std::pow(imu_data.at(i).at(j), 2);
+      }
+      clean_accelerometer_data.at(i) = std::sqrt(magnitude);
+    }
+    const uint8_t num_reliable_accelerometers = std::accumulate(are_imus_reliable_.begin(), are_imus_reliable_.end(), 0);
+    if (num_reliable_accelerometers == 4)
+    { 
+      core::ImuData clean_accelerometer_data_copy = std::copy(clean_accelerometer_data.begin(), clean_accelerometer_data.end());
+      std::sort(clean_accelerometer_data_copy.begin(), clean_accelerometer_data_copy.end());
```
maxguy2001, line 55: maybe make separate function to do all this? It's starting to get a bit messy?

maxguy2001: i.e. something along the lines of `getOutlierThresholds(const core::ImuData &imu_data)`?

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,120 @@
...
+        }
+      }
+    } else if (num_reliable_accelerometers == 3) {
+      core::ImuData clean_accelerometer_data_copy = std::copy(clean_accelerometer_data.begin(), clean_accelerometer_data.end());
+      for (std::size_t i = 0; i < core::kNumImus; ++i) {
+        if (are_imus_reliable_.at(i) == false) {
+          clean_accelerometer_data_copy.at(i) = -1;
+        }
+      }
+      std::sort(clean_accelerometer_data_copy.begin(), clean_accelerometer_data_copy.end());
```
maxguy2001, line 78: This looks similar to method done above. See if you can merge these cases into a more succinct function?

maxguy2001 (changes requested):

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,120 @@
 #include "preprocess_imu.hpp"
 
-namespace hyped::navigation {
+#include <cmath>
```
maxguy2001, line 3: You're missing headers. You'll need <algorithm>, <iterator> and if it's not already in the header file, <array> too.

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,5 +1,11 @@
 #include "preprocess_imu.hpp"
 
+#include <cmath>
+
+#include <algorithm>
+#include <iterator>
```
maxguy2001, line 6: Don't need iterator anymore

lib/navigation/preprocess_imu.cpp
```diff
@@ -9,48 +15,158 @@ ImuPreprocessor::ImuPreprocessor()
...
+  // TODOLater : change magnitude for directions (hardware configuration?)
+  core::Float magnitude;
+  for (std::size_t i = 0; i < core::kNumImus; ++i) {
+    magnitude = 0;
+    for (std::size_t j = 0; j < 3; ++j) {
+      magnitude += std::pow(imu_data.at(i).at(j), 2);
+    }
+    clean_accelerometer_data.at(i) = std::sqrt(magnitude);
+  }
+  Quartiles quartiles     = getOutlierThresholds(clean_accelerometer_data);
```
maxguy2001, line 36: @aryansharma19 Look at the quartiles struct. This is a neater implementation and will work for encoders data too.

lib/navigation/preprocess_imu.cpp
```diff
@@ -9,48 +15,158 @@ ImuPreprocessor::ImuPreprocessor()
...
+      = (clean_accelerometer_data_copy.at(1) + clean_accelerometer_data_copy.at(2)) / 2.0;
+    const core::Float q3
+      = (clean_accelerometer_data_copy.at(2) + clean_accelerometer_data_copy.at(3)) / 2.0;
+    Quartiles quartiles;
+    quartiles.q1     = q1;
+    quartiles.median = median;
+    quartiles.q3     = q3;
+    return quartiles;
+  }
+  // if an unreliable sensor exists, copy true values into another array
```
maxguy2001, line 79: unclear

lib/navigation/preprocess_imu.cpp
```diff
@@ -9,48 +15,158 @@ ImuPreprocessor::ImuPreprocessor()
...
+  const uint8_t num_reliable_accelerometers
+    = std::accumulate(are_imus_reliable_.begin(), are_imus_reliable_.end(), 0);
+
+  if (num_reliable_accelerometers == core::kNumImus) {
+    core::ImuData clean_accelerometer_data_copy;
+    std::copy(clean_accelerometer_data.begin(),
+              clean_accelerometer_data.end(),
+              clean_accelerometer_data_copy.begin());
+    std::sort(clean_accelerometer_data_copy.begin(), clean_accelerometer_data_copy.end());
+    const core::Float q1
```
maxguy2001, line 67: directly assign instead of making objects and then assigning the object to the struct member?

lib/navigation/preprocess_imu.cpp
```diff
@@ -9,48 +15,158 @@ ImuPreprocessor::ImuPreprocessor()
...
+    Quartiles quartiles;
+    quartiles.q1     = q1;
+    quartiles.median = median;
+    quartiles.q3     = q3;
+    return quartiles;
+  }
+  // if an unreliable sensor exists, copy true values into another array
+  std::array<core::Float, core::kNumImus - 1> clean_accelerometer_data_copy;
+  std::uint8_t index;
+  for (size_t i = 0; i < core::kNumImus - 1; ++i) {
```
maxguy2001, line 82: std::size_t

lib/navigation/preprocess_imu.cpp
```diff
@@ -9,48 +15,158 @@ ImuPreprocessor::ImuPreprocessor()
...
+    const core::Float q3
+      = (clean_accelerometer_data_copy.at(2) + clean_accelerometer_data_copy.at(3)) / 2.0;
+    Quartiles quartiles;
+    quartiles.q1     = q1;
+    quartiles.median = median;
+    quartiles.q3     = q3;
+    return quartiles;
+  }
+  // if an unreliable sensor exists, copy true values into another array
+  std::array<core::Float, core::kNumImus - 1> clean_accelerometer_data_copy;
```
maxguy2001, line 80: not loving this name. Could/should be clearer?

lib/navigation/preprocess_imu.cpp
```diff
@@ -9,48 +15,158 @@ ImuPreprocessor::ImuPreprocessor()
...
+  std::uint8_t index;
+  for (size_t i = 0; i < core::kNumImus - 1; ++i) {
+    if (!are_imus_reliable_.at(i)) { index = i; }
+  }
+
+  for (size_t i = 0; i < core::kNumImus - 1; ++i) {
```
maxguy2001, line 86: Tempted to go ahead and use vectors here. Will discuss Monday.

lib/navigation/preprocess_imu.cpp
```diff
@@ -9,48 +15,158 @@ ImuPreprocessor::ImuPreprocessor()
...
+  std::uint8_t num_unreliable = 0;
+  num_unreliable
+    = core::kNumImus - std::accumulate(are_imus_reliable_.begin(), are_imus_reliable_.end(), 0);
+  // changes reliable sensor to false if max consecutive outliers are reached
+  for (size_t i = 0; i < core::kNumImus; ++i) {
+    if (num_outliers_per_imu_.at(i) >= kNumAllowedImuFailures_) {
+      are_imus_reliable_.at(i) = false;
+    }
+  }
+  if (num_unreliable > 1) {
```
maxguy2001, line 167: might change this function to have a return type but can discuss Monday

maxguy2001 (changes requested):

lib/navigation/preprocess_imu.cpp
```diff
@@ -9,48 +14,126 @@ ImuPreprocessor::ImuPreprocessor()
...
+  // TODOLater : change magnitude for directions (hardware configuration?)
+  core::Float magnitude;
+  for (std::size_t i = 0; i < core::kNumImus; ++i) {
+    magnitude = 0;
+    for (std::size_t j = 0; j < 3; ++j) {
+      magnitude += std::pow(imu_data.at(i).at(j), 2);
+    }
+    clean_accelerometer_data.at(i) = std::sqrt(magnitude);
+  }
+  Quartiles quartiles     = getOutlierThresholds(clean_accelerometer_data);
```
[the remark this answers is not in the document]
maxguy2001: Also rename the function to something closer to what it actually does (i.e. `getQuartiles` or similar)

lib/navigation/preprocess_imu.cpp
```diff
@@ -9,48 +14,126 @@ ImuPreprocessor::ImuPreprocessor()
...
+  } else {
+    std::size_t index;
+    for (size_t i = 0; i < core::kNumImus - 1; ++i) {
+      if (!are_imus_reliable_.at(i)) { index = i; }
+    }
+    for (size_t i = 0; i < core::kNumImus - 1; ++i) {
+      if (i >= index) {
+        clean_accelerometer_data_copy_one_unreliable.at(i) = clean_accelerometer_data.at(i + 1);
+      } else {
+        clean_accelerometer_data_copy_one_unreliable.at(i) = clean_accelerometer_data.at(i);
```
maxguy2001, line 76: `clean_accelerometer_data_copy_one_unreliable`?!?!?!?!?! I want to rip my eyes out!

maxguy2001: If you start using names like that, you're making too many variables!

lib/navigation/preprocess_imu.cpp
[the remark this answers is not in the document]
maxguy2001: Ok have looked again at what you're doing. I think you need to change your `getOutlierThresholds`...

maxguy2001 (changes requested):
Few changes to get started with. Mostly structural changes for now, will get to style guide etc. later.

lib/navigation/preprocess_imu.cpp
```diff
@@ -9,48 +14,126 @@ ImuPreprocessor::ImuPreprocessor()
 
 core::ImuData ImuPreprocessor::processData(const core::RawImuData raw_imu_data)
```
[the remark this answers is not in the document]
BenET01: Is that a null pointer? That's all I can find online

lib/navigation/preprocess_imu.cpp
BenET01: yeah a very bad place holder name. I'm not too sure how to cut down on the amount of variables, as I need to initialise both arrays for different situations, as I can't initialise in loops without massively repetitive code

lib/navigation/preprocess_imu.cpp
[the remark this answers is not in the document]
maxguy2001: Yeah sorry nullpointer. I'm sure there's examples already in the hyped codebase, maybe in navigator class if you're stuck.

lib/navigation/preprocess_imu.cpp
maxguy2001: Templates and helper functions where possible. Otherwise just more careful naming.

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,128 @@
...
 ImuPreprocessor::ImuPreprocessor()
 {
   // TODOLater: implement
 }
 
-core::ImuData ImuPreprocessor::processData(const core::RawImuData raw_imu_data)
+std::optional<core::ImuData> ImuPreprocessor::processData(const core::RawImuData raw_imu_data)
+{
+  core::ImuData imu_data;
+  core::ImuData accelerometer_data;
```
maxguy2001, line 19: This should be defined at the same time as it is used. It is not allocated in a loop so no point defining up here and assigning 10 lines later.

maxguy2001: Also when assigned it can then be `const`.

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,128 @@
...
+    }
+  }
+  return accelerometer_data;
+}
+
+template<std::size_t N>
+Quartiles ImuPreprocessor::getQuartiles(const std::array<core::Float, N> &clean_accelerometer_data)
+{
+  std::array<core::Float, N> accelerometer_data_copy;
+  for (std::size_t i; i < num_reliable_accelerometers_; ++i) {
```
maxguy2001, line 84: `std::copy`?

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,128 @@
...
+  quartiles.q1     = getSpecificQuartile(accelerometer_data_copy, 0.25);
+  quartiles.median = getSpecificQuartile(accelerometer_data_copy, 0.5);
+  quartiles.q3     = getSpecificQuartile(accelerometer_data_copy, 0.75);
+  return quartiles;
 }
 
-core::ImuData ImuPreprocessor::detectOutliers(const core::RawImuData imu_data)
+template<std::size_t N>
+core::Float ImuPreprocessor::getSpecificQuartile(
+  const std::array<core::Float, N> &clean_accelerometer_data_copy, core::Float quartile)
```
maxguy2001, line 98: Naming of the argument of this function. No point calling it a copy here.

lib/navigation/preprocess_imu.cpp
```diff
@@ -9,48 +14,126 @@ ImuPreprocessor::ImuPreprocessor()
...
-  has been an outlier in a row (outlier imus)
-  -if outlier_imus_[i] > n (tbd), mark imu as unreliable
-  in reliable imus.
-  -if number unreliable in reliable_imus > 1, fail state
-
-  -also figure out return type/ what we update and update
-  documentation as appropriate
-  */
+  std::uint8_t num_unreliable;
+  num_unreliable
```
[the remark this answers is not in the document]
maxguy2001: You made a class member for this but didn't use it, you still calculate every time?

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,123 @@
...
+#include <cmath>
+
+#include <algorithm>
+#include <numeric>
+#include <optional>
+
 namespace hyped::navigation {
 
 ImuPreprocessor::ImuPreprocessor()
 {
```
maxguy2001, line 12: implement logger

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,123 @@
...
+      if (i >= index) {
+        faulty_data.at(i) = imu_data.at(i + 1);
+      } else {
+        faulty_data.at(i) = imu_data.at(i);
+      }
+    }
+    quartiles = getQuartiles(faulty_data);
+  }
+
+  const core::Float iqr         = quartiles.q3 - quartiles.q1;
```
maxguy2001, line 59: move these assignments into if-else statements, changing threshold to 1.2 if an imu is faulty

maxguy2001 (changes requested):

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,133 @@
...
+      }
+    }
+    quartiles = getQuartiles(faulty_data);
+  }
+
+  const core::Float iqr = quartiles.q3 - quartiles.q1;
+  core::Float lower_bound;
+  core::Float upper_bound;
+  if (num_reliable_accelerometers_ == core::kNumImus) {
+    lower_bound = quartiles.median - 1.5 * iqr;
```
maxguy2001, line 66: TODOLater for checking these values

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,133 @@
...
+  const core::Float quartile_value       = (clean_accelerometer_data.at(index_quartile_high)
+                                      + clean_accelerometer_data.at(index_quartile_low))
+                                     / 2.0;
+  return quartile_value;
+}
+
+SensorChecks ImuPreprocessor::checkReliable()
+{  // changes reliable sensor to false if max consecutive outliers are reached
+  for (std::size_t i = 0; i < core::kNumImus; ++i) {
+    if (num_outliers_per_imu_.at(i) >= kNumAllowedImuFailures_) {
```
maxguy2001, line 121: If one IMU fails, since you don't reset anything and keep checking against this condition would you not keep reducing the number of reliable imus. Maybe put in a check here for if you've already classified the imu as faulty?

maxguy2001 (changes requested):
Can you also add docstrings into the hpp file so that the functions themselves have some level of documentation? Going to convert this from draft tho, it starting to look ready..

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,137 @@
...
+
+#include <algorithm>
+#include <numeric>
+#include <optional>
+
 namespace hyped::navigation {
 
-ImuPreprocessor::ImuPreprocessor()
+ImuPreprocessor::ImuPreprocessor(core::ILogger &logger)
+    : log_(logger),
```
kshxtij, line 12: We're using logger_ everywhere

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,137 @@
...
+  for (std::size_t i = 0; i < core::kNumImus; ++i) {
+    magnitude = 0;
+    for (std::size_t j = 0; j < 3; ++j) {
+      magnitude += std::pow(raw_imu_data.at(i).at(j), 2);
+    }
+    imu_data.at(i) = std::sqrt(magnitude);
+  }
+
+  const core::ImuData accelerometer_data = detectOutliers(imu_data);
+  checkReliable();
```
kshxtij, line 33: checkReliable returns `SensorChecks`, surely we want to be checking that?

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,137 @@
...
+      ++num_outliers_per_imu_.at(i);
+    } else {
+      num_outliers_per_imu_.at(i) = 0;
+    }
+  }
+  return accelerometer_data;
 }
 
-core::ImuData ImuPreprocessor::detectOutliers(const core::RawImuData imu_data)
+template<std::size_t N>
```
kshxtij, line 91: My understanding is that the templated functions should only be in header files. @mifrandir correct me if I'm wrong, similar to how the mux was done.

kshxtij (changes requested):

lib/navigation/preprocess_imu.cpp
mifrandir: You are right!

mifrandir: Why do we want this template in the first place?

Also, this implementation is far from being asymptotically optimal so I'd like to see a comment noting that and what runtime we currently have. I'm pretty sure this can be done in linear time.

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,137 @@
...
+  }
+
+  Quartiles quartiles;
+  if (num_reliable_accelerometers_ == core::kNumImus) {
+    quartiles = getQuartiles(imu_data);
+  } else {
+    std::array<core::Float, core::kNumImus - 1> faulty_data;
+    std::size_t index;
+    for (std::size_t i = 0; i < core::kNumImus - 1; ++i) {
+      if (!are_imus_reliable_.at(i)) { index = i; }
```
mifrandir, line 51: Here we are making the strong assumption that there is only one faulty IMU. This will be a hassle in case it ever changes and ... we don't really need to?

Why not 
```cpp
std::size_t j = 0;
for (std::size_t i = 0; i < core::kNumImus; ++i) {
  if (are_imus_reliable_.at(i)) {
    filtered_data.at(j) = imu_data.at(i);
    ++j;
  }
 }
```

Of course, you now have to do some thinking about how to make things work for arbitrary sets of data, dependent on the values provided at runtime. An obvious choice is `std::vector` but we need to make sure not to allocate. What we really want is a collection that behaves like an `std::vector` except that we guarantee that we are not going to add more than `kNumImus` elements, ever. This allows for an underlying structure `std::array<T, kNumImus>`. I guess the simplest way of doing this is to simply use `std::array` everywhere and pass an additional `std::size_t length` like in the good old C days? @kshxtij

mifrandir: You could then add a `// TODOLater` to make it into a nice data structure... which I might eventually do.

mifrandir (changes requested):

lib/navigation/preprocess_imu.cpp
BenET01: In this implementation, N is either kNumImus or kNumImus-1. Currently this is either 3 or 4 so should run fairly quickly? The template allows the number of imus to be variable so currently the options seem to be either repetetive code or  template function. Is there a nicer way of doing this? Given that N is either 3 or 4, does this not still run in relatively constant time?

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,137 @@
...
+    }
+    imu_data.at(i) = std::sqrt(magnitude);
+  }
+
+  const core::ImuData accelerometer_data = detectOutliers(imu_data);
+  checkReliable();
+  return accelerometer_data;
+}
+
+core::ImuData ImuPreprocessor::detectOutliers(const core::ImuData imu_data)
```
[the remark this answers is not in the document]
BenET01: Is 20 lines enough to create another function? The rules state that 60 lines is the max?

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,137 @@
...
+  for (std::size_t i; i < core::kNumImus; ++i) {
+    accelerometer_data.at(i) = imu_data.at(i);
+  }
+
+  Quartiles quartiles;
+  if (num_reliable_accelerometers_ == core::kNumImus) {
+    quartiles = getQuartiles(imu_data);
+  } else {
+    std::array<core::Float, core::kNumImus - 1> faulty_data;
+    std::size_t index;
```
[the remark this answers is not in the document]
BenET01: I agree about index but if they go out of sync, then we would catch it on the next 'iteration' which we would expect to happen in relatively quick succession?

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,137 @@
...
+  core::ImuData accelerometer_data;
+  for (std::size_t i; i < core::kNumImus; ++i) {
+    accelerometer_data.at(i) = imu_data.at(i);
+  }
+
+  Quartiles quartiles;
+  if (num_reliable_accelerometers_ == core::kNumImus) {
+    quartiles = getQuartiles(imu_data);
+  } else {
+    std::array<core::Float, core::kNumImus - 1> faulty_data;
```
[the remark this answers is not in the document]
maxguy2001: Maybe faulty isn't 100% the correct word. It's more that there is a fault in the IMU's  and one has been classified as faulty.

lib/navigation/preprocess_imu.cpp
```diff
@@ -30,7 +30,7 @@ std::optional<core::ImuData> ImuPreprocessor::processData(const core::RawImuData
   }
 
   const core::ImuData accelerometer_data = detectOutliers(imu_data);
-  checkReliable();
+  SensorChecks sensorcheck               = checkReliable();
```
mifrandir, line 33: you're still not using this?

lib/navigation/preprocess_imu.hpp
```diff
@@ -19,11 +22,32 @@ class ImuPreprocessor {
...
 
   template<std::size_t N>
-  Quartiles getQuartiles(const std::array<core::Float, N> &imu_data);
+  Quartiles getQuartiles(const std::array<core::Float, N> &imu_data)
+  {
+    std::array<core::Float, N> accelerometer_data_copy;
+    std::copy(imu_data.begin(), imu_data.end(), accelerometer_data_copy.begin());
+
+    std::sort(accelerometer_data_copy.begin(), accelerometer_data_copy.end());
+
```
mifrandir, line 44: whitespace, here and elsewhere

lib/navigation/preprocess_imu.hpp
```diff
@@ -19,11 +22,32 @@ class ImuPreprocessor {
...
+                                       / 2.0;
+    return quartile_value;
+  }
 
   template<std::size_t N>
-  Quartiles getQuartiles(const std::array<core::Float, N> &imu_data);
+  Quartiles getQuartiles(const std::array<core::Float, N> &imu_data)
+  {
+    std::array<core::Float, N> accelerometer_data_copy;
+    std::copy(imu_data.begin(), imu_data.end(), accelerometer_data_copy.begin());
```
mifrandir, line 41: `imu_data` vs `accelerometer_data_copy` doesn't make sense to me

lib/navigation/preprocess_imu.hpp
```diff
@@ -19,11 +22,32 @@ class ImuPreprocessor {
   SensorChecks checkReliable();
 
   template<std::size_t N>
-  core::Float getSpecificQuartile(const std::array<core::Float, N> &clean_accelerometer_data_copy,
-                                  core::Float quartile);
+  core::Float getSpecificQuartile(const std::array<core::Float, N> &clean_accelerometer_data,
```
mifrandir, line 25: I don't like the naming of this function. Firstly, it doesn't return a `Quartile`. Secondly, it doesn't really give a quartile in the first place? If I call this function with `quartile == 0.14` this code still works (and may even be useful in some cases)! Find a better name that describes what this *actually* does.

lib/navigation/preprocess_imu.hpp
```diff
@@ -19,11 +22,32 @@ class ImuPreprocessor {
...
+    std::array<core::Float, N> accelerometer_data_copy;
+    std::copy(imu_data.begin(), imu_data.end(), accelerometer_data_copy.begin());
+
+    std::sort(accelerometer_data_copy.begin(), accelerometer_data_copy.end());
+
+    Quartiles quartiles;
+    quartiles.q1     = getSpecificQuartile(accelerometer_data_copy, 0.25);
+    quartiles.median = getSpecificQuartile(accelerometer_data_copy, 0.5);
+    quartiles.q3     = getSpecificQuartile(accelerometer_data_copy, 0.75);
+    return quartiles;
```
mifrandir, line 49: i'd prefer

```cpp
return {
  .q1 = getSpecificQuartile(...),
  .median = ...,
  ...
};
```

why is this better?
1. we get rid of the mutable variable `quartiles`.
2. this will complain if some fields in `Quartiles` are not properly initialised, thereby catching a source of undefined behaviour.

mifrandir (changes requested):
i'd like to see some minimal tests to see that this compiles if the templates are instantiated. then you could also test that the thing actually works.

lib/navigation/preprocess_imu.hpp
```diff
@@ -35,18 +35,17 @@ class ImuPreprocessor {
...
     std::sort(accelerometer_data_copy.begin(), accelerometer_data_copy.end());
-
     Quartiles quartiles;
-    quartiles.q1     = getSpecificQuartile(accelerometer_data_copy, 0.25);
-    quartiles.median = getSpecificQuartile(accelerometer_data_copy, 0.5);
-    quartiles.q3     = getSpecificQuartile(accelerometer_data_copy, 0.75);
-    return quartiles;
+    return {quartiles.q1     = getSpecificQuantile(accelerometer_data_copy, 0.25),
+            quartiles.median = getSpecificQuantile(accelerometer_data_copy, 0.5),
+            quartiles.q3     = getSpecificQuantile(accelerometer_data_copy, 0.75)};
```
mifrandir, line 48: it should be possible to drop the `Quartiles quartiles` bit and then leave out `quartiles` in the return statement.

mifrandir (changes requested):

lib/navigation/preprocess_imu.hpp
mifrandir: https://pdimov.github.io/blog/2020/09/07/named-parameters-in-c20/

lib/navigation/preprocess_imu.hpp
mifrandir: https://pdimov.github.io/blog/2020/09/07/named-parameters-in-c20/

lib/navigation/preprocess_imu.cpp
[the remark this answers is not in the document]
BenET01: @mifrandir, thoughts?

BenET01:
@mifrandir, I've created the test file for this but am not too sure on how to get properly started on making tests. Any advice on how to start that off so I'm able to get going with that?

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,18 @@
+#include <gtest/gtest.h>
+#include <cmath>
```
maxguy2001, line 2: Not used...

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,106 @@
 #include "preprocess_imu.hpp"
 
+#include <cmath>
+
+#include <algorithm>
+#include <numeric>
```
maxguy2001, line 6: not used

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,106 @@
 #include "preprocess_imu.hpp"
 
+#include <cmath>
```
maxguy2001, line 3: already included in header

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,106 @@
...
+
+#include <algorithm>
```
maxguy2001, line 5: already included in header

lib/navigation/preprocess_imu.cpp
```diff
@@ -1,56 +1,106 @@
...
+#include <numeric>
+#include <optional>
```
maxguy2001, line 7: This should be included in the header, not the cpp file

maxguy2001 (changes requested):

lib/navigation/preprocess_imu.cpp
```diff
@@ -2,55 +2,99 @@
...
+    }
+    imu_data.at(i) = std::sqrt(magnitude);
+  }
+  const core::ImuData accelerometer_data = detectOutliers(imu_data);
+  SensorChecks sensorcheck               = checkReliable();
 
-  */
-  return {0, 0, 0, 0};
+  if (sensorcheck == SensorChecks::kUnacceptable) {
+    return std::nullopt;
```
Fayz-S, line 29: Are we not wanting the logger to log a failure here?

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+TEST(Imu, equal_data)
+{
+  utils::ManualTime manual_time;
+  core::Logger logger("test", core::LogLevel::kFatal, manual_time);
+  navigation::ImuPreprocessor imu_processer(logger);
+  core::RawImuData data                          = {{1, 1, 1}};
+  core::ImuData answer                           = {static_cast<float> (std::sqrt(3.0))};
+  std::optional<std::array<float, 4>> final_data = imu_processer.processData(data);
+  ASSERT_EQ(*final_data, answer);
```
Fayz-S, line 27: As far as I can understand you are comparing arrays here: you cannot use ASSERT_EQ for whole arrays, needs to be done element-wise. Might be good to check the lengths of both arrays are equal first

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,56 @@
...
+  core::Logger logger("test", core::LogLevel::kFatal, manual_time);
+  navigation::ImuPreprocessor imu_processer(logger);
+  core::RawImuData data = {{1, 1, 1}};
+  data.at(0)            = {3, 5, 6};
+  for (size_t i; i < 30; ++i) {
+    imu_processer.processData(data);
+  }
+  core::ImuData answer                           = {static_cast<float> (std::sqrt(3.0))};
+  std::optional<std::array<float, 4>> final_data = imu_processer.processData(data);
+  ASSERT_EQ(*final_data, answer);
```
Fayz-S, line 54: ASSERT_EQ can't be used here- see previous comment

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,56 @@
...
+TEST(Imu, not_equal_data)
+{
+  utils::ManualTime manual_time;
+  core::Logger logger("test", core::LogLevel::kFatal, manual_time);
+  navigation::ImuPreprocessor imu_processer(logger);
+  core::RawImuData data                          = {{1, 1, 1}};
+  data.at(0)                                     = {3, 5, 6};
+  core::ImuData answer                           = {static_cast<float> (std::sqrt(3.0))};
+  std::optional<std::array<float, 4>> final_data = imu_processer.processData(data);
+  ASSERT_EQ(*final_data, answer);
```
Fayz-S, line 39: ASSERT_EQ can't be used here- see previous comment

lib/navigation/preprocess_imu.cpp
BenET01: The logger logs the failure before sensor checks changes to kUnacceptable in the checkReliable function

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,56 @@
...
+}
+
+TEST(Imu, not_equal_data)
+{
+  utils::ManualTime manual_time;
+  core::Logger logger("test", core::LogLevel::kFatal, manual_time);
+  navigation::ImuPreprocessor imu_processer(logger);
+  core::RawImuData data                          = {{1, 1, 1}};
+  data.at(0)                                     = {3, 5, 6};
```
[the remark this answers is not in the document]
BenET01: I tried to initialise as {{3, 5, 6}, {1, 1, 1}, {1, 1, 1}, {1, 1, 1}} however that caused an error, I thought that was the way to write arrays but is that wrong?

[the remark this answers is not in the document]
mifrandir: almost, C++ is weird; you need an extra pair of parantheses on the outside 

```cpp
#include <array>
#include <iostream>

int main() {
  std::array<std::array<int, 3>, 3> my_array = {{{ 1, 1, 1 }, { 2, 2, 2 }, { 3, 3, 3 }}};
  for (const auto &col : my_array) {
    for (const auto x : col) {
      std::cout << x << " ";
    }
    std::cout << std::endl;
  }
}
```

Output:

```
$ g++ test.cpp && ./a.out
1 1 1 
2 2 2 
3 3 3 
```

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <navigation/preprocess_imu.hpp>
+#include <utils/manual_time.hpp>
+
+namespace hyped::test {
+
+bool checkArrayEquality(core::ImuData &imu_data_a, core::ImuData &imu_data_b)
+{
+  if (imu_data_a.size() == imu_data_b.size()) {
+    for (std::size_t i; i < imu_data_a.size(); ++i) {
+      if (imu_data_a.at(i) == imu_data_b.at(i)) {
```
maxguy2001, line 16: Could just do 
```
if(imu_data_a.at(i) != imu_data_b.at(i)){
return false;
}
```
instead of the additional else statement?

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,70 @@
...
+  core::Logger logger("test", core::LogLevel::kFatal, manual_time);
+  navigation::ImuPreprocessor imu_processer(logger);
+}
+
+TEST(Imu, equal_data)
+{
+  utils::ManualTime manual_time;
+  core::Logger logger("test", core::LogLevel::kFatal, manual_time);
+  navigation::ImuPreprocessor imu_processer(logger);
+  core::RawImuData data                          = {{1, 1, 1}};
```
maxguy2001, line 40: Are these all initialised with correct dimensions?

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,70 @@
...
+  core::ImuData answer                           = {static_cast<float>(std::sqrt(3.0))};
```
maxguy2001, line 41: `ImuData` is of type `core::Float`, not float.

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,70 @@
...
+  std::optional<std::array<float, 4>> final_data = imu_processer.processData(data);
```
maxguy2001, line 42: same again, `core::Float`. Everywhere else too.

test/navigation/imu_preprocessing.cpp
maxguy2001: Also given that these are floats (of type `core::Float`), shouldn't we be using something along the lines of:
```
if(std::abs(imu_data_a.at(i) - imu_data_b.at(i)) < epsilon){
...
```
??

This feels like it would stop any potential failure of tests due to things like rounding errors at a very reasonable level of precision?

test/navigation/imu_preprocessing.cpp
maxguy2001: also if you have a return type like that, pretty sure you would set the type to `auto` (i.e. `auto final_data = imu...`)? @kshxtij?

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,70 @@
...
+  ASSERT_TRUE(checkArrayEquality(*final_data, answer));
+}
+
+TEST(Imu, one_unreliable_sensor)
+{
+  utils::ManualTime manual_time;
+  core::Logger logger("test", core::LogLevel::kFatal, manual_time);
+  navigation::ImuPreprocessor imu_processer(logger);
+  core::RawImuData data = {{{3, 5, 6}, {1, 1, 1}, {1, 1, 1}, {1, 1, 1}}};
+  for (std::size_t i; i < 30; ++i) {
```
maxguy2001, line 63: set it closer to the maximum allowed consecutive failures pls

maxguy2001 (changes requested):
Overall excellent start! I think implement these changes for now and we'll see how it's looking but I'm liking it so far!

test/navigation/imu_preprocessing.cpp
BenET01: Yeah it initialises an array with all equal elements to the size of the array

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,68 @@
...
+TEST(Imu, one_unreliable_sensor)
+{
+  utils::ManualTime manual_time;
+  core::Logger logger("test", core::LogLevel::kFatal, manual_time);
+  navigation::ImuPreprocessor imu_processer(logger);
+  core::RawImuData data = {{{3, 5, 6}, {1, 1, 1}, {1, 1, 1}, {1, 1, 1}}};
+  for (std::size_t i; i < 22; ++i) {
+    imu_processer.processData(data);
+  }
+  core::ImuData answer                    = {static_cast<float>(std::sqrt(3.0))};
```
maxguy2001, line 64: `core::Float`

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <gtest/gtest.h>
+
+#include <core/logger.hpp>
+#include <core/types.hpp>
+#include <navigation/preprocess_imu.hpp>
+#include <utils/manual_time.hpp>
+
+namespace hyped::test {
+
+core::Float epsilon = 0.001;
```
maxguy2001, line 12: epsilon is too large here. This would allow a 1-step rounding error of that size which would mess up our estimates quite quickly given the frequency of iterations when the pod is running.

maxguy2001 (changes requested):

test/navigation/imu_preprocessing.cpp
BenET01: Ive now used core::kEpsilon, is that a better order of magnitude?

maxguy2001: Possibly not. Leave it for now and we can talk about it a bit this evening (also weirdly enough the ideas in the error analysis section of your numerical ode course might be of interest but not directly applicable...). A build up of error is one of the things we have to be really careful of in the navigation team. We should be fine on a 100m track but we should be aiming to reduce all errors to allow the potential scaling of our system at a later date.

test/navigation/imu_preprocessing.cpp
kshxtij: Yeah, use auto

maxguy2001 (approved):
Looks good! Couple quick merge conflicts to sort out but otherwise ready to merge

test/navigation/imu_preprocessing.cpp
mifrandir: i commented in another place about how to do this from a code pov

as for error analysis: i think this would warrant something really cool. the HYPED could propose a stats or computing project doing this. i think the error problems combined with the computational limitations would make this really interesting. i think the project wouldn't require any actual C++ code as HYPED is probably better at doing that themselves. @kshxtij talk to the TDs? I'm happy to help with the proposal. Needs to be soon-ish, though.

lib/navigation/preprocess_imu.hpp
```diff
@@ -1,27 +1,98 @@
...
+    std::copy(
+      accelerometer_data.begin(), accelerometer_data.end(), accelerometer_data_copy.begin());
+
+    std::sort(accelerometer_data_copy.begin(), accelerometer_data_copy.end());
+    return {.q1     = getSpecificQuantile(accelerometer_data_copy, 0.25),
+            .median = getSpecificQuantile(accelerometer_data_copy, 0.5),
+            .q3     = getSpecificQuantile(accelerometer_data_copy, 0.75)};
+  }
 
   // initialised as {0, 0, 0, 0}, count of consecutive outliers
```
mifrandir, line 84: don't like these; people can just read this in the implementation?

lib/navigation/preprocess_imu.hpp
```diff
@@ -1,27 +1,98 @@
...
 
   // initialised as all true, bool mask of reliable sensors
   std::array<bool, core::kNumImus> are_imus_reliable_;
+
+  // number of allowed consecutive outliers from single accelerometer
+  static constexpr std::uint8_t kNumAllowedImuFailures_ = 20;
+
+  std::size_t num_reliable_accelerometers_;  // intitialised as= core::kNumImus
+
+  core::ILogger &logger_;
```
mifrandir, line 95: the ordering should put the class members in the order they are in the constructor; then all the internal structure after

long story short: this should be the first private member

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+TEST(Imu, equal_data)
+{
+  utils::ManualTime manual_time;
+  core::Logger logger("test", core::LogLevel::kFatal, manual_time);
+  navigation::ImuPreprocessor imu_processer(logger);
+  core::RawImuData data = {{1, 1, 1}};
+  core::ImuData answer  = {static_cast<core::Float>(std::sqrt(3.0))};
+  auto final_data       = imu_processer.processData(data);
```
mifrandir, line 39: fix `const`s

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,67 @@
...
+    for (std::size_t i; i < imu_data_a.size(); ++i) {
+      if (!(std::abs(imu_data_a.at(i) - imu_data_b.at(i)) < epsilon)) { return false; }
+    }
+  } else {
+    return false;
+  }
+  return true;
+}
+
+void test()
```
mifrandir, line 25: what is this good for?

mifrandir (changes requested):

test/navigation/imu_preprocessing.cpp
BenET01: This was just to instantiate the class to see if it compiled before the other tests were made, will just delete this now as the class is getting instantiated in every other test

test/navigation/imu_preprocessing.cpp
maxguy2001: @mifrandir this is a complete problem across navigation. I don't think we currently have an explicit method for reducing the error since the closest thing we have to a direct way of viewing the state of the dynamic system is the wheel encoder displacement (although the kalman filter does it implicitly from what I remember). Currently the approach is more "check we haven't gone too far wrong" than considering the bounds on the errors or implementing any sort of correction/rigorous error reduction methods. This would probably fit quite well as a computational statistics problem (investigating Bayesian filters) and would probably work best if it was implemented in some sort of code (e.g. python, R, etc.) so that implementation is at least considered. Once we get the car up and running, we should be able to provide plenty of real world data for working with.

lib/navigation/consts.hpp
```diff
@@ -17,6 +17,12 @@ static constexpr core::Float kStripeDistance    = 10.0;       // m TODOLater:che
 // define sensor checks return type
 enum class SensorChecks { kUnacceptable = 0, kAcceptable };
 
+struct Quartiles {
```
TomLonergan03, line 20: is consts.hpp the best name for this file? it seems to have picked up a bunch of definitions that aren't constants. 
(ik this is somewhat outside the scope of this PR sorry)

lib/navigation/preprocess_imu.cpp
```diff
@@ -2,55 +2,99 @@
...
+std::optional<core::ImuData> ImuPreprocessor::processData(const core::RawImuData raw_imu_data)
 {
-  /*
-  TODOLater: implement
-  basic plan:
-  - call detectOutliers. Return of that function is return of this function
-  - call checkReliable
+  core::ImuData imu_data;
+  // TODOLater : check on direction of travel
+  core::Float magnitude;
```
TomLonergan03, line 17: this should be defined within the loop as it's only needed in that scope imo

test/navigation/imu_preprocessing.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace hyped::test {
+
+core::Float epsilon = 1e-5;
+bool checkArrayEquality(const core::ImuData &imu_data_a, const core::ImuData &imu_data_b)
+{
+  if (imu_data_a.size() == imu_data_b.size()) {
+    for (std::size_t i; i < imu_data_a.size(); ++i) {
+      if (!(std::abs(imu_data_a.at(i) - imu_data_b.at(i)) < epsilon)) { return false; }
+    }
```
TomLonergan03, line 18: this can be restructured to use guards instead of nesting it so much.

ie `if data_a != data_b return false`
then the for loop

lib/navigation/preprocess_imu.cpp
```diff
@@ -2,55 +2,99 @@
 
 namespace hyped::navigation {
 
-ImuPreprocessor::ImuPreprocessor()
+ImuPreprocessor::ImuPreprocessor(core::ILogger &logger)
+    : logger_(logger),
+      num_outliers_per_imu_({0, 0, 0, 0}),
+      are_imus_reliable_({true, true, true, true}),
+      num_reliable_accelerometers_(core::kNumImus)
```
TomLonergan03, line 9: is there a rule for when imu vs accelerometer is used? here I feel like it should be either `are_imus_reliable_` and `num_reliable_imus_` or both accelerometer

test/navigation/imu_preprocessing.cpp
mifrandir: i agree fully; let's propose something

i assume you know some people in the school that do this sort of stuff?

lib/navigation/preprocess_imu.hpp
```diff
@@ -26,6 +26,17 @@ class ImuPreprocessor {
   std::optional<core::ImuData> processData(const core::RawImuData raw_imu_data);
 
  private:
+  core::ILogger &logger_;
+
+  std::array<std::uint16_t, core::kNumImus> num_outliers_per_imu_;
+
+  std::array<bool, core::kNumImus> are_imus_reliable_;
+
+  std::size_t num_reliable_accelerometers_;
```
mifrandir, line 35: remove whitespace

mifrandir (changes requested):

lib/navigation/preprocess_imu.cpp
maxguy2001: @BenET01 think sensors prefer accelerometer so stick to that for the sake of naming consistency?

kshxtij: +1 on accelerometer

test/navigation/imu_preprocessing.cpp
kshxtij: I think this is a great idea, do we know what the timelines in SoM/SoI look like for self proposed projects?

I've heard SoI is March but I can't find the dates on a website anywhere (or atleast a website i can access as a 2nd year)

kshxtij (approved):
LGTM, after fixing the minor changes above ^

test/navigation/imu_preprocessing.cpp
mifrandir: SoI is 3 March, but we need a supervisor.

mifrandir (approved):

lib/navigation/crosscheck.cpp
```diff
@@ -24,7 +24,8 @@ SensorChecks Crosscheck::checkTrajectoryAgreement(const core::ImuData imu_data,
   return SensorChecks::kAcceptable;
 }
 
-SensorChecks checkEncoderImu(const core::ImuData imu_data, const core::EncoderData encoders_data)
+SensorChecks checkEncoderImu(const core::AccelerometerData accelerometer_data,
```
TomLonergan03, line 27: `checkEncoderAccelerometer`

lib/navigation/navigator.cpp
```diff
@@ -40,7 +40,7 @@ void Navigator::encoderUpdate(const core::EncoderData &encoder_data)
   */
 }
 
-void Navigator::imuUpdate(const core::RawImuData &imu_data)
+void Navigator::imuUpdate(const core::RawAccelerometerData &accelerometer_data)
```
TomLonergan03, line 43: `accelerometerUpdate`

TomLonergan03 (changes requested):
also if accelerometer is being used instead of imu everywhere then it should be preprocess_accelerometer.[ch]pp and same for tests

lib/navigation/navigator.hpp
```diff
@@ -32,9 +32,9 @@ class Navigator : public INavigator {
   /**
    * @brief preprocesses imu data and updates trajectory
```
TomLonergan03, line 33: one more imu -> accelerometer then good to go

TomLonergan03 (changes requested):

lib/navigation/navigator.hpp
BenET01: damn, forgot to check the other files

TomLonergan03 (approved):

PR #40, opened 2022-11-07
Nav - Keyence Preprocessing 

Fayz-S:
Implemented preprocess_keyence constructor and checkKeyenceAgrees function. Also changed is_keyence_agreeing_ to has_keyence_disagreed_ to improve readability

### DIFF
--- a/lib/navigation/preprocess_keyence.cpp
+++ b/lib/navigation/preprocess_keyence.cpp
@@ -28,7 +28,6 @@ SensorChecks KeyencePreprocessor::checkKeyenceAgrees(const core::KeyenceData &ke
   } else if (!(keyence_data_disagrees) && has_keyence_disagreed_) {
     has_keyence_disagreed_ = false;
 
-  } else {
     return SensorChecks::kAcceptable;
   }
 }
--- a/lib/navigation/preprocess_keyence.cpp
+++ b/lib/navigation/preprocess_keyence.cpp
@@ -2,17 +2,20 @@
 
 namespace hyped::navigation {
 
-KeyencePreprocessor::KeyencePreprocessor()
+KeyencePreprocessor::KeyencePreprocessor(hyped::core::ILogger &logger) : logger_(logger)
 {
-  // TODOLater: implement
+  // Currently no extra parameters so constructor is empty
+  has_keyence_disagreed_ = false;
 }
 
-SensorChecks checkKeyenceAgrees(const core::KeyenceData &keyence_data)
+SensorChecks KeyencePreprocessor::checkKeyenceAgrees(const core::KeyenceData &keyence_data)
 {
   /*
    TODOLater: implement:
    roughly:
+    if 
    - if keyence data disagrees and does_keyence_agree_ is false, fail state, return kUnacceptable
+
    - if keyence data disagrees and does_keyence_agree_ is true, set does_keyence_agree_ false return
    kAcceptable
    - if keyence data agrees and does_keyence_agree_ is false, et does_keyence_agree_ true return
@@ -20,6 +23,23 @@ SensorChecks checkKeyenceAgrees(const core::KeyenceData &keyence_data)
    - if keyence data agrees and does_keyence_agree_ is true, all is good in the world, return
    kAcceptable.
   */
+  bool keyence_data_disagrees = false;
+  for (size_t i; i < keyence_data.size() - 1; i++){
+    if (keyence_data.at(i) != keyence_data.at(i+1)){
+      keyence_data_disagrees = true;
+    }
+  }
+  if (keyence_data_disagrees && has_keyence_disagreed_){
+    logger_.log(hyped::core::LogLevel::kFatal, "Keyence disagreed more than once");
+    return SensorChecks::kUnacceptable;
+  }
+  else if (keyence_data_disagrees && !(has_keyence_disagreed_)){
+    has_keyence_disagreed_ = true;
+  }
+  else if (!(keyence_data_disagrees) && has_keyence_disagreed_){
+    has_keyence_disagreed_ = false;
+
+  }
   return SensorChecks::kAcceptable;
 }
 
--- a/lib/navigation/preprocess_keyence.cpp
+++ b/lib/navigation/preprocess_keyence.cpp
@@ -20,18 +20,10 @@ SensorChecks KeyencePreprocessor::checkKeyenceAgrees(const core::KeyenceData &ke
 
   if (current_data_status == KeyenceDataStatus::kDisagreed
       && previous_data_status_ == KeyenceDataStatus::kDisagreed) {
-    
     // TODOLater: Implement logging
     return SensorChecks::kUnacceptable;
   }
-  if (current_data_status == KeyenceDataStatus::kDisagreed
-      && previous_data_status_ == KeyenceDataStatus::kAgreed) {
-    previous_data_status_ = KeyenceDataStatus::kDisagreed;
-  }
-  if (current_data_status == KeyenceDataStatus::kAgreed
-      && previous_data_status_ == KeyenceDataStatus::kDisagreed) {
-    previous_data_status_ = current_data_status;
-  }
+  previous_data_status_ = current_data_status;
 
   return SensorChecks::kAcceptable;
 }
--- a/lib/navigation/preprocess_keyence.cpp
+++ b/lib/navigation/preprocess_keyence.cpp
@@ -27,9 +27,9 @@ SensorChecks KeyencePreprocessor::checkKeyenceAgrees(const core::KeyenceData &ke
 
   } else if (!(keyence_data_disagrees) && has_keyence_disagreed_) {
     has_keyence_disagreed_ = false;
-
-    return SensorChecks::kAcceptable;
   }
+
+  return SensorChecks::kAcceptable;
 }
 
 }  // namespace hyped::navigation
--- a/lib/navigation/preprocess_keyence.cpp
+++ b/lib/navigation/preprocess_keyence.cpp
@@ -20,8 +20,7 @@ SensorChecks KeyencePreprocessor::checkKeyenceAgrees(const core::KeyenceData &ke
 
   if (current_data_status == KeyenceDataStatus::kDisagreed
       && previous_data_status_ == KeyenceDataStatus::kDisagreed) {
-    log_.log(core::LogLevel::kFatal,
-             "Keyence disagreement for two consecutive readings.");
+    log_.log(core::LogLevel::kFatal, "Keyence disagreement for two consecutive readings.");
     return SensorChecks::kUnacceptable;
   }
   previous_data_status_ = current_data_status;
--- a/lib/navigation/preprocess_keyence.cpp
+++ b/lib/navigation/preprocess_keyence.cpp
@@ -2,7 +2,7 @@
 
 namespace hyped::navigation {
 
-KeyencePreprocessor::KeyencePreprocessor(hyped::core::ILogger &logger) : logger_(logger)
+KeyencePreprocessor::KeyencePreprocessor(hyped::core::ILogger &logger) : log_(logger)
 {
   // Currently no extra parameters so constructor is empty
   has_keyence_disagreed_ = false;
@@ -27,14 +27,18 @@ SensorChecks KeyencePreprocessor::checkKeyenceAgrees(const core::KeyenceData &ke
   for (size_t i = 0; i < keyence_data.size() - 1; ++i) {
     if (keyence_data.at(i) != keyence_data.at(i + 1)) { keyence_data_disagrees = true; }
   }
+
   if (keyence_data_disagrees && has_keyence_disagreed_) {
     log_.log(hyped::core::LogLevel::kFatal, "Keyence disagreed more than once");
     return SensorChecks::kUnacceptable;
+
   } else if (keyence_data_disagrees && !(has_keyence_disagreed_)) {
     has_keyence_disagreed_ = true;
+
   } else if (!(keyence_data_disagrees) && has_keyence_disagreed_) {
     has_keyence_disagreed_ = false;
   }
+
   return SensorChecks::kAcceptable;
 }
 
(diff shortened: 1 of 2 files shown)


lib/navigation/preprocess_keyence.cpp
```diff
@@ -2,24 +2,44 @@
...
+
    - if keyence data disagrees and does_keyence_agree_ is true, set does_keyence_agree_ false return
    kAcceptable
    - if keyence data agrees and does_keyence_agree_ is false, et does_keyence_agree_ true return
    kAcceptable
    - if keyence data agrees and does_keyence_agree_ is true, all is good in the world, return
    kAcceptable.
   */
+  bool keyence_data_disagrees = false;
+  for (size_t i; i < keyence_data.size() - 1; i++){
```
maxguy2001, line 27: `size_t i = 0; i < keyence_data.size(); ++i`

maxguy2001: you didn't initialize i and you're post incrementing

maxguy2001 (changes requested):

lib/navigation/preprocess_keyence.hpp
```diff
@@ -4,17 +4,22 @@
...
+ /**
+  * @brief Construct a new Keyence Preprocessor object
+  */
+  KeyencePreprocessor(hyped::core::ILogger &logger);
 
   SensorChecks checkKeyenceAgrees(const core::KeyenceData &keyence_data);
 
  private:
-  bool is_keyence_agreeing_;
+  hyped::core::ILogger &logger_;
```
maxguy2001, line 21: `&log_` seems to be the codebase default?

Fayz-S: logger_  was used in adc class in io, which is where I pulled this code from . Checked their other files and log_ is used, so I'll change this. @SnickeyX you'll want to know this.

lib/navigation/preprocess_keyence.hpp
```diff
@@ -3,18 +3,23 @@
...
 
 namespace hyped::navigation {
 
 class KeyencePreprocessor {
  public:
-  KeyencePreprocessor();
+  /**
+   * @brief Construct a new Keyence Preprocessor object
+   */
+  KeyencePreprocessor(hyped::core::ILogger &log_);
```
maxguy2001, line 16: Docstring

lib/navigation/preprocess_keyence.hpp
```diff
@@ -3,18 +3,23 @@
...
 
   SensorChecks checkKeyenceAgrees(const core::KeyenceData &keyence_data);
 
  private:
-  bool is_keyence_agreeing_;
+  hyped::core::ILogger &log_;
+  bool has_keyence_disagreed_;
```
maxguy2001, line 22: comment describing briefly purpose of this flag

lib/navigation/preprocess_keyence.hpp
```diff
@@ -3,18 +3,23 @@
 #include <array>
```
maxguy2001, line 3: not used

lib/navigation/preprocess_keyence.hpp
```diff
@@ -3,18 +3,23 @@
...
 #include <cstdint>
```
maxguy2001, line 4: should be in cpp file since not used in header

lib/navigation/preprocess_keyence.cpp
```diff
@@ -2,24 +2,43 @@
...
 {
-  // TODOLater: implement
+  // Currently no extra parameters so constructor is empty
+  has_keyence_disagreed_ = false;
 }
 
-SensorChecks checkKeyenceAgrees(const core::KeyenceData &keyence_data)
+SensorChecks KeyencePreprocessor::checkKeyenceAgrees(const core::KeyenceData &keyence_data)
 {
   /*
```
maxguy2001, line 13: remove very large comment and comment any bits of code where the purpose is not immediately obvious.

lib/navigation/preprocess_keyence.cpp
```diff
@@ -2,24 +2,43 @@
...
+
    - if keyence data disagrees and does_keyence_agree_ is true, set does_keyence_agree_ false return
    kAcceptable
    - if keyence data agrees and does_keyence_agree_ is false, et does_keyence_agree_ true return
    kAcceptable
    - if keyence data agrees and does_keyence_agree_ is true, all is good in the world, return
    kAcceptable.
   */
+  bool keyence_data_disagrees = false;
+  for (size_t i = 0; i < keyence_data.size() - 1; ++i) {
```
maxguy2001, line 27: `std::size_t` now

lib/navigation/preprocess_keyence.cpp
```diff
@@ -2,24 +2,43 @@
...
    - if keyence data disagrees and does_keyence_agree_ is false, fail state, return kUnacceptable
+
    - if keyence data disagrees and does_keyence_agree_ is true, set does_keyence_agree_ false return
    kAcceptable
    - if keyence data agrees and does_keyence_agree_ is false, et does_keyence_agree_ true return
    kAcceptable
    - if keyence data agrees and does_keyence_agree_ is true, all is good in the world, return
    kAcceptable.
   */
+  bool keyence_data_disagrees = false;
```
maxguy2001, line 26: comment what the purpose of this is probably (we both know because we fully understand the desired functionality but someone looking at this fresh might not know).

lib/navigation/preprocess_keyence.cpp
```diff
@@ -2,24 +2,43 @@
...
+    log_.log(hyped::core::LogLevel::kFatal, "Keyence disagreed more than once");
+    return SensorChecks::kUnacceptable;
+
+  } else if (keyence_data_disagrees && !(has_keyence_disagreed_)) {
+    has_keyence_disagreed_ = true;
+
+  } else if (!(keyence_data_disagrees) && has_keyence_disagreed_) {
+    has_keyence_disagreed_ = false;
+  }
+
```
maxguy2001, line 41: maybe use `else` here to show that we have a default case

maxguy2001 (changes requested):
Minor fixes remaining and then convert to full pr to get other reviews and merged!

maxguy2001 (approved):
I'll allow it

lib/navigation/preprocess_keyence.hpp
mifrandir: `last_keyence_measurement_was_in_disagreement_`? Or is this getting to verbose? I find boolean flags terribly difficult to name.

Here's a better approach: make an enum class, say `KeyenceAgreement`, and call the field `last_measurement_agreement_`? The values of `KeyenceAgreement` could be `kAgree` and `kDisagree` or similar.

mifrandir (changes requested):

lib/navigation/preprocess_keyence.hpp
```diff
@@ -25,9 +27,9 @@ class KeyencePreprocessor {
    *
    */
  private:
-  hyped::core::ILogger &log_;
-  // Flag for if the previous keyence measurements disagreed.
-  bool has_keyence_disagreed_;
+  core::ILogger &log_;
+  // Enum variable for if the previous keyence data disagreed.
```
mifrandir, line 31: you have just made this comment useless; the code explains all that is needed :)

lib/navigation/preprocess_keyence.cpp
```diff
@@ -4,28 +4,32 @@
...
+    if (keyence_data.at(i) != keyence_data.at(i + 1)) {
+      current_data_status = KeyenceDataStatus::kDisagreed;
+    }
   }
 
-  if (keyence_data_disagrees && has_keyence_disagreed_) {
-    log_.log(hyped::core::LogLevel::kFatal, "Keyence disagreed more than once");
+  if (current_data_status == KeyenceDataStatus::kDisagreed
+      && previous_data_status_ == KeyenceDataStatus::kDisagreed) {
+    log_.log(core::LogLevel::kFatal, "Keyence disagreed more than once");
```
mifrandir, line 23: We need to be careful with the logging here; if keyence gets out of sync once this will be spammed. Not sure how to solve this now, maybe add a `// TODOLater`?

lib/navigation/preprocess_keyence.cpp
```diff
@@ -4,28 +4,32 @@
...
-  } else if (!(keyence_data_disagrees) && has_keyence_disagreed_) {
-    has_keyence_disagreed_ = false;
+  }
+  if (current_data_status == KeyenceDataStatus::kDisagreed
+      && previous_data_status_ == KeyenceDataStatus::kAgreed) {
+    previous_data_status_ = KeyenceDataStatus::kDisagreed;
+  }
+  if (current_data_status == KeyenceDataStatus::kAgreed
+      && previous_data_status_ == KeyenceDataStatus::kDisagreed) {
+    previous_data_status_ = KeyenceDataStatus::kAgreed;
```
mifrandir, line 32: These two clauses are equivalent to `previous_data_status_ = current_data_status`.

mifrandir (changes requested):

lib/navigation/preprocess_keyence.cpp
kshxtij: Could maintain a counter and check that counter while logging but i'll have a think about something similar to glog's `LOG_ONLY_ONCE` and `LOG_ONLY_N`

mifrandir: Yeah that shouldn't be too hard to implement... though it would require some template/macro shenanigans if you want to keep it usable.

kshxtij: I think the simplest solution is having `std::unordered_map<std::string, int> log_counts_;` in logger and 
```cpp
void Logger::logNTimes(const LogLevel level, int n, const char *format, ...)
{
  int &log_count = log_counts_.find(format)->second;
  if (log_count < n) {
    log(level, format);
    log_count++;
  }
}
```

Most standard logging libraries have something similar, I'll dig into some other definitions to see how they do it.

lib/navigation/preprocess_keyence.cpp
```diff
@@ -29,7 +30,7 @@ SensorChecks KeyencePreprocessor::checkKeyenceAgrees(const core::KeyenceData &ke
   }
   if (current_data_status == KeyenceDataStatus::kAgreed
       && previous_data_status_ == KeyenceDataStatus::kDisagreed) {
```
mifrandir, line 32: you can get rid of all of these checks!

lib/navigation/preprocess_keyence.hpp
```diff
@@ -1,20 +1,34 @@
...
+   * @return SensorChecks: enum indicating if Keyence Sensors have failed
+   */
   SensorChecks checkKeyenceAgrees(const core::KeyenceData &keyence_data);
-
+  /**
+   *
+   */
  private:
-  bool is_keyence_agreeing_;
+  core::ILogger &log_;
```
mifrandir, line 30: We're calling this `logger_` everywhere else now.

lib/navigation/preprocess_keyence.cpp
```diff
@@ -1,25 +1,31 @@
...
-   - if keyence data agrees and does_keyence_agree_ is true, all is good in the world, return
-   kAcceptable.
-  */
+  KeyenceDataStatus current_data_status = KeyenceDataStatus::kDisagreed;
+  for (std::size_t i = 0; i < keyence_data.size() - 1; ++i) {
+    if (keyence_data.at(i) != keyence_data.at(i + 1)) {
+      current_data_status = KeyenceDataStatus::kDisagreed;
+    }
+  }
+
```
mifrandir, line 20: empty line

lib/navigation/preprocess_keyence.cpp
```diff
@@ -1,25 +1,31 @@
...
-   - if keyence data disagrees and does_keyence_agree_ is true, set does_keyence_agree_ false return
-   kAcceptable
-   - if keyence data agrees and does_keyence_agree_ is false, et does_keyence_agree_ true return
-   kAcceptable
-   - if keyence data agrees and does_keyence_agree_ is true, all is good in the world, return
-   kAcceptable.
-  */
+  KeyenceDataStatus current_data_status = KeyenceDataStatus::kDisagreed;
+  for (std::size_t i = 0; i < keyence_data.size() - 1; ++i) {
+    if (keyence_data.at(i) != keyence_data.at(i + 1)) {
```
mifrandir, line 16: Why are we only comparing neighbours? What's going on here?

If the fundamental assumption is that there will only ever be two keyence readings then just compare those two directly?

lib/navigation/preprocess_keyence.cpp
```diff
@@ -1,25 +1,31 @@
...
+      current_data_status = KeyenceDataStatus::kDisagreed;
+    }
+  }
+
+  if (current_data_status == KeyenceDataStatus::kDisagreed
+      && previous_data_status_ == KeyenceDataStatus::kDisagreed) {
+    log_.log(core::LogLevel::kFatal,
```
mifrandir, line 23: I would like to see a comment somewhere explaining the rationale behind this outlier detection other than "it's arbitrarily flagging some outliers"

mifrandir (approved):

kshxtij:
Fix the formatting

Fayz-S:
Apologies, post franz's approval I've not been checking this branch. Will fix when I'm back home.  Also to answer your question about checking neighbours- max wanted time to generalise such that if we had more thanc2 sensors, the code will still work.

Merry Christmas,
Fayz
________________________________

lib/navigation/preprocess_keyence.cpp
maxguy2001: This is designed without assuming that there will only ever be 2 keyence readings. Therefore we set a disagreement flag if any keyence sensor disagrees with the others and check it on the next iteration (making comparison of neighbours a reasonable method surely?). 2 disagreements in a row is a critical failure.

kshxtij (approved):

PR #52, opened 2022-11-17
SNS - ADC fix

SnickeyX:
The adc value ranges from 1-4 digits but we were erroring whenever we didnt read 4 bytes, so adc values from 0-99 coudn't be read. This pr fixes this and has been tested in the lab with and without the debugger (it worked when we gave an input of a sine wave from 0-1.8V and were able to successfully read values from 0-4095 inclusive).

### DIFF
--- a/lib/io/adc.cpp
+++ b/lib/io/adc.cpp
@@ -40,7 +40,7 @@ std::optional<std::uint16_t> Adc::resetAndRead4(const int file_descriptor)
     = read(file_descriptor,
            &read_buffer,
            sizeof(read_buffer));  // actually consume new data, changes value in buffer
-  if (num_bytes_read < 2) {       // 2 bytes minimum as value ranges from 1 to 4 digits
+  if (num_bytes_read < 2) {       // 2 bytes minimum as value ranges from [0, 4095]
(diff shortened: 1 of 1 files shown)


lib/io/adc.cpp
```diff
@@ -40,9 +40,8 @@ std::optional<std::uint16_t> Adc::resetAndRead4(const int file_descriptor)
     = read(file_descriptor,
            &read_buffer,
            sizeof(read_buffer));  // actually consume new data, changes value in buffer
-  if (num_bytes_read != sizeof(read_buffer)) {
-    log_.log(core::LogLevel::kFatal, "Failed to read 4 bytes");
-
+  if (num_bytes_read < 2) {       // 2 bytes minimum as value ranges from 1 to 4 digits
```
mifrandir, line 43: i feel like the use of the word digits here is confusing; what base are you talking about? decimal? binary? hex? maybe put the exact range instead?

mifrandir (changes requested):

mifrandir (approved):

maxguy2001 (approved):

kshxtij (approved):

PR #53, opened 2022-11-17
SNS - Mux Failure Threshold

SnickeyX:
We ideally don't want to straight away return a nullopt whenever a particular sensor on the mux doesnt work - we only want to do this when a X amount of sensors behave this way - in which case we can assume the mux is broken.  At the moment, the threshold is at 25%, however this could very well be changed. 

For when a particular sensor doesn't work and we are under the failure threshold limit, we simply assign a 0 to the value we "read" from it. During normal behaviour, we should not expect a 0 value from any of the sensors* on the mux anyway. 

*to be confirmed with electronics of course 
-  only thing that I could think of was the accelerometer giving us 0 initially / sometime during deceleration but it still won't be 0 in all axes - thanku gravity
-  and also if temp reaches 0, we have bigger problems at hand


lib/sensors/mux.cpp
```diff
@@ -67,15 +76,17 @@ std::optional<std::array<T, N>> Mux<T, N>::readAllChannels()
     const auto configure_result = sensor->configure();
     if (configure_result == core::Result::kFailure) {
       log_.log(core::LogLevel::kFatal, "Failed to configure sensor at mux channel %d", channel);
-      return std::nullopt;
+      num_unusable_sensors++;
```
mifrandir, line 79: `++x`

lib/sensors/mux.cpp
```diff
@@ -67,15 +76,17 @@ std::optional<std::array<T, N>> Mux<T, N>::readAllChannels()
...
+      continue;
     }
     // Finally read sensor data
     const std::optional<T> sensor_data = sensor->read();
     if (sensor_data.has_value()) {
       mux_data.at(i) = sensor_data.value();
     } else {
       log_.log(core::LogLevel::kFatal, "Failed to get mux data from channel %d", channel);
-      return std::nullopt;
+      num_unusable_sensors++;
```
mifrandir, line 88: again

lib/sensors/mux.hpp
```diff
@@ -28,6 +30,8 @@ class Mux {
   core::Result closeAllChannels();
   core::ILogger &log_;
   io::I2c i2c_;
+  std::uint8_t num_unusable_sensors;
```
mifrandir, line 33: not sure this should be a class variable

lib/sensors/mux.cpp
```diff
@@ -53,8 +55,15 @@ core::Result Mux<T, N>::closeAllChannels()
 template<typename T, std::uint8_t N>
 std::optional<std::array<T, N>> Mux<T, N>::readAllChannels()
 {
-  std::array<T, N> mux_data;
+  // Zero-initialize the array
+  std::array<T, N> mux_data{};
   for (std::uint8_t i = 0; i < N; ++i) {
+    if (num_unusable_sensors > max_num_unusable_sensors_) {
```
mifrandir, line 61: This should move to the end; what if the last sensor is unusable and reaches the threshold?

lib/sensors/mux.cpp
```diff
@@ -67,15 +76,17 @@ std::optional<std::array<T, N>> Mux<T, N>::readAllChannels()
     const auto configure_result = sensor->configure();
     if (configure_result == core::Result::kFailure) {
       log_.log(core::LogLevel::kFatal, "Failed to configure sensor at mux channel %d", channel);
-      return std::nullopt;
+      num_unusable_sensors++;
+      continue;
     }
     // Finally read sensor data
     const std::optional<T> sensor_data = sensor->read();
```
mifrandir, line 83: why is this not just `auto`?

lib/sensors/mux.cpp
```diff
@@ -67,15 +76,17 @@ std::optional<std::array<T, N>> Mux<T, N>::readAllChannels()
...
     if (sensor_data.has_value()) {
```
mifrandir, line 84: just `if (sensor_data)` should be fine

mifrandir (commented):
Questionable logic I would say.

SnickeyX:
> Questionable logic I would say.

I agree to some extent. This threshold should ideally not be the same for every mux but since on the hardware-end we haven't even got a working mux yet - I will leave the finer details for later. This PR's purpose is simply to say we at least thought of this issue.

lib/sensors/mux.hpp
```diff
@@ -13,6 +13,7 @@ static constexpr std::uint8_t kDefaultMuxAddress = 0x70;
 static constexpr std::uint8_t kMaxNumChannels    = 8;
 static constexpr core::Float kFailureThreshold
   = 0.25;  // TODOLater: finalize this value with Electronics
+std::uint8_t num_unusable_sensors = 0;
```
mifrandir, line 16: no, no, no, no! don't put it here. now it's a program-wide global variable. put it inside the function!

mifrandir (changes requested):

lib/sensors/mux.hpp
SnickeyX: oh whoops, will fix asap.

lib/sensors/mux.cpp
```diff
@@ -53,7 +54,9 @@ core::Result Mux<T, N>::closeAllChannels()
 template<typename T, std::uint8_t N>
 std::optional<std::array<T, N>> Mux<T, N>::readAllChannels()
 {
-  std::array<T, N> mux_data;
+  // Zero-initialize the array
+  std::array<T, N> mux_data{};
+  std::uint8_t num_unusable_sensors = 0;
   for (std::uint8_t i = 0; i < N; ++i) {
     const auto sensor          = sensors_[i];
```
mifrandir, line 61: I just noticed this: This is a copy. Make the lhs a reference.

lib/sensors/mux.cpp
```diff
@@ -67,21 +70,30 @@ std::optional<std::array<T, N>> Mux<T, N>::readAllChannels()
...
       log_.log(core::LogLevel::kFatal, "Failed to configure sensor at mux channel %d", channel);
-      return std::nullopt;
+      ++num_unusable_sensors;
+      continue;
     }
     // Finally read sensor data
-    const std::optional<T> sensor_data = sensor->read();
-    if (sensor_data.has_value()) {
+    const auto sensor_data = sensor->read();
+    if (sensor_data) {
```
mifrandir, line 78: not a massive fan of this

how about
```cpp
if (!sensor_data) {
  // handle error
  continue;
}
mux_data.at(i) = sensor_data.value();
```

lib/sensors/mux.cpp
mifrandir: Also the RHS should use `.at(i)` for consistency.

mifrandir (changes requested):
getting there :)

lib/sensors/mux.cpp
SnickeyX: oh yes, this could have been really bad - thanks for pointing it out

mifrandir: well, if it could have been really bad then you change your class to not allow this. for me it's a performance thing. if you don't want these sensors to be copied then you should go and remove the copy constructors.

SnickeyX: i tried doing this for II2cMuxSensors instances but that didn't work. now with just a reference to the array, without templating, the same code throws errors if you don't do `&sensor` but with templating doesn't. am a bit confused.

SnickeyX: `shared_ptr` on the other hand does seem to work

lib/sensors/mux.hpp
```diff
@@ -29,10 +29,10 @@ class Mux {
   core::Result selectChannel(const std::uint8_t channel);
   core::Result closeAllChannels();
   core::ILogger &log_;
-  io::I2c i2c_;
+  io::I2c &i2c_;
   const std::uint8_t max_num_unusable_sensors_;
   const std::uint8_t mux_address_;
-  const std::array<std::unique_ptr<II2cMuxSensor<T>>, N> sensors_;
+  const std::array<std::unique_ptr<II2cMuxSensor<T>>, N> &sensors_;
```
mifrandir, line 35: I feel like this is dangerous. Surely we are fine with the mux owning the sensors? You don't want to be required to make sure that the sensors aren't deallocated while the mux is still around.

Moving the unique pointers into the class should be sufficient.

SnickeyX: Thing is, if we don't do this, the unique pointers get deleted as soon as we do `sensors_(std::move(sensors))`. This error doesn't come up when we do templating which is what is concerning.

mifrandir: What do you mean they get deleted? They're just moved into the array. Of course you can't access the old pointers anymore.

SnickeyX: Essentially [this](https://stackoverflow.com/questions/39724272/error-call-to-implicitly-deleted-copy-constructor-of-std-1unique-ptra-s#:~:text=Your%20constructor%20argument%20is%20pass%20by%20value%20which%20will%20make%20a%20copy%2C%20but%20you%20cannot%20copy%20a%20std%3A%3Aunique_ptr.%20Passing%20by%20reference%20should%20work%3A)

SnickeyX: Now, if we do this and try to do `sensor  = sensors_.at(i)` without lhs referencing (this is without templating)
image

mifrandir: There is a fix in #58. It required some housekeeping to make a test that actually instantiates the class. One thing you've done wrong, which I have failed to notices, is that you've put your templates in .cpp files. This is not what you want! A .cpp only containing templates is useless as there will be no code generated and the templates aren't accessible to other compilation units.

mifrandir: The TL;DR of the fix is: Don't make your sensors array `const` if you're going to move them.

SnickeyX: Okay keeping everything to `.hpp` is something I completely didn't realise you could do with templating but makes a lot of sense. Also agree with storing `sensors_` and not a reference without `const` (did not get why it kept copying but this makes sense). Thanks for fixing tho.

lib/sensors/mux.cpp
mifrandir: templates aren't checked unless you put a concrete value in because no code is generated; this is why we need the dummy tests! (among other reasons)

lib/sensors/mux.hpp
```diff
@@ -27,9 +30,107 @@ class Mux {
...
+      continue;
+    }
+    mux_data.at(i)                    = sensor_data.value();
+    const auto closing_channel_result = closeAllChannels();
+    if (closing_channel_result == core::Result::kFailure) {
+      log_.log(core::LogLevel::kFatal, "Failed to close all mux channels while reading");
+      return std::nullopt;
+    }
+    // If too many sensors are unusable, assume mux is faulty
+    if (num_unusable_sensors > max_num_unusable_sensors_) {
```
mifrandir, line 126: hm, this check is weird.  i think it should live outside the loop?

we always `continue` whenever the counter is incremented so at least one sensor has to work for us to get here...

lib/utils/dummy_i2c.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <functional>
```
mifrandir, line 3: unused import, sorry

mifrandir (changes requested):
Lots of things I really like about this:

1. Good, helpful comments.
2. Mostly clean and uncomplicated control flow.
3. Minimal use of mutable data.

lib/sensors/mux.hpp
SnickeyX: good point, have changed this

lib/io/hardware_i2c.hpp
```diff
@@ -2,10 +2,7 @@
 
 #include "i2c.hpp"
 
-#include <optional>
-
 #include <core/logger.hpp>
-#include <core/types.hpp>
```
mifrandir: this file uses `core::Result` which comes from the header; wouldn't say it's unused

why do we care? well, i assume `core::Result` comes from one of the other headers, say `i2c.hpp`. what happens if `i2c.hpp` doesn't need `core/types.hpp` anymore? removing the import will cause errors in lots of other files and lead to very weird diffs. so you should import the things that are used in this file (even if the imports don't do anything, technically).

mifrandir (changes requested):

lib/io/hardware_i2c.hpp
SnickeyX: thats fair, will add it back

kshxtij (approved):

maxguy2001 (approved):

PR #54, opened 2022-11-17
Stm outline

DylanCavers:
Start of the outline of the state machine code, is very far from done. This will be the basic structure of the state machine instances with appropriate states and transitions.


lib/state_machine/state_machine.cpp
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// TODOLater implement
+void StateMachine::checkTransition()
+{
+  /*
+  update data
+  should check if a transition is needed with current data
+  then run transition to move to next state
+  return (None || transition_message)
```
kshxtij, line 16: This should be an `std::optional` return type then

kshxtij (changes requested):

lib/state_machine/state_machine.cpp
```diff
@@ -0,0 +1,35 @@
...
+  */
+}
+
+// Transition to next state
+void StateMachine::transition(Message message)
```
maxguy2001, line 21: `const`

lib/state_machine/state_machine.cpp
```diff
@@ -0,0 +1,35 @@
...
+  for (Transition transition : transitions) {
+    if (transition.from == current_state && transition.message == message) {
+      current_state = transition.to;
+    }
+  }
+}
+
+void StateMachine::reset()
+{
+  current_state = states[0];
```
maxguy2001, line 32: Isn't the point of having an enum class so that you can just set `current_state = State::kInitialState`? Idk if I'm missing something here but why are we instantiating a local object that is kind of similar to the enum class we already defined?

lib/state_machine/state_machine.hpp
```diff
@@ -0,0 +1,33 @@
...
+  StateMachine();
+
+  void transition(Message message);
+
+  void checkTransition();
+
+  void reset();
+
+  // TODOLater change states to actual states
+  State states[5]
```
maxguy2001, line 19: Why do we do this?

lib/state_machine/state_machine.hpp
```diff
@@ -0,0 +1,33 @@
...
+    = {State::kInitialState, State::state1, State::state2, State::state3, State::state4};
+
+  // TODOLater change transitions to actual transitions
+  Transition transitions[5] = {
```
maxguy2001, line 23: You've implicitly used the number of states a few times now without stating where the number 5 comes from. Better design would be to create a variable somewhere (e.g. `kNumStates = 5`) and refer to that?

lib/state_machine/state_machine.hpp
```diff
@@ -0,0 +1,33 @@
...
+    Transition(State::kInitialState, State::state1, Message::mStart),
+    Transition(State::state1, State::state2, Message::message1),
+    Transition(State::state1, State::state3, Message::message2),
+    Transition(State::state2, State::state4, Message::message2),
+    Transition(State::state3, State::state4, Message::message1),
+  };
+
+  State current_state = states[0];
```
maxguy2001, line 31: Would this not be initialized in constructor?

lib/state_machine/transition.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include "state.hpp"
+
+namespace hyped::state_machine {
+
+// Transition class.
+// Each transition has a current state and next state, as well as a message indicating the kind of
+// transition.
+class Transition {
+ public:
+  Transition(State from_state, State to_state, Message transition_message)
```
maxguy2001, line 13: Should this not be a separate function within this class? Also why are we storing all of this as class members?

maxguy2001 (changes requested):
A few structural changes.

lib/state_machine/transition.hpp
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "message.cpp"
```
maxguy2001, line 3: Don't include a cpp file. Always include headers (all functions, classes, etc. should be defined there anyways).

lib/state_machine/state_machine.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+class StateMachine {
+ public:
+  StateMachine() { current_state = State::kInitialState; };
+
+  void transition(const Message message);
+
+  std::optional<Message> checkTransition();
+
+  void reset();
```
maxguy2001, line 19: This is not defined in your set of transitions so how do you define this behavior?

maxguy2001 (changes requested):

PR #58, opened 2022-11-24
Abstract I2c, I2c tests, and fix unique pointer

mifrandir:


lib/io/hardware_i2c.hpp
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include "i2c.hpp"
+
+#include <optional>
```
SnickeyX, line 5: already included in `i2c.hpp` - and `types.hpp` is already included too

lib/utils/dummy_i2c.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+std::optional<std::uint8_t> DummyI2c::readByte(const std::uint8_t device_address,
+                                               const std::uint8_t register_address)
+{
+  return std::nullopt;
+}
+
+core::Result DummyI2c::writeByteToRegister(const std::uint8_t device_address,
+                                           const std::uint8_t register_address,
+                                           const std::uint8_t)
```
SnickeyX, line 17: `data`

lib/io/hardware_i2c.hpp
```diff
@@ -0,0 +1,33 @@
...
+ public:
+  HardwareI2c(const std::uint8_t bus_address, core::ILogger &log);
+  ~HardwareI2c();
+
+  std::optional<std::uint8_t> readByte(const std::uint8_t device_address,
+                                       const std::uint8_t register_address);
+  core::Result writeByteToRegister(const std::uint8_t device_address,
+                                   const std::uint8_t register_address,
+                                   const std::uint8_t data);
+  core::Result writeByte(const std::uint8_t device_address, std::uint8_t data);
```
SnickeyX, line 22: `const` for data

lib/utils/dummy_i2c.cpp
```diff
@@ -0,0 +1,27 @@
...
+{
+  return core::Result::kFailure;
+}
+
+core::Result DummyI2c::writeByte(const std::uint8_t device_address, std::uint8_t data)
```
SnickeyX, line 22: `const`

lib/utils/dummy_i2c.hpp
```diff
@@ -0,0 +1,28 @@
...
+class DummyI2c : public io::II2c {
+ public:
+  DummyI2c();
+
+  virtual std::optional<std::uint8_t> readByte(const std::uint8_t device_address,
+                                               const std::uint8_t register_address);
+  virtual core::Result writeByteToRegister(const std::uint8_t device_address,
+                                           const std::uint8_t register_address,
+                                           const std::uint8_t data);
+  virtual core::Result writeByte(const std::uint8_t device_address, std::uint8_t data);
```
SnickeyX, line 25: `const`

SnickeyX (changes requested):
Looks great! Plan right now for me is to get all IO code (want to add SPI, PWM and get GPIO perfected) for bbb implemented and tested over dec/jan before moving onto working on these tests. @kshxtij has planned a hackathon in early sem2 during which I want to focus on getting the IO tests implemented. 

Plan for tests rn is to mock reasonable data (within the range given by the datasheet) for each type of sensor on each type of IO (*where appliacable, PWM will be more interesting) and then read using the sensor code we have with the `dummy_io` instance. This way we can safely say that within the scope of every sensor we use, we have ensured our code works for whatever value we can possibly get from it.

lib/io/hardware_i2c.hpp
mifrandir: we don't like transitive includes because they can make things intransparent so this is fine imo

lib/utils/dummy_i2c_sensor.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <sensors/i2c_sensors.hpp>
+
+namespace hyped::utils {
+
+class DummyI2cSensor : public sensors::II2cMuxSensor<std::uint8_t> {
```
SnickeyX, line 7: if this is supposed to be able to imitate any i2c sensor, an `int16_t` might be more suitable here. (_unless we do templating for this as well_  (❛‿❛✿̶̥̥) )

SnickeyX (changes requested):

lib/utils/dummy_i2c_sensor.hpp
mifrandir: this is literally the most trivial sensor i could conjure in order to instantiate a mux. we can make this class less trivial in the future but it's literally just so that i can instantiate the mux template and catch some errors in the generated code.

SnickeyX: okay fair lmao

SnickeyX (approved):
interfaces ftw!

PR #59, opened 2022-11-28
SNS - Fixing Parameter Order

SnickeyX:
pr to fix order to parameters from most general to more specific

### DIFF
--- a/lib/sensors/mux.hpp
+++ b/lib/sensors/mux.hpp
@@ -18,10 +18,10 @@ static constexpr core::Float kFailureThreshold
 template<class T, std::uint8_t N>
 class Mux {
  public:
-  Mux(io::II2c &i2c,
+  Mux(core::ILogger &log,
+      io::II2c &i2c,
       const std::uint8_t mux_address,
-      std::array<std::unique_ptr<II2cMuxSensor<T>>, N> &sensors,
-      core::ILogger &log);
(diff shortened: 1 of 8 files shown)


mifrandir (changes requested):
While we're at it: make sure the class members are in the right order, too. Also, I think we've agreed to call it `logger_` now. This wouldn't be a terrible place to fix that.

lib/sensors/mux.hpp
```diff
@@ -18,33 +18,33 @@ static constexpr core::Float kFailureThreshold
...
-      std::array<std::unique_ptr<II2cMuxSensor<T>>, N> &sensors,
-      core::ILogger &log);
+      std::array<std::unique_ptr<II2cMuxSensor<T>>, N> &sensors);
   ~Mux();
 
   std::optional<std::array<T, N>> readAllChannels();
 
  private:
   core::Result selectChannel(const std::uint8_t channel);
   core::Result closeAllChannels();
```
SnickeyX, line 31: actually having these here looks a bit ugly - do you think these should go below the variables with a space in between? (or something else) @mifrandir

mifrandir: Just put a space between class members and private methods? Wouldn't do anything else tbh.

mifrandir (approved):

kshxtij (approved):

PR #60, opened 2022-12-19
NAV - imu displacement for crosschecker

maxguy2001:
For getting (kalman filtered) imu acceleration into a format that can be used to directly compare against wheel encoder displacement


lib/navigation/imu_displacement.hpp
```diff
@@ -0,0 +1,50 @@
+#pragma once
+
+#include <array>
```
BenET01, line 3: not used...

lib/navigation/imu_displacement.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <cstdint>
+#include <optional>
```
BenET01, line 5: are you using this?

lib/navigation/imu_displacement.cpp
```diff
@@ -0,0 +1,50 @@
...
+      previous_timestamp_micros_(std::chrono::duration_cast<std::chrono::microseconds>(
+                                   std::chrono::system_clock::now().time_since_epoch())
+                                   .count())
+{
+}
+
+void ImuDisplacement::updateImuDisplacement(const core::Float imu_acceleration)
+{
+  const core::TimePoint time_now = std::chrono::system_clock::now();
+  const std::int64_t time_now_micros
```
BenET01, line 21: uint64_t. Time cannot be negative. Everywhere else as well

lib/navigation/imu_displacement.cpp
```diff
@@ -0,0 +1,50 @@
...
+    = std::chrono::duration_cast<std::chrono::microseconds>(time_now.time_since_epoch()).count();
+  const std::int64_t time_elapsed_seconds
+    = static_cast<double>((time_now_micros - previous_timestamp_micros_) / 1'000'000);
+
+  // from equation v=u+at
+  const core::Float velocity_estimate = imu_velocity_ + (imu_acceleration * time_elapsed_seconds);
+
+  // from equation s = ut + 0.5*a*(t^2)
+  imu_displacement_ = (imu_velocity_ * time_elapsed_seconds)
+                      + (0.5 * imu_acceleration * std::pow(time_elapsed_seconds, 2));
```
mifrandir, line 31: I think `std::pow` isn't great in the small integer case. Just `time_elapsed_seconds * time_elapsed_seconds`, even though it's ugly.

lib/navigation/imu_displacement.cpp
```diff
@@ -0,0 +1,50 @@
...
+                                   std::chrono::system_clock::now().time_since_epoch())
+                                   .count())
+{
+}
+
+void ImuDisplacement::updateImuDisplacement(const core::Float imu_acceleration)
+{
+  const core::TimePoint time_now = std::chrono::system_clock::now();
+  const std::int64_t time_now_micros
+    = std::chrono::duration_cast<std::chrono::microseconds>(time_now.time_since_epoch()).count();
```
mifrandir, line 22: have a look at `core::ITimeSource` :)

don't make your code untestable!

lib/navigation/imu_displacement.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+ private:
+  // imu estimate of acceleration
+  core::Float imu_displacement_;
+
+  // imu estimate of velocity
+  core::Float imu_velocity_;
+
+  // timestamp of the last measured acceleration value
+  std::int64_t previous_timestamp_micros_;
```
mifrandir, line 47: `core::TimePoint`

if there is utility that we're missing, go implement it for everybody!

lib/navigation/imu_displacement.cpp
```diff
@@ -0,0 +1,50 @@
...
+  const std::int64_t time_elapsed_seconds
+    = static_cast<double>((time_now_micros - previous_timestamp_micros_) / 1'000'000);
```
mifrandir, line 24: `core::Timer`

mifrandir (changes requested):

lib/navigation/imu_displacement.hpp
maxguy2001: Not currently but will probably be implementing some sort of fail state

lib/navigation/imu_displacement.hpp
maxguy2001: This is just a big integer, not a timestamp?

lib/navigation/imu_displacement.cpp
maxguy2001: So make a class which inherits from `ITimeSource` and implement the functionality I want in there so that we can make a similar class that can be used to create tests but with the same basic structure?

lib/navigation/imu_displacement.cpp
maxguy2001: It's the time between measurements, not the time a task takes. I'm not sure how `core::Timer` will do what I'm looking for here (unless I should just implement this functionality in `core::Timer?`)

lib/navigation/imu_displacement.cpp
mifrandir: not quite. just pass have your functions take an instance of `core::ITimeSource`

then the caller can make it e.g. a `core::WallClock` (for production) or `utils::ManualTime` (for testing)

lib/navigation/imu_displacement.hpp
mifrandir: it looks like it wants to be a timestamp, though?

lib/navigation/imu_displacement.cpp
mifrandir: the latter i think

make it use the `core::Duration` and `core::TimePoint` types and take a `core::ITimeSource`

lib/core/timer.cpp
```diff
@@ -14,4 +14,11 @@ Duration Timer::measure_execution_time(const std::function<void(void)> task)
   return after - before;
 }
 
+Duration Timer::measure_lapsed_time(const TimePoint previous_timepoint)
+{
+  const TimePoint time_now = time_.now();
+  const Duration duration  = previous_timepoint - time_now;
+  return duration;
```
TomLonergan03, line 21: is it more readable to split this versus `return previous_timepoint - time_.now();`?

TomLonergan03: i could see an arguement either way tbh

TomLonergan03 (changes requested):

lib/core/timer.cpp
maxguy2001: I think I did it this way to both make it more readable and make it clear that `time_now` was explicitly a `TimePoint`

TomLonergan03 (approved):

lib/core/timer.cpp
```diff
@@ -14,4 +14,11 @@ Duration Timer::measure_execution_time(const std::function<void(void)> task)
   return after - before;
 }
 
+Duration Timer::measure_lapsed_time(const TimePoint previous_timepoint)
```
mifrandir, line 17: naming style

mifrandir: i think this whole paradigm is wrong, by the way

i fixed the timer class in the navigation benchmarker PR but never got around to finishing and merging that

maybe you can extract the stuff from there? it should give you the utility that you need

lib/navigation/imu_displacement.cpp
```diff
@@ -0,0 +1,55 @@
...
+void ImuDisplacement::initialiseTimePoint(const core::TimePoint initial_timepoint)
+{
+  // TODOLater: Implement this initialisation in main algorithm (some sort of startup process?,
+  // Potentailly in initialisation state depending on how dependent navigation wants to be on state)
+  previous_timestamp_ = initial_timepoint;
+}
+
+core::Float ImuDisplacement::getImuDisplacement()
+{
+  // TODOLater (/elsewhere?): error handling
```
mifrandir, line 45: this, too

lib/navigation/imu_displacement.cpp
```diff
@@ -0,0 +1,55 @@
...
+  // from equation v=u+at
+  const core::Float velocity_estimate = imu_velocity_ + (imu_acceleration * time_elapsed_seconds);
+
+  // from equation s = ut + 0.5*a*(t^2)
+  imu_displacement_ = (imu_velocity_ * time_elapsed_seconds)
+                      + (0.5 * imu_acceleration * time_elapsed_seconds * time_elapsed_seconds);
+
+  // update class members
+  imu_velocity_       = velocity_estimate;
+  previous_timestamp_ = time_now;
```
mifrandir, line 33: lots of whitespace

i'd like to see more `auto`s

lib/navigation/imu_displacement.cpp
```diff
@@ -0,0 +1,55 @@
...
+{
+}
+
+void ImuDisplacement::updateImuDisplacement(const core::Float imu_acceleration)
+{
+  const core::TimePoint time_now = time_.now();
+  core::Timer timer_(time_);
+  const core::Duration time_elapsed = timer_.measure_lapsed_time(previous_timestamp_);
+  const core::Float time_elapsed_seconds
+    = std::chrono::duration_cast<std::chrono::microseconds>(time_elapsed).count() / 1'000'000;
```
mifrandir, line 22: i believe i added a helper for this in `time.hpp` too, possibly on an unmerged branch (see above)

if not, make one

we don't want this ugly chrono syntax anywhere outside `core`

lib/navigation/imu_displacement.hpp
```diff
@@ -0,0 +1,60 @@
...
+  core::Float imu_displacement_;
+
+  // imu estimate of velocity
+  core::Float imu_velocity_;
+
+  // timestamp of the last measured acceleration value
+  core::TimePoint previous_timestamp_;
+
+  // time source
+  const core::ITimeSource &time_;
```
mifrandir, line 57: all of these descriptions are pointless

lib/navigation/imu_displacement.hpp
```diff
@@ -0,0 +1,60 @@
...
+   *
+   * @return imu estimate of acceleration
+   */
+  core::Float getImuDisplacement();
+
+  /**
+   * @brief Get the Imu Velocity object
+   *
+   * @return imu estimate of velocity
+   */
```
mifrandir, line 43: obvious doc, same above

lib/navigation/imu_displacement.cpp
```diff
@@ -0,0 +1,55 @@
...
+  // from equation s = ut + 0.5*a*(t^2)
+  imu_displacement_ = (imu_velocity_ * time_elapsed_seconds)
+                      + (0.5 * imu_acceleration * time_elapsed_seconds * time_elapsed_seconds);
+
+  // update class members
+  imu_velocity_       = velocity_estimate;
+  previous_timestamp_ = time_now;
+}
+
+void ImuDisplacement::initialiseTimePoint(const core::TimePoint initial_timepoint)
```
mifrandir, line 36: put this in the constructor

mifrandir (changes requested):
1. don't confuse `TODO` with `TODOLater`; adding error handling is definitely something that should happen *before* we merge the changes
2. don't add pointless comments; only comment when there is something non-trivial; otherwise nobody will (a) read and (b) update the comments making them no better than noise

lib/core/timer.cpp
maxguy2001: I see what you've done in the benchmark pr but once the accelerometer setup is working, this will probably have to change to use the actual timepoints given with input data, making your other implementation of timer with the reset function not  suited to this. Think I'll change how this works to your implementation for now with a TODOLater for fixing this for final functionality.

maxguy2001: I've changed how this works after rethinking what information will actually be passed to the class and its methods. Timer utility wasn't there already but was probably quite a lot simpler than the previous implementation.

lib/navigation/imu_displacement.cpp
```diff
@@ -6,20 +6,20 @@
 
 namespace hyped::navigation {
 
-ImuDisplacement::ImuDisplacement(const core::ITimeSource &time)
+ImuDisplacement::ImuDisplacement(const core::ITimeSource &time, const core::TimePoint initial_time)
```
mifrandir, line 9: why do we want to have this as a parameter rather than taking `time.now()`?

maxguy2001: Because it is the timepoint of a measurement (class will take the timestamp of the first accelerometer reading as a constructor argument). The time difference between these should be very minimal but when we're working at nanosecond precision and process accelerometer data as heavily as we do, we probably shouldn't ignore this difference?

TomLonergan03 (approved):

kshxtij (approved):

lib/navigation/imu_displacement.cpp
mifrandir: surely this wants to be constructed during initialisation, i.e. when we're not moving anyways?

in that case, we're talking about seconds before we actually end up moving, no? if this is absolutely necessary, which it may very well be, find a way to make this intention clear in the types (not in comments!)

lib/navigation/imu_displacement.cpp
```diff
@@ -0,0 +1,44 @@
+#include "imu_displacement.hpp"
+
+#include <cmath>
+
+#include <chrono>
```
mifrandir, line 5: does clang-format group the headers like this? i'd be surprised if it adds a space between these two; if that's the case we should fix the `.clang_format` to not do that

lib/core/timer.hpp
```diff
@@ -9,10 +10,17 @@ namespace hyped::core {
...
  public:
   Timer(const ITimeSource &time);
-  Duration measure_execution_time(const std::function<void(void)> task);
+
+  Duration measureExecutionTime(const std::function<void(void)> task);
+
+  Duration measureElapsedTime(const TimePoint current_timepoint,
+                              const TimePoint previous_timepoint);
+
+  Float elapsedTimeInSeconds(const Duration time_elapsed);
```
mifrandir, line 19: i feel like this wants a warning along the lines of "don't use when precision matters"

lib/navigation/imu_displacement.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include <cstdint>
+#include <optional>
+
+#include "core/time.hpp"
+#include "core/timer.hpp"
+#include "core/types.hpp"
+
+namespace hyped::navigation {
+
+class ImuDisplacement {
```
mifrandir, line 12: I thought we were going with Accelerometer rather than Imu?

lib/navigation/imu_displacement.hpp
```diff
@@ -0,0 +1,37 @@
...
+  ImuDisplacement(const core::ITimeSource &time, const core::TimePoint initial_time);
+  /**
+   * @brief update the imu estimate for both displacement and velocity. Displacement is
+   * used for cross checking agreement with wheel encooders, velocity is used for us to
+   * estimate velocity in the case we have no direct velocity sensors.
+   *
+   * @param imu_acceleration kalman filtered single value for acceleration
+   * @param imu_timestamp time at which easurement was taken
+   */
+  void updateImuDisplacement(const core::Float imu_acceleration,
```
mifrandir, line 23: just `update` the rest is in the name

`ImuDisplacement::updateImuDisplacement`...

lib/navigation/imu_displacement.hpp
```diff
@@ -0,0 +1,37 @@
...
+                             const core::TimePoint imu_timestamp);
+
+  core::Float getImuDisplacement();
+
+  core::Float getImuVelocity();
```
mifrandir, line 28: The class is called displacement... why are we getting a velocity here? also, put a `const` after

lib/navigation/imu_displacement.hpp
mifrandir: how about `AccelerometerTrajectoryEstimator`?

lib/navigation/imu_displacement.hpp
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * @brief update the imu estimate for both displacement and velocity. Displacement is
+   * used for cross checking agreement with wheel encooders, velocity is used for us to
+   * estimate velocity in the case we have no direct velocity sensors.
+   *
+   * @param imu_acceleration kalman filtered single value for acceleration
+   * @param imu_timestamp time at which easurement was taken
+   */
+  void updateImuDisplacement(const core::Float imu_acceleration,
+                             const core::TimePoint imu_timestamp);
```
mifrandir, line 24: I feel like we want something like the old `DataPoint` here; the sensor should timestamp things in the first instance and from then on the timestamp should be carried around without any possibility for error; this is quite important for the correctness of the whole subsystem.

lib/navigation/imu_displacement.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+  core::Float getImuDisplacement();
+
+  core::Float getImuVelocity();
+
+ private:
+  core::Float imu_displacement_;
```
mifrandir, line 31: `displacement_estimate_`?

mifrandir (changes requested):
naming and interfaces could do with some improvement

lib/navigation/imu_displacement.hpp
maxguy2001: I just wanted to make very clear that this is the imu estimate for displacement since we will also have encoder and keyence estimates

lib/navigation/imu_displacement.hpp
maxguy2001: Will add a todolater. Will fox this once we have the data formats implemeted from sensors.

lib/navigation/imu_displacement.cpp
maxguy2001: It does. I'll have a look into it later.

lib/core/timer.hpp
maxguy2001: Precision probably does matter though. Would a double be accurate enough?

lib/navigation/imu_displacement.cpp
maxguy2001: I guess it depends on how/when this is called. I'll have to consider when this is constructed and how this will fit in with accelerometer calibration.

lib/navigation/imu_displacement.hpp
mifrandir: but you're inside the IMU related class? I also want to emphasise that this is an estimate.

lib/navigation/imu_displacement.hpp
mifrandir: Please get to that asap, shouldn't take long.

lib/navigation/imu_displacement.hpp
```diff
@@ -20,12 +21,11 @@ class ImuDisplacement {
    * @param imu_acceleration kalman filtered single value for acceleration
    * @param imu_timestamp time at which easurement was taken
    */
-  void updateImuDisplacement(const core::Float imu_acceleration,
-                             const core::TimePoint imu_timestamp);
+  void update(const core::Float imu_acceleration, const core::TimePoint imu_timestamp);
 
-  core::Float getImuDisplacement();
+  core::Float getAccelerometerDisplacement() const;
```
mifrandir, line 26: i don't think `accelerometer_trajectory_estimator.getAccelerometerDisplacement()` reads well?

Surely `accelerometer_trajectory_estimator.getDisplacementEstimate()` is better?

mifrandir: Or even just `.getDisplacement()` if you don't like the estimate bit

lib/core/timer.hpp
mifrandir: If you care about precision then you should return `Duration` as much as possible, that's nano-second precision

mifrandir: i think this function is still fine for testing purposes and that sort of stuff; it just shouldn't be used in critical environments

lib/core/timer.cpp
```diff
@@ -6,12 +6,26 @@ Timer::Timer(const ITimeSource &time) : time_(time)
...
+Duration Timer::measureElapsedTime(const TimePoint current_timepoint,
+                                   const TimePoint previous_timepoint)
+{
+  const Duration duration = previous_timepoint - previous_timepoint;
+  return duration;
+}
+
+Float Timer::elapsedTimeInSeconds(const Duration time_elapsed)
+{
+  std::uint64_t nanoseconds_elapsed
```
mifrandir, line 26: `const`

lib/core/timer.cpp
```diff
@@ -6,12 +6,26 @@ Timer::Timer(const ITimeSource &time) : time_(time)
...
-Duration Timer::measure_execution_time(const std::function<void(void)> task)
+Duration Timer::measureExecutionTime(const std::function<void(void)> task)
 {
   const auto before = time_.now();
   task();
   const auto after = time_.now();
   return after - before;
 }
 
+Duration Timer::measureElapsedTime(const TimePoint current_timepoint,
```
mifrandir, line 17: I don't like this function; I think it should be 

```cpp
Duration Timer::elapsed() { /* ... */ }
```

mifrandir: basically, just copy the files from [here](https://github.com/Hyp-ed/hyped-2023/blob/navigation/navigator-benchmarker/lib/core/timer.cpp)

mifrandir (changes requested):

lib/core/timer.hpp
maxguy2001: Yes but I need to multiply a duration value by a Float and we don't want ugly stuff like duration casts outside of core. What's the best way to do this?

mifrandir: https://en.cppreference.com/w/cpp/chrono/duration/operator_arith4

lib/navigation/accelerometer_trajectory.hpp
```diff
@@ -0,0 +1,37 @@
...
+ public:
+  AccelerometerTrajectoryEstimator(const core::ITimeSource &time,
+                                   const core::TimePoint initial_time);
+  /**
+   * @brief update the imu estimate for both displacement and velocity. Displacement is
+   * used for cross checking agreement with wheel encooders, velocity is used for us to
+   * estimate velocity in the case we have no direct velocity sensors.
+   *
+   * @param imu_acceleration kalman filtered single value for acceleration
+   * @param imu_timestamp time at which easurement was taken
```
SnickeyX, line 22: spelling

lib/navigation/accelerometer_trajectory.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include "core/time.hpp"
+#include "core/timer.hpp"
+#include "core/types.hpp"
+
+namespace hyped::navigation {
+
+class AccelerometerTrajectoryEstimator {
+ public:
+  AccelerometerTrajectoryEstimator(const core::ITimeSource &time,
+                                   const core::TimePoint initial_time);
```
SnickeyX, line 15: are we not going to be logging anything here? `update` could surely have a kDebug log or two

lib/navigation/accelerometer_trajectory.cpp
```diff
@@ -0,0 +1,48 @@
...
+core::Float AccelerometerTrajectoryEstimator::getDisplacementEstimate() const
+{
+  return displacement_estimate_;
+}
+
+core::Float AccelerometerTrajectoryEstimator::getVelocityEstimate() const
+{
+  return velocity_estimate_;
+}
+}  // namespace hyped::navigation
```
SnickeyX, line 48: a space just before the last brace

SnickeyX (changes requested):

lib/navigation/accelerometer_trajectory.hpp
maxguy2001: What would I need to log? This is just going to be done automatically on receiving new accelerometer data and all the maths is numerically stable so shouldn't really have issues with it. I can put a couple debug logs in if it's better but I didn't feel it was necessary?

lib/navigation/accelerometer_trajectory.cpp
```diff
@@ -0,0 +1,49 @@
...
+      previous_timestamp_(initial_time),
+      displacement_estimate_(0),
+      velocity_estimate_(0)
+{
+}
+
+// TODOLater: change arguments to instead take some sort of "datapoint" struct with acc_val and
+// timestamp instead
+void AccelerometerTrajectoryEstimator::update(const core::Float imu_acceleration,
+                                              const core::TimePoint imu_timestamp)
```
mifrandir, line 21: don't like the argument names

lib/navigation/accelerometer_trajectory.cpp
```diff
@@ -0,0 +1,49 @@
...
+{
+  core::Timer timer_(time_);
+  const auto time_elapsed                = timer_.elapsed(imu_timestamp, previous_timestamp_);
+  const core::Float time_elapsed_seconds = timer_.elapsedTimeInSeconds(time_elapsed);
```
mifrandir, line 25: surely this class should have a private member of type `core::Timer` on which you call `elapsed` and then `reset` whenever you receive a new value

lib/navigation/accelerometer_trajectory.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+  // from equation v=u+at
+  const core::Float velocity_estimate
+    = velocity_estimate_ + (imu_acceleration * time_elapsed_seconds);
+
+  // from equation s = ut + 0.5*a*(t^2)
+  displacement_estimate_ = (velocity_estimate_ * time_elapsed_seconds)
+                           + (0.5 * imu_acceleration * time_elapsed_seconds * time_elapsed_seconds);
+
```
mifrandir, line 34: whitespace

mifrandir (changes requested):

lib/navigation/accelerometer_trajectory.cpp
maxguy2001: I probably could do but wouldn't that introduce more error into the system since we would now have inconsistencies between timer values and the timestamps of the incoming data (reset uses time.now() when we would be using accelerometer data collected at a particular timepoint)? Could this be something we either suggest modifying for the honors project or play about with in the navigator benchmarker on a separate branch?

SnickeyX (approved):

lib/core/timer.hpp
```diff
@@ -9,10 +10,16 @@ namespace hyped::core {
...
+
+  Duration measureExecutionTime(const std::function<void(void)> task);
+
+  Duration elapsed(const TimePoint current_timepoint, const TimePoint previous_timepoint);
+
+  Float elapsedTimeInSeconds(const Duration time_elapsed);
 
  private:
   const ITimeSource &time_;
+  static constexpr std::uint32_t kOneSecond = 1'000'000'000;
```
TomLonergan03, line 22: don't need `static` i believe

lib/navigation/accelerometer_trajectory.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// TODOLater: change arguments to instead take some sort of "datapoint" struct with acc_val and
+// timestamp instead
+void AccelerometerTrajectoryEstimator::update(const core::Float acceleration,
+                                              const core::TimePoint timestamp)
+{
+  core::Timer timer_(time_);
+  const auto time_elapsed                = timer_.elapsed(timestamp, previous_timestamp_);
+  const core::Float time_elapsed_seconds = timer_.elapsedTimeInSeconds(time_elapsed);
```
TomLonergan03, line 25: `auto`

TomLonergan03 (changes requested):

lib/navigation/accelerometer_trajectory.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include <cstdint>
+#include <optional>
+
+#include "core/time.hpp"
+#include "core/timer.hpp"
+#include "core/types.hpp"
```
TomLonergan03, line 8: These 3 get `<>` as they are not relative paths

TomLonergan03 (changes requested):
One last nitpick then I'll approve

BenET01 (approved):

TomLonergan03 (approved):

PR #61, opened 2022-12-25
SNS - PWM

SnickeyX:
Adds PWM


lib/io/pwm.cpp
```diff
@@ -0,0 +1,151 @@
...
+  std::string polarity_address = pwm_address + "polarity";
+  polarity_file_               = open(polarity_address.c_str(), O_WRONLY);
+  if (polarity_file_ < 0) {
+    logger_.log(core::LogLevel::kFatal, "Failed to open PWM polarity file");
+  }
+  // Finally get the file descriptor for the enable file
+  std::string enable_address = pwm_address + "enable";
+  enable_file_               = open(enable_address.c_str(), O_WRONLY);
+  if (enable_file_ < 0) { logger_.log(core::LogLevel::kFatal, "Failed to open PWM enable file"); }
+}
```
mifrandir, line 31: 1. Don't do I/O in the constructor, have an initialise function or something.
2. `const`s
3. Return on error (rather than just logging)

One problem I have with NOT doing I/O in the constructor is that it's easy to forget calling the other function. Maybe we want to change it into a static function that returns `std::optional<Pwm>`?

lib/io/pwm.hpp
```diff
@@ -0,0 +1,48 @@
...
+  ~Pwm();
+
+  // The valid values for duty cycle are 0.0 to 1.0 (0% to 100%)
+  core::Result setDutyCycleByPercentage(const core::Float duty_cycle);
+  // The valid values for time active are 0 to the period (0% to 100%)
+  core::Result setDutyCycleByTime(const std::uint32_t time_active);
+  core::Result setPeriod(const std::uint32_t period);
+  core::Result setPolarity(const Polarity polarity);
+  core::Result setMode(const Mode mode);
+  std::string getPwmFolderName(const PwmOutput pwm_output);
```
mifrandir, line 37: private

lib/io/pwm.cpp
```diff
@@ -0,0 +1,151 @@
...
+      return "pwm-1:1";
+    case PwmOutput::kEHRPwm1A:
+      return "pwm-4:0";
+    case PwmOutput::kEHRPwm1B:
+      return "pwm-4:1";
+    case PwmOutput::kEHRPwm2A:
+      return "pwm-7:0";
+    case PwmOutput::kEHRPwm2B:
+      return "pwm-7:1";
+    default:
```
mifrandir, line 146: remove default case

lib/io/pwm.cpp
```diff
@@ -0,0 +1,151 @@
...
+  return core::Result::kSuccess;
+}
+
+core::Result Pwm::setMode(const Mode mode)
+{
+  if (enable_file_ < 0) {
+    logger_.log(core::LogLevel::kFatal, "Failed to find PWM enable file while setting run mode");
+    return core::Result::kFailure;
+  }
+  std::uint8_t mode_value      = mode == Mode::kStop ? 0 : 1;
```
mifrandir, line 116: `const`

also ternary operator is questionable; i don't think you gain much over just using a simple if/else or a cast from bool to int

lib/io/pwm.cpp
```diff
@@ -0,0 +1,151 @@
...
+  return core::Result::kSuccess;
+}
+
+core::Result Pwm::setPolarity(const Polarity polarity)
+{
+  if (polarity_file_ < 0) {
+    logger_.log(core::LogLevel::kFatal, "Failed to find PWM polarity file while setting polarity");
+    return core::Result::kFailure;
+  }
+  std::uint8_t polarity_value  = polarity == Polarity::kActiveHigh ? 0 : 1;
```
mifrandir, line 100: maybe store the last written value to avoid repeated I/O operations?

mifrandir (changes requested):
looks sensible; didn't read with too much care just yet

lib/io/pwm.cpp
SnickeyX: won't changing it to a static function mean there is only ever one unique pwm object?

SnickeyX: i do agree with a separate initalise function tho, its a waste not knowing if a open threw an error during construction. how about if we a `has_initialized_` value that tells us if the initialize function was called or not, i'll commit it in a bit

kshxtij:
Could you also add commands to the debug cli while you're at it?

lib/io/pwm.cpp
mifrandir: > won't changing it to a static function mean there is only ever one unique pwm object?

not quite; the static function would replace the constructor in the public interface. you could call it as, say, `Pwm::create(logger, pwm_out)`, which would internally construct a `Pwm` object (or fail).

the `static` keyword is a bit overloaded. for class members it simply means that its not part of a particular object but rather of the class. this is similar to Java.

static variables and static functions (outside of classes) are both different to this.

SnickeyX: ah okay makes sense, will make that change then

lib/io/pwm.cpp
```diff
@@ -0,0 +1,214 @@
...
+  if (period_file_ < 0) {
+    logger_.log(core::LogLevel::kFatal, "Failed to find PWM period file while setting period");
+    return core::Result::kFailure;
+  }
+  // Avoid I/O if no change is required
+  if (current_period_ == period) {
+    logger_.log(core::LogLevel::kDebug, "PWM period is already set to %d", period);
+    return core::Result::kSuccess;
+  }
+  current_period_ = period;
```
mifrandir, line 131: i'm not sure about the semantics here, but i believe this should only happen once the io succeeds? otherwise, if the first attempt fails, the second one will be "successful" but the io will never go through

lib/io/pwm.cpp
```diff
@@ -0,0 +1,214 @@
...
+  if (polarity_file_ < 0) {
+    logger_.log(core::LogLevel::kFatal, "Failed to find PWM polarity file while setting polarity");
+    return core::Result::kFailure;
+  }
+  // Avoid I/O if no change is required
+  if (current_polarity_ == polarity) {
+    logger_.log(core::LogLevel::kDebug, "PWM polarity is already set to %d", polarity);
+    return core::Result::kSuccess;
+  }
+  current_polarity_                 = polarity;
```
mifrandir, line 154: as above

lib/io/pwm.cpp
```diff
@@ -0,0 +1,214 @@
...
+    logger_.log(core::LogLevel::kFatal,
+                "Failed to set duty cycle, active time cannot be greater than period");
+    return core::Result::kFailure;
+  }
+  // Avoid I/O if the duty cycle is the same
+  if (time_active == current_time_active_) {
+    logger_.log(core::LogLevel::kDebug, "Duty cycle is the same as before, skipping I/O");
+    return core::Result::kSuccess;
+  }
+  current_time_active_ = time_active;
```
mifrandir, line 108: ...

lib/io/pwm.cpp
```diff
@@ -0,0 +1,214 @@
...
+  if (enable_file_ < 0) {
+    logger_.log(core::LogLevel::kFatal, "Failed to find PWM enable file while setting mode");
+    return core::Result::kFailure;
+  }
+  // Avoid I/O if no change is required
+  if (current_mode_ == mode) {
+    logger_.log(core::LogLevel::kDebug, "PWM mode is already set to %d", mode);
+    return core::Result::kSuccess;
+  }
+  current_mode_                 = mode;
```
mifrandir, line 178: tagging once again because this has bug potential

lib/io/pwm.cpp
```diff
@@ -0,0 +1,214 @@
...
+{
+  Pwm pwm(logger, pwm_module);
+  const auto initialisation_result = pwm.initialise();
+  if (initialisation_result == core::Result::kFailure) {
+    pwm.logger_.log(core::LogLevel::kFatal, "Failed to initialise PWM");
+    return std::nullopt;
+  }
+  pwm.logger_.log(core::LogLevel::kDebug, "Successfully initialised PWM");
+  return pwm;
+}
```
mifrandir, line 18: by the time the object is creates we should have four non-zero file descriptors; this would allow you to make the descriptors `const`

so, here you should open each file individually and then call a constructor that takes the logger, the pwm_module, and the descriptors

lib/io/pwm.cpp
```diff
@@ -0,0 +1,214 @@
...
+    logger_.log(core::LogLevel::kFatal, "Failed to open PWM enable file");
+    return core::Result::kFailure;
+  }
+  logger_.log(core::LogLevel::kDebug, "Successfully initialised PWM");
+  return core::Result::kSuccess;
+}
+
+core::Result Pwm::setDutyCycleByPercentage(const core::Float duty_cycle)
+{
+  // Ensure duty cycle is between 0 and 1 inlcusive
```
mifrandir, line 75: trivial

lib/io/pwm.cpp
```diff
@@ -0,0 +1,214 @@
...
+    return core::Result::kFailure;
+  }
+  // Next get the file descriptor for the polarity file
+  const std::string polarity_address = pwm_address + "polarity";
+  polarity_file_                     = open(polarity_address.c_str(), O_WRONLY);
+  if (polarity_file_ < 0) {
+    logger_.log(core::LogLevel::kFatal, "Failed to open PWM polarity file");
+    return core::Result::kFailure;
+  }
+  // Finally get the file descriptor for the enable file
```
mifrandir, line 62: i don't like these either

mifrandir (changes requested):

lib/io/pwm.cpp
SnickeyX: shoot yeah that's bad, will change

lib/debug/repl.cpp
```diff
@@ -101,6 +101,34 @@ std::optional<std::unique_ptr<Repl>> Repl::fromFile(const std::string &path)
       repl->addI2cCommands(bus.GetUint());
     }
   }
+
```
mifrandir, line 104: whitespace

mifrandir (approved):
didn't really read the repl stuff but y'all will test that anyways

lib/debug/repl.cpp
```diff
@@ -148,7 +126,29 @@ std::optional<std::unique_ptr<Repl>> Repl::fromFile(const std::string &path)
           core::LogLevel::kFatal, "Invalid module id %d in configuration file", module_id);
         return std::nullopt;
       }
-      repl->addPwmCommands(static_cast<io::PwmModule>(module_id));
+      repl->addPwmCommands(module_id);
+    }
+  }
+
```
mifrandir, line 132: ...

kshxtij: I think this is fine to increase readability considering the code for each command looks very similar.

mifrandir: if you are concerned about readability then this should go in a separate function

i think using whitespace to fix shortcomings of your code structure is an anti-pattern

SnickeyX: I've removed the whitespaces, I think making a function for it does sound like a good idea though

kshxtij (approved):

PR #62, opened 2022-12-26
SNS - ADC Error Handling

SnickeyX:
Realised we never checked for fd error during the actual read and only logged during the construction. Also made helper function private.


SnickeyX:
Was wondering if there was a way to handle the fd throwing an error from the constructor itself but the solutions online seemed really ugly imo, so this is my sol atm.

mifrandir (approved):

kshxtij (approved):
LGTM, @SnickeyX resolve the merge conflicts whenever possible and i'll merge it

PR #63, opened 2022-12-26
SNS - SPI

SnickeyX:
TODO:
- [x] Make changes so code compiles 
- [x] Add support for variable SPI parameters
- [x]  Refactor so code fits our standard 
- [x] Add useful comments to make sense of this mess
- [x] Add support for both spi busses
- [x] Separate hardware spi class
- [x] Add commands to debug binary

Complete for now  ᕙ(`▿´)ᕗ

### DIFF
--- a/lib/io/spi.cpp
+++ b/lib/io/spi.cpp
@@ -6,38 +6,6 @@
 #include <sys/ioctl.h>
 #include <sys/mman.h>
 
-#if LINUX
-#include <linux/spi/spidev.h>
-#else
-// Relevant definitions from linux/spi/spidev.h for non-Linux systems
-#define _IOC_SIZEBITS 14
-#define SPI_IOC_MAGIC 'k'
-#define SPI_IOC_WR_MODE _IOW(SPI_IOC_MAGIC, 1, std::uint8_t)
-#define SPI_IOC_WR_MAX_SPEED_HZ _IOW(SPI_IOC_MAGIC, 4, std::uint32_t)
-#define SPI_IOC_WR_LSB_FIRST _IOW(SPI_IOC_MAGIC, 2, std::uint8_t)
-#define SPI_IOC_WR_BITS_PER_WORD _IOW(SPI_IOC_MAGIC, 3, std::uint8_t)
-
-struct spi_ioc_transfer {
-  std::uint64_t tx_buf;
-  std::uint64_t rx_buf;
-  std::uint32_t len;
-  std::uint32_t speed_hz;
-  std::uint16_t delay_usecs;
-  std::uint8_t bits_per_word;
-  std::uint8_t cs_change;
-  std::uint8_t tx_nbits;
-  std::uint8_t rx_nbits;
-  std::uint16_t pad;
-};
-
-#define SPI_MSGSIZE(N)                                                                             \
-  ((((N) * (sizeof(struct spi_ioc_transfer))) < (1 << _IOC_SIZEBITS))                              \
-     ? ((N) * (sizeof(struct spi_ioc_transfer)))                                                   \
-     : 0)
-#define SPI_IOC_MESSAGE(N) _IOW(SPI_IOC_MAGIC, 0, char[SPI_MSGSIZE(N)])
-#define SPI_CS_HIGH 0x04
-#endif  // if LINUX
-
 namespace hyped::io {
 
 // define what the address space of SPI looks like, see
@@ -71,6 +39,7 @@ struct Spi_Registers {
 
 Spi::Spi(core::ILogger &logger)
     : logger_(logger),
+      file_descriptor_(-1),
       spi_registers_(0),
       spi_channel0_registers_(0),
       has_initialised_(Initialised::kFalse)
@@ -85,7 +54,7 @@ Spi::~Spi()
 core::Result Spi::read(std::uint8_t addr, std::uint8_t *rx, std::uint16_t len)
 {
   if (has_initialised_ == Initialised::kFalse) {
-    logger_.log(core::LogLevel::kFatal, "SPI device has not been initialised");
+    logger_.log(core::LogLevel::kFatal, "Failed to read as SPI device has not been initialised");
     return core::Result::kFailure;
   }
   if (file_descriptor_ < 0) {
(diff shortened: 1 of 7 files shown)


kshxtij:
Similar to PWM, could you add debug cli commands?

lib/io/spi.cpp
```diff
@@ -0,0 +1,233 @@
...
+    logger_.log(core::LogLevel::kFatal, "Failed to read from SPI device");
+    return core::Result::kFailure;
+  }
+  logger_.log(core::LogLevel::kDebug, "Successfully read from SPI device");
+  return core::Result::kSuccess;
+}
+
+core::Result Spi::write(std::uint8_t addr, std::uint8_t *tx, std::uint16_t len)
+{
+  if (has_initialised_ == Initialised::kFalse) {
```
mifrandir, line 84: this is why i don't like this flag-style approach; once the caller is in possession of an `Spi` object it should just work.

lib/io/spi.cpp
```diff
@@ -0,0 +1,233 @@
...
+  const auto write_result = ioctl(file_descriptor_, SPI_IOC_MESSAGE(2), message);
+  if (write_result < 0) {
+    logger_.log(core::LogLevel::kFatal, "Failed to write to SPI device");
+    return core::Result::kFailure;
+  }
+  logger_.log(core::LogLevel::kDebug, "Successfully wrote to SPI device");
+  return core::Result::kSuccess;
+}
+
+core::Result Spi::initialiseSpi(const SpiBus bus,
```
mifrandir, line 110: `initialise`?

lib/io/spi.cpp
```diff
@@ -0,0 +1,233 @@
...
+                                const SpiWordSize word_size,
+                                const SpiBitOrder bit_order)
+{
+  // SPI bus only works in kernel mode on Linux, so we neeed to call the provided driver
+  char spi_bus_address[15];
+  if (bus == SpiBus::kSpi0) {
+    strncat(spi_bus_address, "/dev/spidev0.0", 15);
+  } else {
+    strncat(spi_bus_address, "/dev/spidev1.0", 15);
+  }
```
mifrandir, line 121: this wants to be a function

```cpp
static const char *getSpiBusAdress(...) {
  if (...) {
    return "/dev/spidev0.0";
  }
  // ...
}
```

lib/io/spi.cpp
```diff
@@ -0,0 +1,233 @@
...
+  Spi_Channel_Registers channel3;  // 0x168 - 0x17c
+  std::uint32_t xferlevel;         // 0x17c
+};
+
+Spi::Spi(core::ILogger &logger)
+    : logger_(logger),
+      file_descriptor_(-1),
+      spi_registers_(0),
+      spi_channel0_registers_(0),
+      has_initialised_(Initialised::kFalse)
```
mifrandir, line 45: making an enum with values `kTrue` and `kFalse` seems to be missing the point?

lib/io/spi.hpp
```diff
@@ -0,0 +1,121 @@
+#pragma once
+
+#include <cstdint>
+#include <cstring>
+
+#include <core/logger.hpp>
+#include <core/types.hpp>
+
+#if __linux__
```
mifrandir, line 9: is this a system macro?

lib/io/spi.hpp
```diff
@@ -0,0 +1,121 @@
...
+   * @brief Set the clock frequency of the SPI bus
+   * @param clk - clock frequency to be set
+   */
+  core::Result setClock(Clock clk);
+
+  int file_descriptor_;
+  Initialised has_initialised_;
+  Spi_Registers *spi_registers_;
+  Spi_Channel_Registers *spi_channel0_registers_;
+  core::ILogger &logger_;
```
mifrandir, line 118: ordering

mifrandir (changes requested):
i'll have to trust you that the hardware stuff works

lib/io/spi.cpp
SnickeyX: yeah will make the static fucntion to return `Spi` as well

lib/io/spi.hpp
SnickeyX: this is from the old code except the old one would not actually detect linux. since I'm already defining the relevant stuff from spidev, would this even be necessary?

SnickeyX:
> i'll have to trust you that the hardware stuff works

will test but the fundamental functionality hasn't changed from last year and that was working. a bit stumped by a single spi bus having multiple channels on the bbb specifically - gotto ask electronics how that works and if they will need it, otherwise using channel 0 should work just fine.

SnickeyX:
I've omitted full-duplex communcation which would essentially allow for simulataneous read and writes, except we do not have a use for it (just yet anyway). The entire virtual mapping business done before was to allow for super fast async full-duplex communcation (or at least an attempt to do so), but according to the spidev doc [here](https://www.kernel.org/doc/Documentation/spi/spidev#:~:text=To%20make%20a%20full%20duplex%20request%2C%20provide%20both%20rx_buf%20and%20tx_buf%20for%20the%0Asame%20transfer) ioctl calls should suffice provided we give both a tx and rx buffer AND provided the sensor itself supports it (some only support half-duplex). 

For now I think I'm going to leave it at half-duplex calls since I know those will work, and test full-duplex with an arduino if I have too much time. Considering no sensors we are using just yet need SPI anyway, this should be fine.

lib/io/spi.cpp
```diff
@@ -4,45 +4,26 @@
...
+                               const SpiBitOrder bit_order)
+{
+  Spi spi(logger);
+  const auto initialisation_result = spi.initialise(bus, mode, word_size, bit_order);
+  if (initialisation_result == core::Result::kFailure) {
+    spi.logger_.log(core::LogLevel::kFatal, "Failed to initialise SPI");
+    return std::nullopt;
+  }
+  spi.logger_.log(core::LogLevel::kDebug, "Successfully initialised SPI");
+  return spi;
```
mifrandir, line 23: similar to the other pr

mifrandir (changes requested):
getting rid of overly general stuff, i like it.

mifrandir (approved):
again, didn't read the repl stuff; otherwise it looks good

kshxtij (approved):

PR #65, opened 2022-12-29
SNS - UART 

SnickeyX:
TODO:
- [x] Finish read and send functions
- [x] Figuring out building on a Mac
- [x] Add to REPL
- [x] Dummy Implementaion with naive test

Also changed success logs in the REPL to `kDebug` instead of `kInfo` and my bug in adding SPI commands.


kshxtij:
Honestly, I'm not sure what the debug interface for this would look like but worth thinking about that as well

lib/debug/repl.cpp
```diff
@@ -300,12 +322,68 @@ void Repl::addSpiCommands(const std::uint8_t bus)
...
+    identifier << "uart " << static_cast<int>(bus) << " read";
+    uart_read_command.name = identifier.str();
+    std::stringstream description;
+    description << "Read from UART bus " << static_cast<int>(bus);
+    uart_read_command.description = description.str();
+    uart_read_command.handler     = [this, uart, bus]() {
+      std::uint8_t read_buffer[255];
+      const core::Result result = uart->readBytes(read_buffer, 255);
+      if (result == core::Result::kSuccess) {
+        logger_.log(core::LogLevel::kDebug, "UART value from bus %d: %s", bus, read_buffer);
```
mifrandir, line 355: do we want `read_buffer` to be parsed as a string here? surely the UART values are raw binary?

lib/io/uart.cpp
```diff
@@ -0,0 +1,107 @@
...
+  return Uart(logger, file_descriptor);
+}
+
+core::Result Uart::configureFileForOperation(core::ILogger &logger,
+                                             const int file_descriptor,
+                                             const std::uint32_t baud_mask,
+                                             const std::uint32_t bits_per_byte_mask)
+{
+  struct termios tty;
+  // ensuring all bits are initially 0, else any set bit could lead to undefined behavior
```
mifrandir, line 50: not sure this is required

mifrandir (changes requested):

lib/debug/repl.cpp
SnickeyX: we don't have a use case for this atm so I've kept it as bytes, will change if and when we do actually need UART

mifrandir (changes requested):
two more things:

1. make an interface `IUart` in `uart.[ch]pp` and move the implementation to `hardware_uart.[ch]pp`
2. write a dummy implementation and instantiate it in a super basic super dumb test; this can be essentially trivial

lib/debug/repl.cpp
mifrandir: i still disagree but i see your point

kshxtij (approved):

PR #66, opened 2023-01-02
SNS - GPIO (read/write)

SnickeyX:
This PR provides an (alternative) implementation of GPIO using read/write only. Using mmap provides a faster(* in certain scenarios) implementation but I want to test if the difference is significant at all between the two. 

Goal:
To test the real difference in I/O between mmap v read/write during a test run (with our car).

### DIFF
--- a/lib/io/hardware_gpio.cpp
+++ b/lib/io/hardware_gpio.cpp
@@ -66,7 +66,7 @@ core::Result HardwareGpioWriter::writeToPin(const core::DigitalSignal state)
     return core::Result::kFailure;
   }
   // Convert DigitalSignal to a string
-  const uint8_t signal_value = static_cast<uint8_t>(state);
+  const std::uint8_t signal_value = static_cast<std::uint8_t>(state);
   char write_buffer[2];
   snprintf(write_buffer, sizeof(write_buffer), "%d", signal_value);
   // Write the value to the file
--- a/lib/io/hardware_gpio.cpp
+++ b/lib/io/hardware_gpio.cpp
@@ -1,223 +1,36 @@
 #include "hardware_gpio.hpp"
 
-#include <fcntl.h>
-#include <unistd.h>
-
 namespace hyped::io {
-
-HardwareGpioReader::HardwareGpioReader(core::ILogger &log, const int read_file_descriptor)
-    : logger_(log),
-      read_file_descriptor_(read_file_descriptor)
-{
-}
-
-HardwareGpioReader::~HardwareGpioReader()
-{
-  close(read_file_descriptor_);
-}
 
 std::optional<core::DigitalSignal> HardwareGpioReader::readPin()
 {
-  // Read the value from the file
-  char read_buffer[2];
-  const auto offset = lseek(read_file_descriptor_, 0, SEEK_SET);
-  if (offset != 0) {
-    logger_.log(core::LogLevel::kFatal, "Failed to reset file offset");
-    return std::nullopt;
-  }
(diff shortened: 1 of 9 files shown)


lib/io/hardware_gpio.cpp
```diff
@@ -2,13 +2,13 @@
...
 
-std::optional<core::DigitalSignal> HardwareGpioReader::read()
+std::optional<core::DigitalSignal> HardwareGpioReader::readPin()
 {
   // TODOLater: implement
   throw -1;
 }
 
-core::Result HardwareGpioWriter::write(const core::DigitalSignal state)
+core::Result HardwareGpioWriter::writeToPin(const core::DigitalSignal state)
```
mifrandir, line 11: not sure i agree with this rename? this function doesn't take a pin as an argument so why is it called `writeToPin`?

lib/io/hardware_gpio_alt.cpp
```diff
@@ -0,0 +1,223 @@
...
+    logger_.log(core::LogLevel::kFatal, "Failed to initialise GPIO %d", pin);
+    return std::nullopt;
+  }
+  const int read_file_descriptor = getFileDescriptor(pin, Direction::kIn);
+  if (read_file_descriptor < 0) {
+    logger_.log(core::LogLevel::kFatal, "Failed to get file descriptor for GPIO %d", pin);
+    return std::nullopt;
+  }
+  HardwareGpioReader reader(logger_, read_file_descriptor);
+  return std::make_shared<HardwareGpioReader>(reader);
```
mifrandir, line 90: i want to comment on something here, for your understanding; feel free to resolve after reading (RAR)

The "right" way to do this would be `return std::make_shared<...>(logger_, read_file_descriptor);`. Of course, this doesn't work because the constructor is private so that wouldn't work. 

How could we resolve this? Option 1 is to use nested classes and make the reader/write a private class inside the Gpio. Then we could make the constructor public (and within the `HardwareGpio` class). We like to avoid nested classes so we accept the ugliness of this. Option 2 would be to make the constructor public. This usually isn't a problem if we assume that the user understands the design. However, this is a bold assumption.

Besides style, why would the single line way be better? Well, we avoid an unnecessary copy. Then again, these sorts of functions are only called on initialisation (i.e. O(1) times) so we don't really care about performance here.

mifrandir (changes requested):
didn't read in too much detail as it's experimental anyways

lib/io/hardware_gpio.cpp
SnickeyX: `writeState` and `readState`?

SnickeyX: or signal instead of state

lib/io/hardware_gpio_alt.cpp
SnickeyX: i see, do appreciate this btw. not always sure what the "right" way to do things even are

lib/io/hardware_gpio.cpp
mifrandir: but like... why not just `write`? I get that we want to make things descriptive but `brakes_pin_.write(core::DigitalSignal::kHigh)` doesn't look particularly hard to parse, for example?

SnickeyX: because the compiler confuses the linux `write` with the gpio `write` and thinks its a recursive function. not sure how to get around it with c libraries not having a namepsace -> was what I was thinking, just found out that global namespace is a thing

mifrandir (approved):

kshxtij (approved):

PR #69, opened 2023-01-16
STM: State Machine Outline

DylanCavers:


lib/state_machine/state_machine.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <optional>
+
+namespace hyped::state_machine {
+
+StateMachine::StateMachine()
+{
+}
+
+// TODOLater implement
+std::optional<Message> StateMachine::checkTransition()
```
mifrandir, line 13: do we have a better name?

lib/state_machine/state_machine.cpp
```diff
@@ -0,0 +1,38 @@
...
+  /*
+  check if a transition is needed with current data
+  then run transition to move to next state
+  return transition message or no transition
+  */
+  return std::nullopt;
+}
+
+// Transition to next state
+void StateMachine::transition(const Message message)
```
mifrandir, line 24: `handleMessage`?

lib/state_machine/state_machine.hpp
```diff
@@ -0,0 +1,72 @@
...
+  StateMachine() { current_state = State::kIdle; };
+
+  void transition(const Message message);
+
+  std::optional<Message> checkTransition();
+
+  void reset();
+
+  // TODO change transitions to actual transitions
+  std::array<Transition, kNumTransitions> transitions{
```
mifrandir, line 23: I don't agree with the data structure here.

Think about "What do I want to do?". I claim there is exactly one way in which you are going to look things up. The current lookup time is `O(n * m)` where `n` is the number of states and `m` the number of messages; this should be constant. Talk to my IADS tutees, they should know how to make this better.

lib/state_machine/types.hpp
```diff
@@ -0,0 +1,14 @@
...
+
+#include "message.hpp"
+#include "state.hpp"
+
+namespace hyped::state_machine {
+
+struct Transition {
+  State from;
+  State to;
+  Message message;
```
mifrandir, line 11: source, target, message?

something like `lookup(transition.to, transition.message)` doesn't read well, I think.

I am also not convinced we need this struct at all but I'll see.

mifrandir (changes requested):
Remove the `.DS_Store`

lib/state_machine/state_machine.hpp
DylanCavers: Would an unordered map be more appropriate for this?

mifrandir: certainly

you probably want to use a key like `State * Message` but put that in a struct like `Event` and then figure out how to hash it.

kshxtij: For the hashing, i think we can just copy over Boost's hash_combine() and put it in utils::hash.

mifrandir: I mean we could also just actually depend on boost, like approximately any other C++ project :P

mifrandir: But there's another thing: you either need to make a custom hasher and carry it around or invoke some black magic which I unfortunately forgot (it involved defining an empty template inside the `std` namespace).

Ofc you might say "but why not use `std::pair<S,T>` and the answer is because the actual code (outside the types) will be ugly if we do this.

lib/state_machine/state_machine.cpp
```diff
@@ -0,0 +1,34 @@
+#include "state_machine.hpp"
+#include "types.hpp"
+
+#include <optional>
```
maxguy2001, line 4: Already included in header

lib/state_machine/types.hpp
```diff
@@ -0,0 +1,27 @@
...
+  bool operator==(const TransitionKey &key) const
+  {
+    return key.source == source && key.message == message;
+  }
+};
+
+struct TransitionHasher {
+  std::size_t operator()(const TransitionKey &key) const
+  {
+    using std::size_t;
```
maxguy2001, line 21: ?

lib/state_machine/types.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+    return ((int)key.source);
```
maxguy2001, line 23: `static_cast<int>(key.source)`?

maxguy2001 (changes requested):

lib/state_machine/state_machine.cpp
```diff
@@ -0,0 +1,28 @@
+#include "state_machine.hpp"
+#include "types.hpp"
```
TomLonergan03, line 2: already included in header

lib/state_machine/state_machine.hpp
```diff
@@ -0,0 +1,72 @@
+#pragma once
+
+#include "consts.hpp"
+#include "state.hpp"
+#include "types.hpp"
+
+#include <array>
```
TomLonergan03, line 7: unused as far as i can see

lib/state_machine/state_machine.hpp
```diff
@@ -0,0 +1,72 @@
...
+       {{State::kMitigate, Message::mPostEmergency}, State::kPECruising},
+       {{State::kMitigate, Message::mPostEmergency}, State::kPEBraking},
+       {{State::kMitigate, Message::mPostEmergency}, State::kMotorBraking},
+       {{State::kMitigate, Message::mPostEmergency}, State::kFrictionBraking},
+       {{State::kMitigate, Message::mPostEmergency}, State::kFailureBraking},
+       {{State::kPEAccelerating, Message::mResume}, State::kAccelerating},
+       {{State::kPECruising, Message::mResume}, State::kCruising},
+       {{State::kPEBraking, Message::mResume}, State::kNominalBraking}};
+
+  State current_state;
```
TomLonergan03, line 70: `current_state_`, also should probably be private, with a getter function if needed

TomLonergan03 (changes requested):

lib/state_machine/state_machine.hpp
mifrandir: that's the magic i was talking about

lib/utils/hash.hpp
```diff
@@ -0,0 +1,12 @@
+namespace hyped::utils {
+
+inline void hash_combine(std::size_t& seed) { }
+
+template <class T>
+inline void hash_combine(std::size_t & seed, const T& v)
+{
+  std::hash<T> hasher;
+  seed ^= hasher(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+};
```
mifrandir: I don't like this snippet because it looks so arbitrary; can we not just depend on boost? Just require it to be present on the host like we do with eigen and we should be good. @kshxtij

lib/state_machine/types.hpp
```diff
@@ -18,9 +20,10 @@ struct TransitionKey {
 struct TransitionHasher {
```
mifrandir, line 22: We would rather do the thing where we add a hash function in `std::` than use a hasher like this.

lib/state_machine/types.hpp
```diff
@@ -18,9 +20,10 @@ struct TransitionKey {
...
   std::size_t operator()(const TransitionKey &key) const
```
mifrandir, line 23: `TransitionKey` is a bad name; what does it contain? If there's no more meaningful name call it `SourceAndMessage`

mifrandir (changes requested):
remove `lib/.DS_Store`

lib/state_machine/types.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include "message.hpp"
+#include "state.hpp"
+
+#include <cstdint>
+
+#include "utils/hash.hpp"
+
+namespace hyped::state_machine {
+
+struct TransitionKey {
```
mifrandir, line 12: This could just live inside `state_machine::StateMachine`? It's not going to be used anywhere else and it's just for the internals.

I'm not sure whether that's possible, but I believe you can make it private. Might be an issue with the `std::` template, though...

lib/state_machine/types.hpp
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include "message.cpp"
```
maxguy2001, line 3: Include headers, not cpp files

maxguy2001 (changes requested):

test/state_machine/stm_tests.cpp
```diff
@@ -0,0 +1,5 @@
+#include <gtest/gtest.h>
+
+namespace hyped::test {
+
```
mifrandir, line 4: todo?

lib/state_machine/state_machine.hpp
```diff
@@ -0,0 +1,72 @@
...
+ public:
+  StateMachine() { current_state = State::kIdle; };
+
+  void handleMessage(const Message message);
+
+  std::optional<Message> handleData();
+
+  void reset();
+
+  const std::unordered_map<SourceAndMessage, State> transitions
```
maxguy2001, line 22: Should this definitely be public?

maxguy2001 (changes requested):

lib/state_machine/types.hpp
```diff
@@ -0,0 +1,33 @@
...
+  bool operator==(const SourceAndMessage &key) const
+  {
+    return key.source == source && key.message == message;
+  }
+};
+}  // namespace hyped::state_machine
+
+namespace std {
+template<>
+struct hash<hyped::state_machine::SourceAndMessage> {
```
maxguy2001, line 24: redundant `hyped`

maxguy2001: And elsewhere in the struct.

maxguy2001 (changes requested):

lib/state_machine/state_machine.hpp
mifrandir: no!

test/state_machine/stm_tests.cpp
```diff
@@ -0,0 +1,5 @@
+#include <gtest/gtest.h>
+
+namespace hyped::test {
+// TODO transition testing
```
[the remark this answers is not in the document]
mifrandir: nope, `TODO`; let them fix it in this PR

lib/state_machine/state_machine.cpp
```diff
@@ -0,0 +1,27 @@
...
+  return std::nullopt;
+}
+
+// Transition to next state
+void StateMachine::handleMessage(const Message message)
+{
+  current_state = transitions.at({current_state, message});
+}
+
+void StateMachine::reset()
```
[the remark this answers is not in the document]
mifrandir: I agree. In fact I don't like this function at all. You should get rid of it in order to avoid bugs. The user can construct a new state machine if that's what's required.

lib/state_machine/state_machine.hpp
mifrandir: also naming

lib/state_machine/state_machine.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+namespace hyped::state_machine {
+
+class StateMachine {
+ public:
+  StateMachine() { current_state = State::kIdle; };
+
+  void handleMessage(const Message message);
+
+  std::optional<Message> handleData();
```
mifrandir, line 18: what is this function supposed to do? it says handle but there isn't any input?

lib/state_machine/types.hpp
maxguy2001: Then we'll have to change the job to check for redundant hyped since this currently causes it to fail.

mifrandir: @kshxtij :joy:

lib/state_machine/state_machine.hpp
DylanCavers: unsure on the input at the moment, aim is for this function to handle the messages received from other the other nodes and transition appropriately, will try figure something out for this on Saturday.

lib/state_machine/state_machine.hpp
```diff
@@ -17,8 +19,18 @@ class StateMachine {
...
+
  private:
-  const std::unordered_map<SourceAndMessage, State> transitions
+  const std::unordered_map<std::string, Message> string_to_message
+    = {{"mForward", Message::mForward}, {"mFailure", Message::mFailure}};
+
+  const std::unordered_map<Message, std::string> message_to_string
+    = {{Message::mForward, "mForward"}, {Message::mFailure, "mFailure"}};
+
+  const boost::unordered_map<SourceAndMessage, State> transitionMap
```
mifrandir, line 33: naming style

lib/state_machine/state_machine.hpp
```diff
@@ -17,8 +19,18 @@ class StateMachine {
 
   std::optional<Message> handleData();
 
+  Message stringToMessage(const std::string message_name);
+
+  std::string messageToString(const Message message);
```
mifrandir, line 24: references

mifrandir (changes requested):

lib/state_machine/state_machine.hpp
```diff
@@ -19,18 +19,18 @@ class StateMachine {
...
-  std::string messageToString(const Message message);
+  std::string messageToString(const Message &message);
 
  private:
-  const std::unordered_map<std::string, Message> string_to_message
+  const std::unordered_map<std::string, Message> StringMap
     = {{"mForward", Message::mForward}, {"mFailure", Message::mFailure}};
 
-  const std::unordered_map<Message, std::string> message_to_string
+  const std::unordered_map<Message, std::string> MessageMap
```
mifrandir, line 30: What happened here?

This should be `message_to_string_` or `kMessageToString`

similar for the others

lib/state_machine/state_machine.cpp
```diff
@@ -0,0 +1,34 @@
+#include "state_machine.hpp"
+
+namespace hyped::state_machine {
+
+// TODOLater implement
+std::optional<Message> StateMachine::handleData()
```
maxguy2001, line 6: Should this not be `checkTransiton` or similar?

maxguy2001: The name just doesn't seem to describe the functionality of this?

maxguy2001 (changes requested):

lib/state_machine/state_machine.cpp
DylanCavers: https://github.com/Hyp-ed/hyped-2023/pull/69#discussion_r1083266641

lib/state_machine/message.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+namespace hyped::state_machine {
+
+// State class containing all the states
+enum class Message {
+  mForward,
```
TomLonergan03, line 7: what is `mForward`? is there a better name for this or is this placeholder messages?

lib/state_machine/state_machine.cpp
TomLonergan03: I prefer `checkTransition` to `handleData` tbh

lib/state_machine/state_machine.cpp
```diff
@@ -0,0 +1,34 @@
...
+// Transition to next state
+void StateMachine::handleMessage(const Message message)
+{
+  current_state = transition_to_state_.at({current_state, message});
+}
+
+Message StateMachine::stringToMessage(const std::string &message_name)
+{
+  Message message = string_to_message_.at(message_name);
+  return message;
```
TomLonergan03, line 25: `return string_to_message_.at(message_name);`

lib/state_machine/state_machine.cpp
```diff
@@ -0,0 +1,34 @@
...
+}
+
+std::string StateMachine::messageToString(const Message &message)
+{
+  std::string message_name = message_to_string_.at(message);
+  return message_name;
```
TomLonergan03, line 31: `return message_to_string_.at(message);`

lib/state_machine/state_machine.hpp
TomLonergan03: if no progress has happened with this, add a TODOLater

lib/state_machine/message.hpp
TomLonergan03: looking at the rest of the PR this would be better as something like `kNextNominalState`

TomLonergan03 (changes requested):

lib/state_machine/state_machine.cpp
```diff
@@ -0,0 +1,32 @@
...
+  /*
+  check if a transition is needed with current data from ros node
+  then run transition to move to next state
+  return transition message or no transition
+  */
+  return std::nullopt;
+}
+
+// Transition to next state
+void StateMachine::handleMessage(const Message message)
```
TomLonergan03, line 17: should this take a reference?

lib/state_machine/state_machine.hpp
```diff
@@ -0,0 +1,61 @@
...
+       {{State::kNominalBraking, Message::kFailureBrake}, State::kFailureBraking},
+       {{State::kMotorBraking, Message::kFailureBrake}, State::kFailureBraking},
+       {{State::kMotorBraking, Message::kFrictionBrake}, State::kFrictionBraking},
+       {{State::kMotorBraking, Message::kFailure}, State::kFailureStopped},
+       {{State::kFrictionBraking, Message::kFailureBrake}, State::kFailureBraking},
+       {{State::kFrictionBraking, Message::kFailure}, State::kFailureStopped},
+       {{State::kFailureBraking, Message::kFailure}, State::kFailureStopped},
+       {{State::kFailureStopped, Message::kFailure}, State::kOff}};
+
+  State current_state;
```
TomLonergan03, line 59: `current_state_`

lib/state_machine/state_machine.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include <optional>
+#include <unordered_map>
+
+#include <boost/unordered_map.hpp>
+
+namespace hyped::state_machine {
+
+class StateMachine {
+ public:
+  StateMachine() { current_state = State::kIdle; };
```
TomLonergan03, line 16: constructor should not be in the header, also initialise `current_state_` in the initialiser list

TomLonergan03 (changes requested):

lib/state_machine/state_machine.cpp
kshxtij: @mifrandir what was the problem with `checkTransition`?

TomLonergan03 (approved):
Looks good

lib/state_machine/message.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+namespace hyped::state_machine {
+
+// State class containing all the states
```
SnickeyX, line 5: State class?

lib/state_machine/state_machine.hpp
```diff
@@ -0,0 +1,59 @@
...
+  StateMachine();
+
+  // TODOLater handle ros messages and transition accordingly
+  std::optional<Message> checkTransition();
+
+  void handleMessage(const Message &message);
+
+  Message stringToMessage(const std::string &message_name);
+
+  std::string messageToString(const Message &message);
```
SnickeyX, line 24: if there aren't any comments, you can leave out the whitespace

lib/state_machine/state_machine.hpp
```diff
@@ -0,0 +1,59 @@
...
+       {{State::kMotorBraking, Message::kPreFrictionBrake}, State::kPreFrictionBraking},
+       {{State::kPreFrictionBraking, Message::kFrictionBrake}, State::kFrictionBraking},
+       {{State::kFrictionBraking, Message::kFailureBrake}, State::kFailureBraking},
+       {{State::kFrictionBraking, Message::kFailure}, State::kFailureStopped},
+       {{State::kFailureBraking, Message::kFailure}, State::kFailureStopped},
+       {{State::kFailureStopped, Message::kFailure}, State::kOff}};
+
+  State current_state_;
+};
+}  // namespace hyped::state_machine
```
SnickeyX, line 59: space before last bracket (pedantic ik)

lib/state_machine/types.hpp
```diff
@@ -0,0 +1,28 @@
...
+};
+
+std::size_t hash_value(SourceAndMessage const &key)
+{
+  std::size_t seed = 0;
+  boost::hash_combine(seed, key.message);
+  boost::hash_combine(seed, key.source);
+  return seed;
+}
+}  // namespace hyped::state_machine
```
SnickeyX, line 28: newline before last bracket

test/state_machine/stm_tests.cpp
SnickeyX, line 4: TODOLater since I'm assuming this is going to be a different PR?

SnickeyX (changes requested):
only formatting changes

lib/state_machine/state.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+namespace hyped::state_machine {
+
+// State class containing all the states
```
SnickeyX, line 5: this comment feels a bit redundant given the context of state machine

SnickeyX (changes requested):
a lil more, will approve right after

SnickeyX (approved):
wheyyy

lib/state_machine/message.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+namespace hyped::state_machine {
+
+// Message class containing messages that prompt transitions
+enum class Message {
+  kNextNominalState,
```
mifrandir, line 7: How are you going to make sure that receiving two of these in a row is not going to advance the state machine too far? Like, how do you avoid one of the controllers sending this message twice or, more realistically, two individual controllers sending this message and the receiver transitioning twice (instead of the intended once).

DylanCavers: Was initially decided on to allow different transition orders to end up in the same state, but I guess this doesn't work as well for the NextNominalState transitions as if two messages are sent from different controller to transition from the same state, they would transition too far and essentially skip a state. This would be needed for failure states to ensure that no matter the order of received messages the failure transition is always made. So might be better to add separate messages for all the non-failure transitions, but keep the general transitions for failure states?

mifrandir: That may be a good idea, yes.

kshxtij (approved):

PR #71, opened 2023-01-19
SNS - Dummy Adc

SnickeyX:
Adding missing dummy adc with trivial construction test.
Again, adding this in preparation for the completion of the software-only I/O tests during the HYPED™ hackathon.

### DIFF
--- a/lib/io/hardware_i2c.cpp
+++ b/lib/io/hardware_i2c.cpp
@@ -17,7 +17,7 @@ std::optional<HardwareI2c> HardwareI2c::create(core::ILogger &logger,
                                                const std::uint8_t bus_address)
 {
   char path[13];  // up to "/dev/i2c-2"
-  sprintf(path, "/dev/i2c-%d", bus_address);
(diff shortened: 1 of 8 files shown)


lib/io/hardware_i2c.cpp
```diff
@@ -17,7 +17,7 @@ std::optional<HardwareI2c> HardwareI2c::create(core::ILogger &logger,
                                                const std::uint8_t bus_address)
 {
   char path[13];  // up to "/dev/i2c-2"
-  sprintf(path, "/dev/i2c-%d", bus_address);
+  snprintf(path, sizeof(path), "/dev/i2c-%d", bus_address);
```
mifrandir, line 20: correct me if i'm wrong, but isn't `sizeof(path)` just the size of the pointer? or has this got to do with the C-array not having decayed just yet?

SnickeyX: yep its exactly as you suspect

kshxtij (approved):

PR #74, opened 2023-01-23
MTC - CAN messages from JSON

TomLonergan03:
Store all predefined messages in a JSON file

- [x] JSON file format
- [x] Data structure for messages
- [x] Parser


CMakeLists.txt
```diff
@@ -12,6 +12,17 @@ set(project HYPED)
...
+include(FetchContent)
+FetchContent_Declare(
+  rapidjson
+  GIT_REPOSITORY https://github.com/Tencent/rapidjson.git
+  GIT_TAG        master
+)
+FetchContent_GetProperties(rapidjson)
+if(NOT rapidjson_POPULATED)
+    FetchContent_Populate(rapidjson)
+endif()
```
mifrandir, line 24: I feel like we prefer this being locally installed now? It's what we do with eigen. Most Linux distros will have a package for rapidjson and otherwise you can install it manually.

@kshxtij ?

lib/motors/messages.json
```diff
@@ -0,0 +1,22 @@
+{
+    "config_messages": [
+        {
+            "example_config_message": {
+                "id": 123,
```
mifrandir, line 5: is this going to be hex or dec?

lib/motors/controller.cpp
```diff
@@ -1,9 +1,98 @@
...
+  }
+  const auto messages = document["messages"].GetObject();
+  for (auto &message : messages) {
+    const std::optional<core::CanFrame> new_message = parseJsonCanFrame(message.value.GetObject());
+    if (!new_message) {
+      logger_.log(
+        core::LogLevel::kFatal, "Invalid CAN frame in JSON message file at path %s", path.c_str());
+      return core::Result::kFailure;
+    }
+    auto x = new_message.value();
```
mifrandir, line 58: this line seems a bit pointless? `x` doesn't tell me what it is, and its not constant

juts do `*new_message` below?

lib/motors/controller.hpp
```diff
@@ -1,19 +1,35 @@
...
 
 namespace hyped::motors {
 enum class ControllerStatus { kControllerTemperatureExceeded, kUnrecoverableWarning, kNominal };
 
 class Controller {
  public:
   Controller(core::ILogger &logger);
+  core::Result parseMessageFile(const std::string &path);
   void processErrorMessage(const std::uint16_t error_code);
   ControllerStatus processWarningMessage(const std::uint8_t warning_code);
```
mifrandir, line 24: I'm a little bit unsure about what the purpose of this class and in particular the message file is, but based on my beliefs here's what I think the interface should be:

A constructor that takes the "list" (or other data structure) of configuration messages, a function (we've started calling them `create` so let's use that name) that parses the message file, collects the messages, constructs the object, and possibly sends the configuration messages. 

What I'm getting at: on one hand we don't want constructors to do I/O because it can fail and leave the object in an undefined state but on the other hand we don't want these magic function that the user has to remember to call (like `parseMessageFile`) in order to make the controller useful at all. There won't be more than one message file I believe, so only passing one should be fine.

lib/motors/controller.hpp
```diff
@@ -1,19 +1,35 @@
...
 
  private:
+  std::optional<core::CanFrame> parseJsonCanFrame(
+    rapidjson::GenericObject<false, rapidjson::Value> message);
   core::ILogger &logger_;
+  // TODO replace core::CanFrame with io::CanFrame once merged
+  std::unordered_map<std::string, core::CanFrame> messages_;
+  std::vector<core::CanFrame> configuration_messages_;
```
mifrandir, line 32: If you implement the above, both of these should end up being `const`

mifrandir (changes requested):

lib/motors/messages.json
TomLonergan03: json doesn't support hex, i think we should use hex so the config messages are identical to those in the manual.  therefore store hex as a string and parse into a value maybe?

lib/motors/controller.hpp
TomLonergan03: yea makes sense,

the message file exists so that we can change what messages get sent to the controller easily, and as EHW will lock our codebase at some point before the competition we can still update for example controller current limits or PID gains or similar

lib/motors/messages.json
TomLonergan03: it is dec at the moment though

mifrandir: hm, thats very ugly

dont know what a good solution is, having strings there sounds weird too

lib/motors/controller.cpp
```diff
@@ -1,9 +1,105 @@
 #include "controller.hpp"
 
+#include <fstream>
+
 namespace hyped::motors {
 
-Controller::Controller(core::ILogger &logger) : logger_(logger)
+std::optional<Controller> Controller::create(core::ILogger &logger, std::string &message_file_path)
```
maxguy2001, line 7: `const`

lib/motors/controller.cpp
```diff
@@ -1,9 +1,105 @@
...
 {
+  std::ifstream input_stream(message_file_path);
+  if (!input_stream.is_open()) {
+    logger.log(core::LogLevel::kFatal, "Failed to open file %s", message_file_path.c_str());
+    return std::nullopt;
+  }
+  rapidjson::IStreamWrapper input_stream_wrapper(input_stream);
+  rapidjson::Document document;
```
maxguy2001, line 15: `const`s?

lib/motors/controller.cpp
```diff
@@ -1,9 +1,105 @@
...
+    return std::nullopt;
+  }
+  if (!message.HasMember("data")) {
+    logger.log(core::LogLevel::kFatal,
+               "Missing required field 'data' in message in CAN message file");
+    return std::nullopt;
+  }
+  core::CanFrame new_message;
+  new_message.can_id  = message["id"].GetInt();
+  new_message.can_dlc = 8;
```
maxguy2001, line 96: Why 8? Is this an obvious thing or does it need to be named and defined somewhere?

maxguy2001 (changes requested):

lib/motors/controller.cpp
TomLonergan03: this is the number of bytes in the message, is always 8 as that's the non-extended CAN standard

lib/motors/controller.cpp
TomLonergan03: the `IStreamWrapper` and `Document` can't be `const`, `ParseResult` can be

lib/motors/controller.cpp
maxguy2001: If it's a constant number that has a meaning then should we either give it a name and store it somewhere or put a comment explaining where it comes from so that future hyped members don't have to deal with a load of "magic numbers" which they don't know the purpose of?

lib/motors/controller.cpp
```diff
@@ -1,9 +1,106 @@
...
+      return std::nullopt;
+    }
+    controller_messages.emplace(message.name.GetString(), *new_message);
+  }
+  return Controller(logger, controller_messages, controller_configuration_messages);
+}
+
+Controller::Controller(core::ILogger &logger,
+                       const std::unordered_map<std::string, core::CanFrame> messages,
+                       const std::vector<core::CanFrame> configuration_messages)
```
mifrandir, line 65: make these references to avoid a copy (they'll be copied below)

lib/motors/controller.cpp
```diff
@@ -1,9 +1,106 @@
...
+    return std::nullopt;
+  }
+  core::CanFrame new_message;
+  new_message.can_id  = message["id"].GetInt();
+  new_message.can_dlc = kControllerCanFrameLength;
+  // TODO sort endianness conversion for index and data
+  //  newmessage.data[0&1] = index
+  new_message.data[2] = message["subindex"].GetInt();
+  new_message.data[3] = 0;  // padding
+  // newmessage.data[4-7] = index
```
mifrandir, line 102: ?

lib/motors/controller.hpp
```diff
@@ -1,19 +1,40 @@
...
+#include <rapidjson/error/en.h>
+#include <rapidjson/istreamwrapper.h>
 
 #include "core/logger.hpp"
+#include "core/types.hpp"
 
 namespace hyped::motors {
 enum class ControllerStatus { kControllerTemperatureExceeded, kUnrecoverableWarning, kNominal };
+// every frame sent to the controller contains 8 bytes
+static constexpr std::uint8_t kControllerCanFrameLength = 8;
```
mifrandir, line 19: is this a more global constant? will all our can frames be of this length? if so, move this to `io/can.hpp`

mifrandir (changes requested):

lib/io/can.hpp
```diff
@@ -3,6 +3,8 @@
 #include <core/types.hpp>
 
 namespace hyped::io {
+// every frame sent to the controller contains 8 bytes
+static constexpr std::uint8_t kControllerCanFrameLength = 8;
```
maxguy2001, line 7: Idk much about can  but is this something that will only really be used here or will sensors need to use this value too (i.e. should it live here or should it live somewhere more accessible?)

CMakeLists.txt
kshxtij: Yeah i think we'd have to install it manually since the brew tap doesn't seem to work on mac 🤔

kshxtij: Yeah i think we'd have to install it manually since the brew tap doesn't seem to work on mac 🤔

kshxtij: and building from scratch and installing keeps throwing errors, i'll investigate this and fix it after I finish some more ROS setup for the hackathon.

lib/io/can.hpp
TomLonergan03: I don't know what length frames the BMS uses. 8 bytes is the maximum length for a (non-extended) CAN frame, but they might be using shorter frames. With #21 the struct for a can frame will be in can.hpp so I think this is the right place for it.

@SnickeyX will all your BMS CAN frames be 8 bytes long?

mifrandir: I just realised that this is in `io`... surely that is THE most accessible place to have it?

SnickeyX: they seem to be 8 bytes for BMS, but ideally we want these to variable because more sensors are going to be on CAN apparently and I don't yet know what their frame sizes are (waiting on a confirmation of the sensor choice atm)

TomLonergan03: there is also already a `CAN_MAX_DLEN` macro defined in the linux header which is set to 8. i think this is a const for mtc only and other cases should define their own sizes in their own places

TomLonergan03:
For parsing hex i think we either do `try {...} catch {...}` or use `std::stringstream`, not sure which is better

lib/motors/controller.cpp
```diff
@@ -1,9 +1,158 @@
...
+  can_id_hex << std::hex << message["id"].GetString();
+  if (!can_id_hex.good()) {
+    logger.log(core::LogLevel::kFatal, "Invalid message ID in CAN message file");
+    return std::nullopt;
+  }
+  if (can_id_hex.eof()) {
+    logger.log(core::LogLevel::kFatal, "No message ID in CAN message file");
+    return std::nullopt;
+  }
+  can_id_hex >> new_message.can_id;
```
SnickeyX, line 107: declare `new_message` right above this

lib/motors/controller.hpp
```diff
@@ -1,19 +1,41 @@
...
+#include <rapidjson/istreamwrapper.h>
 
-#include "core/logger.hpp"
+#include <core/logger.hpp>
+#include <core/types.hpp>
+#include <io/can.hpp>
 
 namespace hyped::motors {
+// every frame sent to the controller contains 8 bytes
+static constexpr std::uint8_t kControllerCanFrameLength = 8;
```
SnickeyX, line 19: remove `static` here

lib/motors/controller.hpp
```diff
@@ -1,19 +1,41 @@
...
 
 class Controller {
  public:
-  Controller(core::ILogger &logger);
+  static std::optional<Controller> create(core::ILogger &logger,
+                                          const std::string &message_file_path);
   void processErrorMessage(const std::uint16_t error_code);
   ControllerStatus processWarningMessage(const std::uint8_t warning_code);
+  static std::optional<core::CanFrame> parseJsonCanFrame(
+    rapidjson::GenericObject<false, rapidjson::Value> message, core::ILogger &logger);
```
SnickeyX, line 29: argument order, logger first

lib/motors/controller.hpp
```diff
@@ -1,19 +1,41 @@
...
 
  private:
+  Controller(core::ILogger &logger,
+             const std::unordered_map<std::string, core::CanFrame> &messages,
+             const std::vector<core::CanFrame> &configuration_messages);
   core::ILogger &logger_;
+  // TODO replace core::CanFrame with io::CanFrame once merged
+  const std::unordered_map<std::string, core::CanFrame> messages_;
+  const std::vector<core::CanFrame> configuration_messages_;
```
SnickeyX, line 38: idt this is in the style guide just yet, but separating private varibles from private methods with another private tag just makes it more readable

lib/motors/controller.hpp
```diff
@@ -1,19 +1,41 @@
...
   ControllerStatus processWarningMessage(const std::uint8_t warning_code);
+  static std::optional<core::CanFrame> parseJsonCanFrame(
+    rapidjson::GenericObject<false, rapidjson::Value> message, core::ILogger &logger);
 
  private:
+  Controller(core::ILogger &logger,
+             const std::unordered_map<std::string, core::CanFrame> &messages,
+             const std::vector<core::CanFrame> &configuration_messages);
   core::ILogger &logger_;
+  // TODO replace core::CanFrame with io::CanFrame once merged
```
SnickeyX, line 36: TODOLater since you are not implementing it in this PR

SnickeyX (changes requested):
just formatting things + do fix your merge conflict

TomLonergan03:
will probably move to using boost.json because rapidjson is bad + unmaintained + L + tencent why + ratio + who asked + no release in 7 years

lib/motors/controller.hpp
```diff
@@ -1,19 +1,43 @@
...
   void processErrorMessage(const std::uint16_t error_code);
   ControllerStatus processWarningMessage(const std::uint8_t warning_code);
+  static std::optional<core::CanFrame> parseJsonCanFrame(
+    core::ILogger &logger, rapidjson::GenericObject<true, rapidjson::Value> message);
 
  private:
+  Controller(core::ILogger &logger,
+             const std::unordered_map<std::string, core::CanFrame> &messages,
+             const std::vector<core::CanFrame> &configuration_messages);
   core::ILogger &logger_;
```
SnickeyX, line 35: should be in the private below

SnickeyX (changes requested):
minor change, wheyyy

SnickeyX (approved):
buggat

kshxtij (approved):

PR #75, opened 2023-01-25
Fixes for #38

mifrandir:
In light of the comments on #38, here are some fixes.

I think it would have been to hard to convey through code review.

### DIFF
--- a/lib/navigation/preprocess_encoders.cpp
+++ b/lib/navigation/preprocess_encoders.cpp
@@ -24,7 +24,7 @@ std::optional<core::EncoderData> EncodersPreprocessor::processData(
 }
 
 std::optional<EncodersPreprocessor::Statistics> EncodersPreprocessor::getStatistics(
-  const core::EncoderData &encoder_data)
+  const core::EncoderData &encoder_data) const
 {
   if (num_reliable_encoders_ > core::kNumEncoders
       || num_reliable_encoders_ < core::kNumEncoders - 1) {
--- a/lib/navigation/preprocess_encoders.cpp
+++ b/lib/navigation/preprocess_encoders.cpp
@@ -1,3 +1,4 @@
+#include "consts.hpp"
 #include "preprocess_encoders.hpp"
 
 #include "core/types.hpp"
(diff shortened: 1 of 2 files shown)


lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,91 +1,94 @@
...
 
 EncodersPreprocessor::EncodersPreprocessor(core::ILogger &logger)
     : logger_(logger),
-      encoder_outliers_({0, 0, 0, 0}),
-      encoders_reliable_({true, true, true, true}),
-      num_reliable_encoders_(core::kNumImus)
+      num_consecutive_outliers_per_encoder_({0, 0, 0, 0}),
+      is_reliable_per_encoder_({true, true, true, true}),
+      num_reliable_encoders_(core::kNumImus),
+      max_consecutive_outliers_(10)
```
maxguy2001, line 13: What's the benefit of initializing here as opposed to just storing it as a constant class member?

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,91 +1,94 @@
...
 {
-  const auto encoder_data = detectOutliers(raw_encoder_data);
-  SensorChecks sensors    = checkReliable();
-  if (!encoder_data.has_value() || sensors == SensorChecks::kUnacceptable) {
-    return std::nullopt;
-  } else {
-    return encoder_data;
-  }
+  const auto encoder_data = sanitise(raw_encoder_data);
+  const auto sensors      = checkReliable();
```
maxguy2001, line 21: Why `auto` if we know return type?

lib/navigation/preprocess_encoders.cpp
```diff
@@ -1,91 +1,94 @@
...
+  if (sensors == SensorChecks::kUnacceptable) { return std::nullopt; }
+  return encoder_data;
 }
 
-std::optional<core::EncoderData> EncodersPreprocessor::detectOutliers(
-  const core::EncoderData encoder_data)
+std::optional<EncodersPreprocessor::Statistics> EncodersPreprocessor::getStatistics(
```
maxguy2001, line 26: Not a fan of "statistics" it doesn't really describe functionality (we specifically calculate quartiles)?

mifrandir:
@maxguy2001 merge whenever you want

lib/navigation/preprocess_encoders.cpp
mifrandir: we always know the return type, for every function call. auto merely leaves it to the compiler to fill in what is known.

i prefer in most cases because it allows us to change the return types without breaking lots of code (that is return-type agnostic)

the exceptions are 
1. primitive types such as `bool` and `core::Float`
2. if we are converting types, e.g. to an interface. think 
   ```cpp
    const sensors::ICan can = sensors::HardwareCan::create();
    ```

lib/navigation/preprocess_encoders.cpp
mifrandir: the input to this function is a set of data, the output is a median, and thresholds

this has nothing to do with quartiles from the callers perspective

i agree, though; there's a TODO comment indicating that I don't like the name

we could go for `MedianAndThresholds` if there isn't a better option

lib/navigation/preprocess_encoders.cpp
mifrandir: this is a tricky question, the short answer is consistency

if we were to agree that, within HYPED, we use in-class initialisation wherever possible then i wouldn't have anything to object to; however, since lots of these were already being initialised here, it feels weird to do it in two places

maxguy2001 (approved):

PR #78, opened 2023-01-30
SNS - BMS

SnickeyX:
Code to interface with ???

- [ ] Figure out structure 
- [ ] Find out if CAN IDs are correct
- [ ] Find out what power want 

Note: the BMS being used is changed so some other company entirely, waiting to know which one that is before I can continue with this

### DIFF
--- a/lib/sensors/bms.cpp
+++ b/lib/sensors/bms.cpp
@@ -1,0 +1,46 @@
+#include "bms.hpp"
+
+namespace hyped::sensors {
+Bms::Bms(core::ILogger &logger, io::HardwareCan &can) : logger_(logger), can_(can)
+{
+}
+
+std::optional<BmsData> Bms::receiveBmsData()
+{
+  // TODO: Implement
+  // Outline of possible implementation:
+  BmsData bms_data;
+  const auto optional_cell_temperatures = receiveCellTemperature();
+  if (!optional_cell_temperatures) {
+    logger_.log(core::LogLevel::kFatal, "failed to receive cell temperatures");
+  }
+  bms_data.temperatures = optional_cell_temperatures.value();
+  // return bms_data;
+  return std::nullopt;
+}
+
+std::optional<std::array<core::Float, NUM_BMS_CELLS>> Bms::receiveCellTemperature()
+{
+  // TODO: Implement
+  // Outline of possible implementation:
+  const auto bmsProcessor = std::make_shared<BmsProcessor>(logger_);
+  can_.addProcessor(static_cast<std::uint16_t>(BmsPacketIds::kBmsStatusTemperatures), bmsProcessor);
+  const auto frame = can_.receive();
+  if (!frame) { logger_.log(core::LogLevel::kFatal, "Failed to receive cell temperatures"); }
+  // would then parse frame and return temperatures
+  return std::nullopt;
+}
+
+std::optional<io::CanFrame> Bms::receiveCellVoltage()
+{
+  // TODO: Implement
+  return std::nullopt;
+}
+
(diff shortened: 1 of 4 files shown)


lib/sensors/bms.hpp
```diff
@@ -0,0 +1,118 @@
...
+#include <optional>
+
+#include <bms_processor.hpp>
+#include <core/logger.hpp>
+#include <io/hardware_can.hpp>
+
+namespace hyped::sensors {
+
+// TODOLater: Confirm with propulsion
+static constexpr std::uint8_t NUM_BMS_CELLS               = 12;
```
maxguy2001, line 13: `kNumBmsCells`?

lib/sensors/bms.hpp
```diff
@@ -0,0 +1,118 @@
...
+static constexpr std::uint8_t NUM_BMS_TEMPERATURE_SENSORS = 4;
```
maxguy2001, line 14: style guide?

lib/sensors/bms.cpp
```diff
@@ -0,0 +1,46 @@
...
+std::optional<BmsData> Bms::receiveBmsData()
+{
+  // TODO: Implement
+  // Outline of possible implementation:
+  BmsData bms_data;
+  const auto optional_cell_temperatures = receiveCellTemperature();
+  if (!optional_cell_temperatures) {
+    logger_.log(core::LogLevel::kFatal, "failed to receive cell temperatures");
+  }
+  bms_data.temperatures = optional_cell_temperatures.value();
```
maxguy2001, line 17: const

lib/sensors/bms.hpp
```diff
@@ -0,0 +1,118 @@
...
+  ~Bms() = default;
+
+  std::optional<BmsData> receiveBmsData();
+
+ private:
+  std::optional<std::array<core::Float, NUM_BMS_CELLS>> receiveCellTemperature();
+  std::optional<io::CanFrame> receiveCellVoltage();
+  std::optional<io::CanFrame> receivePackCurrent();
+
+ private:
```
maxguy2001, line 113: You're already private?

maxguy2001 (changes requested):

lib/sensors/bms.hpp
SnickeyX: 'tis but a draft, the whole PR was  for discussing structure. Once its ready I'll open it for reviews.

lib/sensors/bms.hpp
SnickeyX: separating private methods from variables for readability

lib/sensors/bms.cpp
```diff
@@ -0,0 +1,46 @@
...
+std::optional<std::array<core::Float, NUM_BMS_CELLS>> Bms::receiveCellTemperature()
+{
+  // TODO: Implement
+  // Outline of possible implementation:
+  const auto bmsProcessor = std::make_shared<BmsProcessor>(logger_);
+  can_.addProcessor(static_cast<std::uint16_t>(BmsPacketIds::kBmsStatusTemperatures), bmsProcessor);
+  const auto frame = can_.receive();
+  if (!frame) { logger_.log(core::LogLevel::kFatal, "Failed to receive cell temperatures"); }
+  // would then parse frame and return temperatures
+  return std::nullopt;
```
TomLonergan03, line 31: for reference (i think you know this already but will put it here just in case):
`can_.addProcessor` should be called on initialisation, then when a message with the ID comes through `can_ ` will call `BmsProcessor.processMessage(received_frame)`. This is the function that should then call things like `updateCellTemperatures` for example, which can deal with the actual contents of the message

PR #79, opened 2023-01-30
SNS-Keyence

sa-fx:
Initial push of stripe sensor (_Keyence_) implementation
Feedback greatly welcomed


lib/sensors/stripe_sensor.cpp
```diff
@@ -0,0 +1,21 @@
+#include "stripe_sensor.hpp"
+
+namespace hyped::sensors {
+
+    Sensor::Sensor(const std::uint8_t newPin){
+        pin = newPin;
+        HardwareGpio sensor(&log_);
+        sensor.getReader(newPin);
+    };
```
TomLonergan03, line 9: `pin`, initialise class values in the initialiser list 
ie 
`Sensor::Sensor(const std::uint8_t pin):pin_(pin), logger_(logger) {`
`    // do stuff`
`}`

also you need to include `log` in the constructor args, plus should be called `logger_` for consistency across the repo

TomLonergan03: also this should not be class `Sensor`, probably `Keyence` but I'm not sure what the official term for it is

lib/sensors/stripe_sensor.hpp
```diff
@@ -0,0 +1,20 @@
...
+        public:
+            Sensor(const std::uint8_t newPin);
+            ~Sensor();
+
+            int getStripeCount();
+            void updateStripes();
+        private:
+            std::uint8_t pin;
+            int stripeCount = 0;
+            hyped::core::ILogger &log_;
```
TomLonergan03, line 18: Class data members are named using lower snake and a trailing underscore `foo_bar_` so `pin_` and `stripe_count_`

lib/sensors/stripe_sensor.cpp
TomLonergan03: and the reader will need to be stored in a class data member

lib/sensors/stripe_sensor.hpp
TomLonergan03: as an aside, general variables are named as `foo_bar` and `constexpr`s as `kFooBar`

TomLonergan03 (changes requested):
welcome to hyped, sorry for bully on first commit lol

also run `clang-format` before commits, it will keep everything correctly indented etc.

lib/sensors/stripe_sensor.cpp
```diff
@@ -0,0 +1,21 @@
+#include "stripe_sensor.hpp"
```
maxguy2001, line 1: Probably call these files "keyence_sensor" or similar for consistency?

lib/sensors/stripe_sensor.cpp
```diff
@@ -0,0 +1,21 @@
...
+        sensor.getReader(newPin);
+    };
+    
+    int Sensor::getStripeCount(){
+        return stripeCount;
+    }
+
+    void Sensor::updateStripes(){
+        if(sensor.read() == core::DigitalSignal::kHigh){
+            stripeCount++;
```
maxguy2001, line 17: `++stripe_count_`

lib/sensors/stripe_sensor.hpp
```diff
@@ -0,0 +1,20 @@
...
+    class Sensor{
+        public:
+            Sensor(const std::uint8_t newPin);
+            ~Sensor();
+
+            int getStripeCount();
+            void updateStripes();
+        private:
+            std::uint8_t pin;
+            int stripeCount = 0;
```
maxguy2001, line 17: Should this not be initialized in constructor?

maxguy2001: Also given that the stripe count starts at zero and is strictly increasing, either `uint16_t` or `uint32_t` would probably be more appropriate?

maxguy2001 (changes requested):

lib/sensors/stripe_sensor.cpp
```diff
@@ -0,0 +1,27 @@
...
+}
+
+int StripeSensor::getStripeCount()
+{
+  return stripeCount;
+}
+
+void StripeSensor::updateStripes()
+{
+  if (keyence.read() == core::DigitalSignal::kHigh) { stripeCount++; };
```
maxguy2001, line 24: `++StripeCount`

maxguy2001: also `stripe_count_` (look at naming conventions for class members).

lib/sensors/stripe_sensor.hpp
```diff
@@ -0,0 +1,23 @@
...
+#include <core/logger.hpp>
+#include <io/hardware_gpio.hpp>
+
+namespace hyped::sensors {
+class StripeSensor {
+ public:
+  StripeSensor(hyped::core::ILogger &log, const std::uint8_t newPin);
+  ~StripeSensor();
+
+  int getStripeCount();
```
maxguy2001, line 15: some sort of unsigned integer would be better for this?

lib/sensors/stripe_sensor.cpp
```diff
@@ -0,0 +1,27 @@
+#include "stripe_sensor.hpp"
+
+namespace hyped::sensors {
+
+StripeSensor::StripeSensor(hyped::core::ILogger &log, const std::uint8_t newPin)
+    : pin(newPin),
```
maxguy2001, line 6: `new_pin`

maxguy2001 (changes requested):

lib/sensors/stripe_sensor.hpp
```diff
@@ -0,0 +1,25 @@
...
+class Keyence {
+ public:
+  Keyence(core::ILogger &log, const std::uint8_t newPin);
+  ~Keyence();
+
+  int getStripeCount();
+  void updateStripes();
+
+ private:
+  std::uint8_t pin;
```
maxguy2001, line 20: `pin_`. Look at the google C++ style guide if you're not sure but all class members like this should have should be snake case and have an underscore at the end (useful convention for identifying class members).

lib/sensors/stripe_sensor.hpp
```diff
@@ -0,0 +1,25 @@
...
+  uint16_t stripe_count_;
+  std::optional<std::shared_ptr<IGpioReader>> keyence;
```
maxguy2001, line 22: Same here, `keyence_`.

lib/sensors/stripe_sensor.cpp
```diff
@@ -0,0 +1,26 @@
...
+  stripe_count_ = 0;
+  io::HardwareGpio hardware(logger_);
+  keyence = hardware.getReader(pin);
+};
+
+Keyence::~Keyence()
+{
+}
+
+int Keyence::getStripeCount()
```
maxguy2001, line 16: The return type of this function is `int` but `stripe_count_` is `uint16_t` so you'll have to change the return type here.

lib/sensors/stripe_sensor.cpp
```diff
@@ -0,0 +1,26 @@
+#include "stripe_sensor.hpp"
+
+namespace hyped::sensors {
+
+Keyence::Keyence(core::ILogger &log, const std::uint8_t newPin) : pin(newPin), logger_(log)
```
maxguy2001, line 5: `new_pin`. Function arguments should be in snake case.

maxguy2001 (changes requested):

lib/sensors/stripe_sensor.cpp
SnickeyX: also `&logger` and `logger_(logger)`

lib/sensors/stripe_sensor.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include <cstdint>
+#include <optional>
+
+#include "io/gpio.hpp"
+#include <core/logger.hpp>
+#include <io/hardware_gpio.hpp>
```
SnickeyX, line 8: `<io/gpio.hpp>` and remove `<io/hardware_gpio.hpp>`, the interface should suffice

lib/sensors/stripe_sensor.hpp
SnickeyX: actually, this should be called `gpio_reader_` and not `keyence_`

lib/sensors/stripe_sensor.cpp
```diff
@@ -0,0 +1,26 @@
...
+Keyence::~Keyence()
+{
+}
+
+int Keyence::getStripeCount()
+{
+  return stripe_count_;
+}
+
+void Keyence::updateStripes()
```
SnickeyX, line 21: `updateStripeCount`

lib/sensors/stripe_sensor.cpp
```diff
@@ -0,0 +1,26 @@
+#include "stripe_sensor.hpp"
+
+namespace hyped::sensors {
+
+Keyence::Keyence(core::ILogger &log, const std::uint8_t newPin) : pin(newPin), logger_(log)
+{
+  stripe_count_ = 0;
+  io::HardwareGpio hardware(logger_);
```
SnickeyX, line 8: remove this

lib/sensors/stripe_sensor.cpp
SnickeyX, line 5: as a parameter, take in an `IGpioReader`, then initialise the `gpio_reader_` with it. for this, a `static create` function (like all the IO code) would be good. Its purpose would simply be to create a Keyence object, but the create function itself should take an `IGpio` and a `pin` so it could make a `gpio_reader` to then pass to the constructor of keyence.

SnickeyX (changes requested):

arjunnaha (approved):
lgtm

SnickeyX (approved):
looks good, time to test

HTTYDKing (approved):
Looks good to me

lib/sensors/keyence.cpp
```diff
@@ -0,0 +1,37 @@
+#include "keyence.hpp"
+
+namespace hyped::sensors {
+
+std::optional<Keyence> Keyence::create(core::ILogger &logger, io::IGpio &gpio, std::uint8_t new_pin)
```
maxguy2001, line 5: `const`?

lib/sensors/keyence.hpp
```diff
@@ -0,0 +1,39 @@
...
+   * @brief Configures and creates an instance of Keyence
+   */
+  static std::optional<Keyence> create(core::ILogger &logger,
+                                       io::IGpio &gpio,
+                                       std::uint8_t new_pin);
+  ~Keyence();
+
+  std::uint8_t getStripeCount();
+
+  /**
```
maxguy2001, line 23: Remove this comment. The naming here makes it obvious enough.

lib/sensors/keyence.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include <memory>
+#include <optional>
+
+#include <core/logger.hpp>
+#include <io/gpio.hpp>
+
+namespace hyped::sensors {
+class Keyence {
+ public:
+  /**
```
maxguy2001, line 13: Is this comment necessary?

maxguy2001 (changes requested):

kshxtij (approved):

PR #81, opened 2023-02-02
Link against Boost

kshxtij:


mifrandir (changes requested):
can you please add a test to convince me that boost works?

maybe for eigen, too

kshxtij:
Sure, will do that just within test/core

mifrandir (approved):

SnickeyX (approved):
Code is fine, could you add eigen + boost as an explicit dependency that needs to be downloaded to the readme or wiki

TomLonergan03 (approved):
builds

maxguy2001 (approved):

PR #91, opened 2023-02-04
SNS - Overhaul

SnickeyX:
- [x] Update temp + accelerometer code
- [x] Update REPL
- [x] Update Mux code + change it to be I2c specfic as an analog mux is on its way 
- [x] kWarn
- [x] Separate function to check for if "data ready" before calling read on sensors
- [x] Update ADC code to convert raw value to actual voltage value
- [x] Adding skeleton code for sensors to be implemented soon™
- [x] virtual tags for derived methods in I/O
- [x] enforcing 0 default initializations
- [x] fixed mistakes in I/O  
- [x] static constexpr for values that never need to accessed outside of a class (register addresses or class constants)


lib/debug/repl.cpp
```diff
@@ -196,6 +196,27 @@ std::optional<std::unique_ptr<Repl>> Repl::fromFile(const std::string &path)
...
+                "Missing required field 'sensors.temperature.enabled' in configuration file");
+    return std::nullopt;
+  }
+  if (temperature["enabled"].GetBool()) {
+    if (!temperature.HasMember("bus")) {
+      logger_.log(core::LogLevel::kFatal,
+                  "Missing required field 'sensors.temperature.bus' in configuration file");
+      return std::nullopt;
+    }
+    const auto device_address = temperature["device_address"].GetUint();
```
TomLonergan03, line 216: should check if device_address exists before accessing it

lib/debug/repl.cpp
```diff
@@ -548,4 +529,71 @@ void Repl::addUartCommands(const std::uint8_t bus)
...
 }
 
+void Repl::addAccelerometerCommands(const std::uint8_t bus, const std::uint8_t device_address)
+{
+  const auto optional_i2c = io::HardwareI2c::create(logger_, bus);
+  if (!optional_i2c) {
+    logger_.log(core::LogLevel::kFatal, "Failed to create I2C instance on bus %d", bus);
+    return;
+  }
+  const auto i2c = std::move(*optional_i2c);
```
TomLonergan03, line 539: is `std::move` definitely necessary? my CAN testing didn't end up requiring it as `shared_ptr` should not be copying when it is passed around

TomLonergan03 (changes requested):

lib/debug/repl.cpp
SnickeyX: should have not simply copied

lib/debug/repl.cpp
SnickeyX: no, but apparently its quicker to move instead of copying it so maybe keeping the` std::move`

TomLonergan03: looking at it yea that makes sense

lib/core/types.hpp
```diff
@@ -43,7 +44,7 @@ using EncoderData          = std::array<std::uint32_t, kNumEncoders>;
 using KeyenceData          = std::array<std::uint32_t, kNumKeyence>;
 
 // data produced by the accelerometer sensor
-// !! the values are in mg !!
+// values are in mg
```
TomLonergan03, line 47: kinda picky but i would prefer to distinguish between milli g and milligram

lib/debug/repl.cpp
```diff
@@ -548,4 +541,71 @@ void Repl::addUartCommands(const std::uint8_t bus)
   }
 }
 
+void Repl::addAccelerometerCommands(const std::uint8_t bus, const std::uint8_t device_address)
+{
+  const auto optional_i2c = io::HardwareI2c::create(logger_, bus);
```
TomLonergan03, line 546: initialising a new i2c instance for each i2c sensor is probably not good? i'm not sure exactly
or is this The Big Issue With The Repl That Is Yet To Be Solved?

lib/io/hardware_adc.cpp
```diff
@@ -54,7 +55,8 @@ std::optional<std::uint16_t> Adc::resetAndRead4(const int file_descriptor)
     return std::nullopt;
   }
   const int raw_voltage = std::atoi(read_buffer);
-  return static_cast<std::uint16_t>(raw_voltage);  // max value is 2^12-1 = 4095
+  // convert raw voltage to voltage between [0, 1.8] in V
+  return static_cast<std::uint16_t>((raw_voltage / kMaxAdcRawValue) * kMaxAdcVolatage);
```
TomLonergan03, line 59: `kMaxAdcVoltage`

lib/sensors/accelerometer.cpp
```diff
@@ -2,7 +2,36 @@
...
+                                                   const std::uint8_t device_address)
+{
+  // check we are communicating with the correct sensor
+  const auto device_id = i2c->readByte(device_address, kDeviceIdAddress);
+  if (!device_id) {
+    logger.log(core::LogLevel::kFatal, "Failure to read device id of accelerometer");
+    return std::nullopt;
+  }
+  if (*device_id != kExpectedDeviceIdValue) {
+    logger.log(core::LogLevel::kFatal, "Failure accelerometer didn't give correct device id");
```
TomLonergan03, line 17: `"Failure, accelerometer .....`

lib/sensors/accelerometer.cpp
```diff
@@ -13,6 +42,44 @@ Accelerometer::~Accelerometer()
 {
 }
 
+std::optional<core::Result> Accelerometer::isVaueReady()
```
TomLonergan03, line 45: optional result is not a great type imo, but i understand there might not be a better way to do this

TomLonergan03: Also `isValueReady`

lib/sensors/accelerometer.hpp
```diff
@@ -50,21 +51,37 @@ constexpr std::uint8_t kExpectedDeviceIdValue = 0x44;
 
 class Accelerometer : public II2cMuxSensor<core::RawAccelerationData> {
  public:
-  Accelerometer(core::ILogger &logger, io::II2c &i2c, const std::uint8_t channel);
+  static std::optional<Accelerometer> create(core::ILogger &logger,
+                                             std::shared_ptr<io::II2c> i2c,
+                                             const std::uint8_t channel,
+                                             const std::uint8_t device_address
+                                             = kDefaultAccelerometerAddress);
```
TomLonergan03, line 58: why set a default address here? wouldn't this be better stated every time when creating the accelerometer

lib/sensors/i2c_converter.hpp
```diff
@@ -0,0 +1,33 @@
...
+  ~I2cConverter();
+
+  std::uint8_t getAddress();
+
+ private:
+  I2cConverter(const std::uint8_t device_address);
+
+ private:
+  core::ILogger &logger_;
+  std::uint8_t address_;
```
TomLonergan03, line 30: `const` i assume off of the constructor

lib/sensors/i2c_mux.hpp
```diff
@@ -10,19 +10,23 @@
 
 namespace hyped::sensors {
 
-static constexpr std::uint8_t kDefaultMuxAddress = 0x70;
-static constexpr std::uint8_t kMaxNumChannels    = 8;
-static constexpr core::Float kFailureThreshold
-  = 0.25;  // TODOLater: finalize this value with Electronics
+constexpr std::uint8_t kDefaultMuxAddress = 0x70;
+constexpr std::uint8_t kMaxNumChannels    = 8;
+constexpr core::Float kFailureThreshold = 0.25;  // TODOLater: finalize this value with Electronics
```
TomLonergan03, line 15: what does the failure threshold represent? a percentage? or some other value?

lib/sensors/i2c_mux.hpp
```diff
@@ -68,68 +72,61 @@ std::optional<std::array<T, N>> Mux<T, N>::readAllChannels()
...
     return core::Result::kFailure;
   }
   const std::uint8_t channel_buffer = 1 << channel;
-  const auto i2c_write_result       = i2c_.writeByte(mux_address_, channel_buffer);
+  const auto i2c_write_result       = i2c_->writeByte(mux_address_, channel_buffer);
   if (i2c_write_result == core::Result::kSuccess) {
-    logger_.log(core::LogLevel::kInfo, "Mux Channel %d selected", channel);
+    logger_.log(core::LogLevel::kInfo, "I2c Mux Channel %d selected", channel);
     return core::Result::kSuccess;
   } else {
```
TomLonergan03, line 114: don't need the `else`

lib/sensors/i2c_mux.hpp
```diff
@@ -68,68 +72,61 @@ std::optional<std::array<T, N>> Mux<T, N>::readAllChannels()
...
+core::Result I2cMux<T, N>::closeAllChannels()
 {
   const std::uint8_t clear_channel_buffer = 0x00;
-  const auto i2c_write_result             = i2c_.writeByte(mux_address_, clear_channel_buffer);
+  const auto i2c_write_result             = i2c_->writeByte(mux_address_, clear_channel_buffer);
   if (i2c_write_result == core::Result::kSuccess) {
-    logger_.log(core::LogLevel::kInfo, "All mux channels closed");
+    logger_.log(core::LogLevel::kInfo, "All i2c mux channels closed");
     return core::Result::kSuccess;
   } else {
```
TomLonergan03, line 128: drop the `else`

lib/sensors/thermister.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+  std::uint16_t getTemperature();
+
+ private:
+  Thermistor(core::ILogger &logger, std::shared_ptr<io::IAdc> adc, const std::uint8_t new_pin);
+
+ private:
+  core::ILogger &logger_;
+  std::shared_ptr<io::IAdc> adc_;
+  std::uint8_t pin_;
```
TomLonergan03, line 29: `const` I assume

lib/sensors/temperature.cpp
```diff
@@ -15,33 +32,37 @@ Temperature::~Temperature()
...
                 "Failed to read, temperature sensor is not ready to be read at channel %d",
                 channel_);
-    return std::nullopt;
+    return core::Result::kFailure;
   }
   if (status_check_result.value() == kTemperatureOverUpperLimit) {
-    logger_.log(core::LogLevel::kFatal,
+    logger_.log(core::LogLevel::kWarn,
                 "Failed to read, temperature is above upper limit at channel %d",
                 channel_);
```
TomLonergan03, line 52: what does being over upper or below lower limits mean? is it a temporary thing or not? if not then the return should distinguish between these and not ready

TomLonergan03 (changes requested):

lib/debug/repl.cpp
SnickeyX: the latter

lib/sensors/accelerometer.cpp
SnickeyX: copy pasting ;//

lib/sensors/accelerometer.cpp
SnickeyX: yeah not sure how to do this better atm. in my head this is running on a ros node where isValueReady is called before read

lib/sensors/accelerometer.hpp
SnickeyX: dont understand your point here, this is mostly in-case we are told that electronics want to use an address other than the default, in which case we can provide another address if needed. (though this should be an enum ideally, will change)

lib/sensors/i2c_converter.hpp
SnickeyX: so this is fully just skeleton code, but yeah I'll remove the `address` part fully for now

lib/sensors/i2c_mux.hpp
SnickeyX: percentage - courtesy of Bas

lib/sensors/i2c_mux.hpp
SnickeyX: I don't think that removing this matters too much

lib/sensors/temperature.cpp
SnickeyX: not sure since the datasheet was a bit vague here, though you are right, I'll make this an fatal error

lib/core/types.hpp
SnickeyX: i mean, the accelerometer surely wont be giving grams as units

lib/sensors/i2c_mux.hpp
TomLonergan03: stick that in a comment

lib/sensors/accelerometer.hpp
TomLonergan03: i feel like it's better practice to make it clear when calling what the address is instead of implying it sometimes and explicitly stating it other times

lib/core/types.hpp
TomLonergan03: it caught me out for a few seconds but maybe im just dumb

lib/sensors/i2c_mux.hpp
TomLonergan03: it is code that doesn't do anything so it is only visual noise to paraphrase something i think i saw franz say somewhere

fite me

TomLonergan03: also should be `if result == failure return failure` then afterwards `return success`
in keeping with using guards as elsewhere

kshxtij:
Think it's spelled Thermistor btw 😂

SnickeyX:
> Think it's spelled Thermistor btw 😂 

🥺

lib/core/types.hpp
maxguy2001: To be fair, mg is the SI notation for milligrams so you might as well use the full words for clarity (in case someone reding the code isn't familiar with accelerometers)?

lib/sensors/accelerometer.cpp
```diff
@@ -13,6 +42,44 @@ Accelerometer::~Accelerometer()
...
+  const std::int32_t y_acceleration = getAccelerationFromRawValue(*result_y);
+  const auto result_z               = getRawAcceleration(core::Axis::kZ);
+  if (!result_z) { return std::nullopt; }
+  const std::int32_t z_acceleration = getAccelerationFromRawValue(*result_z);
+  const std::optional<core::RawAccelerationData> acceleration_3_axis{
+    std::in_place,
+    x_acceleration,
+    y_acceleration,
+    z_acceleration,
+    std::chrono::system_clock::now()};
```
maxguy2001, line 78: Would you not want to pass in a time source and use time_.now() to make the code testable?

lib/sensors/accelerometer.hpp
```diff
@@ -50,21 +51,37 @@ constexpr std::uint8_t kExpectedDeviceIdValue = 0x44;
...
   ~Accelerometer();
 
-  core::Result configure();
+  /*
```
maxguy2001, line 61: Is this comment necessary? I feel like the function name makes it's purpose quite obvious?

lib/sensors/i2c_converter.hpp
```diff
@@ -0,0 +1,31 @@
...
+ * @brief ADC (Texas Instruments) to be used for the hall-effect sensors (Farnell)
+ * @details This class is used to convert the analog signal from the hall-effect sensors to a
+ * digital signal over I2c
+ */
+class I2cConverter {
+ public:
+  static std::optional<I2cConverter> create(core::ILogger &logger);
+  ~I2cConverter();
+
+  std::uint8_t getAddress();
```
maxguy2001, line 22: you've got a `getAddress()` function but no `address_` class member...

lib/sensors/i2c_mux.hpp
```diff
@@ -10,19 +10,24 @@
 
 namespace hyped::sensors {
 
-static constexpr std::uint8_t kDefaultMuxAddress = 0x70;
-static constexpr std::uint8_t kMaxNumChannels    = 8;
-static constexpr core::Float kFailureThreshold
-  = 0.25;  // TODOLater: finalize this value with Electronics
+constexpr std::uint8_t kDefaultMuxAddress = 0x70;
+constexpr std::uint8_t kMaxNumChannels    = 8;
+// perentage of sensors that can be unusable before the mux is considered unusable
```
maxguy2001, line 15: Would we not want more specific fail states (i.e. "all temperature sensors failing" instead of just a proportion of all sensors)?

lib/sensors/i2c_sensors.hpp
```diff
@@ -16,7 +16,6 @@ class II2cMuxSensor {
   /*
    * This function carries out the initilization steps for a particular sensor.
    */
-  virtual core::Result configure()        = 0;
   virtual std::optional<T> read()         = 0;
```
maxguy2001, line 19: Do all these sensors not use a `create` function too?

lib/sensors/temperature.cpp
```diff
@@ -15,20 +32,19 @@ Temperature::~Temperature()
...
-  // TODOLater: change this from a fatal error to a debug and re-prompt read (with ROS)
   if (status_check_result.value() == kBusy) {
-    logger_.log(core::LogLevel::kFatal,
+    logger_.log(core::LogLevel::kWarn,
                 "Failed to read, temperature sensor is not ready to be read at channel %d",
                 channel_);
-    return std::nullopt;
+    return core::Result::kFailure;
   }
   if (status_check_result.value() == kTemperatureOverUpperLimit) {
```
maxguy2001, line 49: This feels like it's beyond the scope of the function?

lib/sensors/thermister.hpp
```diff
@@ -0,0 +1,29 @@
...
+namespace hyped::sensors {
+
+// Skeleton code for the thermistor class
+// TODOLater: Implement this properly
+class Thermistor {
+ public:
+  static std::optional<Thermistor> create(core::ILogger &logger, std::shared_ptr<io::IAdc> adc);
+  ~Thermistor();
+
+  std::uint16_t getTemperature();
```
maxguy2001: You don't have a temperature to get...

lib/debug/repl.hpp
```diff
@@ -14,6 +14,7 @@
 #include <io/hardware_uart.hpp>
 #include <io/pwm.hpp>
 #include <sensors/accelerometer.hpp>
+#include <sensors/temperature.hpp>
```
maxguy2001, line 17: You also need to include cstdint and optional

lib/core/types.hpp
SnickeyX: I highly doubt that but will change

lib/sensors/i2c_mux.hpp
SnickeyX: fair will change that

lib/io/hardware_adc.cpp
```diff
@@ -54,7 +55,8 @@ std::optional<std::uint16_t> Adc::resetAndRead4(const int file_descriptor)
     return std::nullopt;
   }
   const int raw_voltage = std::atoi(read_buffer);
-  return static_cast<std::uint16_t>(raw_voltage);  // max value is 2^12-1 = 4095
+  // convert raw voltage to voltage between [0, 1.8] in V
```
maxguy2001, line 58: Do we not check the values of the raw voltage against the maximum allowed value?

maxguy2001: Also not sure if I'm missing something here but if the value is between 0 and 1.8, why do we return `uint16_t`?

maxguy2001 (changes requested):

lib/sensors/i2c_converter.hpp
SnickeyX: my b, this is meant to purly be skeleton code, will remove

lib/sensors/i2c_mux.hpp
SnickeyX: this is what electronics wanted, and double confirmed (i personally dont agree with this either but...)

lib/sensors/i2c_sensors.hpp
SnickeyX: yes, so they would be pre-configured and the mux itself would just accept a pointer to them

lib/debug/repl.hpp
SnickeyX: good point

lib/io/hardware_adc.cpp
SnickeyX: this is truly one of the casts of all time :00, will change asap - wrote this after a tiring testing sesh

lib/sensors/accelerometer.hpp
SnickeyX: did it for completeness sake since I wanted to mostly highlight what the return values mean

lib/sensors/temperature.cpp
SnickeyX: i understand that, however the alternative would be to do another i/o operation in `read` itself (i.e. checking the same register but for fatal errors), which I think is worse

lib/sensors/accelerometer.cpp
SnickeyX: good catch! will change

lib/sensors/accelerometer.hpp
maxguy2001: But if any of the return types or implementation changes, is anyone really going to think to update this comment given that it seems quite self explanatory and you can see the return type really quickly just by looking at the function definition...

lib/sensors/i2c_mux.hpp
maxguy2001: This is very upsetting news. Can we do more specific error handling anyways and then just given them that value at the end?

lib/sensors/accelerometer.hpp
SnickeyX: i see that for the @brief part but the @return surely needs to be clarified

lib/sensors/temperature.cpp
maxguy2001: Surely you could just implement a helper function or similar instead?

lib/sensors/accelerometer.cpp
```diff
@@ -2,7 +2,36 @@
...
+  }
+  const core::Result ctrl1_result
+    = i2c->writeByteToRegister(kDefaultAccelerometerAddress, kCtrl1Address, kCtrl1Value);
+  if (ctrl1_result == core::Result::kFailure) { return std::nullopt; };
+  const core::Result ctrl2_result
+    = i2c->writeByteToRegister(kDefaultAccelerometerAddress, kCtrl2Address, kCtrl2Value);
+  if (ctrl2_result == core::Result::kFailure) { return std::nullopt; };
+  const core::Result ctrl6_result
+    = i2c->writeByteToRegister(kDefaultAccelerometerAddress, kCtrl6Address, kCtrl6Value);
+  if (ctrl6_result == core::Result::kFailure) { return std::nullopt; };
```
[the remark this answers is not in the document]
SnickeyX: the log happens in getRawAcc so we good

lib/debug/repl.cpp
```diff
@@ -548,4 +541,71 @@ void Repl::addUartCommands(const std::uint8_t bus)
...
+  temperature_read_command.name = identifier.str();
+  std::stringstream description;
+  description << "Read temperature sensor 0x" << std::hex << static_cast<int>(device_address)
+              << " on "
+              << "I2C bus " << static_cast<int>(bus);
+  temperature_read_command.description = description.str();
+  temperature_read_command.handler     = [this, temperature, bus]() {
+    const auto value = temperature->read();
+    if (!value) {
+      logger_.log(core::LogLevel::kFatal, "Failed to read accelerometer from bus %d", bus);
```
HTTYDKing, line 602: Change the Failed to read accelerometer to temperature

HTTYDKing (changes requested):

lib/sensors/accelerometer.hpp
maxguy2001: Fair enough

lib/sensors/accelerometer.cpp
```diff
@@ -2,17 +2,83 @@
...
+  const auto device_id = i2c->readByte(device_address, kDeviceIdAddress);
+  if (!device_id) {
+    logger.log(core::LogLevel::kFatal, "Failed to read the accelerometer device ID");
+    return std::nullopt;
+  }
+  if (*device_id != kExpectedDeviceIdValue) {
+    logger.log(core::LogLevel::kFatal, "Failure, mismatched device ID for accelerometer");
+    return std::nullopt;
+  }
+  const core::Result ctrl1_result
```
TomLonergan03, line 20: should all these results be autos?

lib/sensors/accelerometer.cpp
```diff
@@ -2,17 +2,83 @@
...
+  if (!result_z) { return std::nullopt; }
+  const std::int32_t z_acceleration = getAccelerationFromRawValue(*result_z);
+  core::WallClock time;
+  const std::optional<core::RawAccelerationData> acceleration_3_axis{
+    std::in_place, x_acceleration, y_acceleration, z_acceleration, time.now()};
+  logger_.log(core::LogLevel::kDebug, "Successfully read accelerometer data");
+  return acceleration_3_axis;
+}
+
 std::uint8_t Accelerometer::getChannel() const
```
TomLonergan03, line 82: when would you need to get the channel of an accelerometer?

also for my info what is the channel? is it the i2c bus that the accelerometer is on?

lib/sensors/i2c_converter.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <cstdint>
+#include <optional>
+
+#include "io/i2c.hpp"
```
TomLonergan03, line 6: `<io/i2c.hpp>` as it's not a relative path

lib/sensors/i2c_mux.hpp
TomLonergan03, line 15: percentage

lib/sensors/temperature.hpp
```diff
@@ -11,32 +11,55 @@
...
 namespace hyped::sensors {
 // Values and register names from data sheet
-static constexpr std::uint8_t kTemperatureDefaultAddress = 0x38;
-static constexpr std::uint8_t kCtrl                      = 0x04;
-static constexpr std::uint8_t kDataTemperatureHigh       = 0x07;
-static constexpr std::uint8_t kDataTemperatureLow        = 0x06;
-static constexpr std::uint8_t kStatus                    = 0x05;
+constexpr std::uint8_t kTemperatureDefaultAddress = 0x38;
+// Adding this since a singluar temperature sensor may be on a different address
+constexpr std::uint8_t kTemperatureAlternativeAddress = 0x3F;
```
TomLonergan03, line 16: what is this? will sensors only have one of these 2 addresses? or are these the most common ones?

TomLonergan03 (changes requested):

lib/sensors/accelerometer.cpp
SnickeyX: Mux

lib/sensors/temperature.hpp
SnickeyX: Only one of these two

lib/sensors/accelerometer.cpp
TomLonergan03: oh ok makes sense

lib/sensors/temperature.hpp
TomLonergan03: make that clear in a comment i think

also singular not sigluar

SnickeyX: going to make it an enum so you can only ever put these two addresses

SnickeyX: on second thought, making the addresses come from an enum actually seems like a much worse solution considering the current state of the repl where we pass in an address manually (might need to change this to ask whether a default or alternative address is going to be read from for the temperature and accelerometer sensor)

lib/debug/repl.cpp
```diff
@@ -548,4 +541,71 @@ void Repl::addUartCommands(const std::uint8_t bus)
...
+  description << "Read accelerometer sensor 0x" << std::hex << static_cast<int>(device_address)
+              << " on "
+              << "I2C bus " << static_cast<int>(bus);
+  accelerometer_read_command.description = description.str();
+  accelerometer_read_command.handler     = [this, accelerometer, bus]() {
+    const auto value = accelerometer->read();
+    if (!value) {
+      logger_.log(core::LogLevel::kFatal, "Failed to read the accelerometer from bus %d", bus);
+    } else {
+      const core::RawAccelerationData accelerometer_result = *value;
```
TomLonergan03, line 569: `auto`

lib/debug/repl.cpp
```diff
@@ -548,4 +541,71 @@ void Repl::addUartCommands(const std::uint8_t bus)
...
+  description << "Read temperature sensor 0x" << std::hex << static_cast<int>(device_address)
+              << " on "
+              << "I2C bus " << static_cast<int>(bus);
+  temperature_read_command.description = description.str();
+  temperature_read_command.handler     = [this, temperature, bus]() {
+    const auto value = temperature->read();
+    if (!value) {
+      logger_.log(core::LogLevel::kFatal, "Failed to read the temperature sensor from bus %d", bus);
+    } else {
+      const std::int16_t temperature_result = *value;
```
TomLonergan03, line 604: `auto`

lib/sensors/accelerometer.cpp
```diff
@@ -2,17 +2,80 @@
...
+  return core::Result::kSuccess;
+}
+
+// TODOLater: current settings of the accelerometer make it read in +-16g but with high noise. Check
+// whether other configuration could be better
+std::optional<core::RawAccelerationData> Accelerometer::read()
+{
+  const auto result_x = getRawAcceleration(core::Axis::kX);
+  if (!result_x) { return std::nullopt; }
+  const std::int32_t x_acceleration = getAccelerationFromRawValue(*result_x);
```
TomLonergan03, line 65: `auto`s for these

lib/sensors/temperature.cpp
```diff
@@ -2,33 +2,51 @@
 
 namespace hyped::sensors {
 
+std::optional<Temperature> Temperature::create(core::ILogger &logger,
+                                               std::shared_ptr<io::II2c> i2c,
+                                               const std::uint8_t channel,
+                                               const std::uint8_t device_address)
+{
+  const core::Result write_result
```
TomLonergan03, line 10: `auto`

lib/sensors/temperature.cpp
```diff
@@ -2,33 +2,51 @@
...
-  // TODOLater: change this from a fatal error to a debug and re-prompt read (with ROS)
   if (status_check_result.value() == kBusy) {
-    logger_.log(core::LogLevel::kFatal,
+    logger_.log(core::LogLevel::kWarn,
                 "Failed to read, temperature sensor is not ready to be read at channel %d",
                 channel_);
-    return std::nullopt;
+    return core::Result::kFailure;
   }
   if (status_check_result.value() == kTemperatureOverUpperLimit) {
```
TomLonergan03, line 51: deref optional with `*status_check_result`

lib/sensors/temperature.cpp
```diff
@@ -2,33 +2,51 @@
...
 {
-  const auto status_check_result = i2c_->readByte(kTemperatureDefaultAddress, kStatus);
+  const auto status_check_result = i2c_->readByte(device_address_, kStatus);
   if (!status_check_result) {
     logger_.log(
       core::LogLevel::kFatal, "Failed to read temperature sensor status at channel %d", channel_);
     return std::nullopt;
   }
-  // TODOLater: change this from a fatal error to a debug and re-prompt read (with ROS)
   if (status_check_result.value() == kBusy) {
```
TomLonergan03, line 45: deref optional with `*`

TomLonergan03 (changes requested):

lib/debug/repl.cpp
SnickeyX: i think we need to add this to our coding guide somewhere, if you can @TomLonergan03 could you add our actual coding practices in a wiki page? i agree with `auto`-ing optionals and enums, but not uints or other datatypes. Wdyt @kshxtij ?

lib/sensors/temperature.cpp
SnickeyX: since this entire function deals with just the status register, I think splitting it would be redundant. I do think that maybe its name could be better? Maybe like `checkStatus` ?

lib/debug/repl.cpp
kshxtij: I agree, we definitely shouldn't auto uints cause @maxguy2001 would have a stroke

TomLonergan03: Makes sense, last I heard `auto`  was for anything that was returned from a function but I could well be wrong.

And derefing optionals is kinda borderline for that still

TomLonergan03: actually, there's quite a few edge cases, so it would be worth working out an explicit policy for `auto`

SnickeyX: dereferecing optionals is purely for speed but it does bypass exception checking (so we handle that). at this point I think it may be a good idea to have a HYPED style-guide, especially to help with next year @TomLonergan03

TomLonergan03: using `auto` for derefing optionals is what i mean is borderline, i'll try to write something up but we should discuss it on monday

SnickeyX: though please do review my PR

SnickeyX:
#47 covers virtual tags being added to temp + acc so wont do it here

lib/io/hardware_uart.hpp
```diff
@@ -69,8 +69,8 @@ class Uart : public IUart {
   Uart(core::ILogger &logger, const int file_descriptor);
   ~Uart();
 
-  core::Result sendBytes(const char *tx, const std::uint8_t length);
-  core::Result readBytes(unsigned char *rx, const std::uint8_t length);
+  virtual core::Result sendBytes(const char *tx, const std::uint8_t length);
+  virtual core::Result readBytes(unsigned char *rx, const std::uint8_t length);
```
TomLonergan03, line 73: Should these be `virtual ..... () = 0;`? Or am I forgetting where that should go

TomLonergan03 (changes requested):

lib/io/hardware_uart.hpp
SnickeyX: that syntax defines pure virtual functions, i.e. ones that must be derived from, but these are not

lib/debug/repl.cpp
maxguy2001: > I agree, we definitely shouldn't auto uints cause @maxguy2001 would have a stroke

CaN ThIs oNLy EvEr bE NeGaTivE???

maxguy2001: Yeah there seems to be a few areas where we disagree (including autos, initializing class members and where we store certain objects/definitions). While they don't usually impact the actual functionality of the  code too heavily, it's worth picking one method and sticking with it across the codebase for the sake of consistency.

lib/io/hardware_uart.hpp
TomLonergan03: Oh yea me dumb

lib/sensors/temperature.cpp
maxguy2001: Yeah this sounds good. Just something that implies the functionality more intuitively.

lib/sensors/i2c_mux.hpp
```diff
@@ -0,0 +1,133 @@
+#pragma once
+
+#include "i2c_sensors.hpp"
+
+#include <array>
+#include <memory>
+
```
maxguy2001, line 7: Check your includes here. Missing at least cstdint, optional and core/types.hpp.

lib/sensors/temperature.cpp
```diff
@@ -2,81 +2,92 @@
...
   }
-  // TODOLater: change this from a fatal error to a debug and re-prompt read (with ROS)
-  if (status_check_result.value() == kBusy) {
-    logger_.log(core::LogLevel::kFatal,
+  if (*status_check_result == kBusy) {
+    logger_.log(core::LogLevel::kWarn,
                 "Failed to read, temperature sensor is not ready to be read at channel %d",
                 channel_);
-    return std::nullopt;
+    return core::Result::kFailure;
```
maxguy2001, line 53: Need to include core/types.hpp?

lib/sensors/temperature.cpp
```diff
@@ -2,81 +2,92 @@
...
-  const auto temperature_low_byte = i2c_->readByte(kTemperatureDefaultAddress, kDataTemperatureLow);
+  const auto temperature_low_byte = i2c_->readByte(device_address_, kDataTemperatureLow);
   if (!temperature_low_byte) {
     logger_.log(
       core::LogLevel::kFatal, "Failed to read low byte for temperature at channel %d", channel_);
     return std::nullopt;
   }
-  const std::int16_t temperature
-    = ((temperature_high_byte.value() << 8) | temperature_low_byte.value());
+  const std::int16_t temperature = ((*temperature_high_byte << 8) | *temperature_low_byte);
```
maxguy2001, line 84: not `uint16_t`?

maxguy2001: Given that it's just bytes at this point?

lib/sensors/temperature.cpp
```diff
@@ -2,81 +2,92 @@
...
   logger_.log(
     core::LogLevel::kDebug, "Successfully read from temperature sensor at channel %d", channel_);
   // Scaling temperature as per the datasheet
   return static_cast<std::int16_t>(temperature * kTemperatureScaleFactor);
```
maxguy2001, line 88: temperature is already int16_t, think you only need to cast scaling factor.

lib/utils/dummy_adc.hpp
```diff
@@ -9,7 +9,7 @@ class DummyAdc : public io::IAdc {
   DummyAdc()  = default;
   ~DummyAdc() = default;
 
-  std::optional<std::uint16_t> readValue();
+  virtual std::optional<core::Float> readValue();
```
maxguy2001, line 12: include core types

lib/utils/dummy_uart.hpp
```diff
@@ -15,8 +15,8 @@ class DummyUart : public io::IUart {
   DummyUart()  = default;
   ~DummyUart() = default;
 
-  core::Result sendBytes(const char *tx, const std::uint8_t length);
-  core::Result readBytes(unsigned char *rx, const std::uint8_t length);
+  virtual core::Result sendBytes(const char *tx, const std::uint8_t length);
```
maxguy2001, line 18: include core types if using `Result`?

maxguy2001 (changes requested):

lib/debug/repl.hpp
```diff
@@ -1,8 +1,11 @@
 #pragma once
 
+#include <cstdint>
 #include <functional>
 #include <iostream>
 #include <map>
+#include <memory>
```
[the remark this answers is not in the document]
SnickeyX: dont think vector is, removing

lib/io/hardware_adc.cpp
```diff
@@ -5,8 +5,13 @@
 
 namespace hyped::io {
 
-std::optional<std::shared_ptr<Adc>> Adc::create(core::ILogger &logger, const std::uint8_t pin)
+std::optional<std::shared_ptr<HardwareAdc>> HardwareAdc::create(core::ILogger &logger,
+                                                                const std::uint8_t pin)
 {
+  if (pin < 0 || pin > 6) {
```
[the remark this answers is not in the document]
SnickeyX: ive mentioned in the .hpp that there are 7 pins (0-indexed), but ill mention it there too

lib/io/hardware_adc.cpp
```diff
@@ -5,8 +5,13 @@
...
+    logger.log(core::LogLevel::kFatal, "Failed to create HardwareAdc object: invalid pin");
```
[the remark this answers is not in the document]
SnickeyX: sure

lib/io/hardware_adc.cpp
```diff
@@ -54,7 +58,8 @@ std::optional<std::uint16_t> Adc::resetAndRead4(const int file_descriptor)
     return std::nullopt;
   }
   const int raw_voltage = std::atoi(read_buffer);
```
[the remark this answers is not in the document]
SnickeyX: It is strictly positive, but std::atoi returns an int. While I agree with your point, I think static_casting it to a uint just to in the next line static cast it to a float is a bit unnecessary

lib/sensors/i2c_mux.hpp
SnickeyX: coming from i2c_sensors - but yeah will add jic

lib/sensors/temperature.cpp
SnickeyX: coming from i2c_sensors - but this time I think its fine to leave it out since temperature is directlty derived from `II2cMuxSensor`  and they must have some basic overlapping imports

lib/sensors/temperature.cpp
SnickeyX: nope, temperature can be negative (though we should never have that occur, but if we make it uint we cant check for it)

lib/sensors/temperature.cpp
SnickeyX: scale factor is a float so we need to static_cast the product else itll be a float

lib/utils/dummy_adc.hpp
SnickeyX: same justification as before, imported in its base class

lib/utils/dummy_uart.hpp
SnickeyX: same as ^^

lib/sensors/temperature.cpp
SnickeyX: image
actually this is quite interesting coz it makes more sense to go with the more limiting variables (in terms of space), but c++ tries to go with more information over less.

lib/io/hardware_adc.cpp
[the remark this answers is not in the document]
maxguy2001: Yeah I agree. Int is fine.

lib/sensors/temperature.cpp
maxguy2001: I see where you're coming from but also I feel like it's worth including what you use to maintain modularity, for best practice and to make clear where everything comes from (i.e. a new software member might find it easier to understand what is going on if they can see where everything comes from?). Maybe another point to decide in hyped best practice guide?

lib/sensors/temperature.cpp
maxguy2001: I'm chatting nonsense, pls ignore me!

lib/io/hardware_spi.cpp
```diff
@@ -118,19 +118,19 @@ const char *HardwareSpi::getSpiBusAddress(const SpiBus bus)
...
-      return 500000;
-    case Clock::k1MHz:
-      return 1000000;
-    case Clock::k4MHz:
-      return 4000000;
-    case Clock::k16MHz:
-      return 16000000;
-    case Clock::k20MHz:
-      return 20000000;
+    case SpiClock::k500KHz:
```
maxguy2001, line 124: I know this is one of my more pedantic comments (and that's really saying something) bit will there ever be a case where the number of `SpiClock` members changes? I can currently deal with the number of cases in this switch statement but if this has score to expand then maybe a different method of handling this is more appropriate?

lib/sensors/accelerometer.cpp
```diff
@@ -2,17 +2,85 @@
...
+  const auto result_y               = getRawAcceleration(core::Axis::kY);
+  if (!result_y) { return std::nullopt; }
+  const std::int32_t y_acceleration = getAccelerationFromRawValue(*result_y);
+  const auto result_z               = getRawAcceleration(core::Axis::kZ);
+  if (!result_z) { return std::nullopt; }
+  const std::int32_t z_acceleration = getAccelerationFromRawValue(*result_z);
+  core::WallClock time;
```
maxguy2001, line 77: Should this not also take an instance of `core::ITimeSource`?

maxguy2001: To make the code more testable?

lib/sensors/accelerometer.cpp
```diff
@@ -38,6 +106,12 @@ std::optional<std::int16_t> Accelerometer::getRawAcceleration(const core::Axis a
   return static_cast<std::int16_t>((*high_byte << 8) | *low_byte);
 }
 
+std::int32_t Accelerometer::getAccelerationFromRawValue(const std::int16_t rawAcceleration)
```
maxguy2001, line 109: in `Accelerometer::read`, the values for acceleration on each axis are already `int32_t`?

lib/sensors/accelerometer.hpp
```diff
@@ -50,24 +49,43 @@ constexpr std::uint8_t kExpectedDeviceIdValue = 0x44;
...
+   * @brief Checks if the accelerometer is ready to be read
+   * @return kSuccess if the sensor is ready to be read,
+   *         kFailure if the sensor is not ready to be read,
+   *         nullopt if there was an error reading the sensor
+   */
+  std::optional<core::Result> isValueReady();
+
+  /**
```
maxguy2001, line 66: Necessary?

lib/sensors/i2c_converter.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+namespace hyped::sensors {
+
+// TODOLater: Implement this properly
+/**
+ * @brief ADC (Texas Instruments) to be used for the hall-effect sensors (Farnell)
+ * @details This class is used to convert the analog signal from the hall-effect sensors to a
+ * digital signal over I2c
+ */
+class I2cConverter {
```
maxguy2001, line 17: From description is there not a better name for this? I know it's a little bit pedantic but the name of this doesn't give you an intuitive sense of it's purpose in my opinion? As a non-sensors person though, I may very well be wrong...

lib/sensors/i2c_mux.hpp
```diff
@@ -0,0 +1,136 @@
...
+#include <optional>
+
+#include <core/logger.hpp>
+#include <core/types.hpp>
+#include <io/i2c.hpp>
+
+namespace hyped::sensors {
+
+constexpr std::uint8_t kDefaultMuxAddress = 0x70;
+constexpr std::uint8_t kMaxNumChannels    = 8;
```
maxguy2001, line 17: Given that this is in `hyped::sensors`, maybe `kMaxNumMuxChannels` for clarity?

lib/sensors/i2c_mux.hpp
```diff
@@ -0,0 +1,136 @@
...
+std::optional<std::array<T, N>> I2cMux<T, N>::readAllChannels()
+{
+  // Zero-initialize the array
+  std::array<T, N> mux_data{};
+  std::uint8_t num_unusable_sensors = 0;
+  for (std::uint8_t i = 0; i < N; ++i) {
+    const auto &sensor         = sensors_.at(i);
+    const std::uint8_t channel = sensor->getChannel();
+    // First ensure correct channel is selected
+    const auto channel_select_result = selectChannel(channel);
```
maxguy2001, line 77: idk if this is outside the scope of this pr but as I've read a lot of sensors code today, would making the `selectChannel` result optional and checking if `!channel_select_result` make more sense?

lib/utils/dummy_adc.hpp
maxguy2001: I guess leave it for now and discuss monday?

maxguy2001 (changes requested):

lib/sensors/i2c_mux.hpp
maxguy2001: It just seems like this name is maybe a bit too general at his scope?

lib/io/hardware_spi.cpp
SnickeyX: no I don't think so? though I still haven't tested spi this year, last year these values worked just fine

lib/sensors/accelerometer.cpp
SnickeyX: actually, this is already being fixed in #47 so might leave it as is here, the implementation changes quite a bit

lib/sensors/accelerometer.cpp
SnickeyX: don't get what's wrong? might be missing something

lib/sensors/i2c_converter.hpp
SnickeyX: it could be `AnalogueToI2cConverter` but since we only use one, I shortened it (there really isn't much more to it, it literally just converts) (but also if we do come up with a better name, it could be done in the PR dealing with this sensor)

lib/sensors/i2c_mux.hpp
SnickeyX: i agree, missed this

lib/sensors/i2c_mux.hpp
SnickeyX: no since we know every sensor on the mux must be initialised with a channel number (it cant ever be that a sensor doesn't have a channel number on the mux)

lib/sensors/temperature.cpp
SnickeyX: that's fair, this is exactly why a guide is important

lib/sensors/accelerometer.hpp
```diff
@@ -50,24 +49,40 @@ constexpr std::uint8_t kExpectedDeviceIdValue = 0x44;
 
 class Accelerometer : public II2cMuxSensor<core::RawAccelerationData> {
  public:
-  Accelerometer(core::ILogger &logger, io::II2c &i2c, const std::uint8_t channel);
+  static std::optional<Accelerometer> create(core::ILogger &logger,
+                                             std::shared_ptr<io::II2c> i2c,
+                                             const std::uint8_t channel,
+                                             const std::uint8_t device_address);
```
Platiniom64, line 55: I think that here we want to make the "device address" an optional parameter, since in the implementation used oustide of the repl we don't care about the address.

Platiniom64 (changes requested):

lib/sensors/accelerometer.hpp
SnickeyX: We are changing this so now we must be explicit about all parameters, i.e. no default values

lib/sensors/accelerometer.cpp
maxguy2001: Ignore me, got my types mixed up :(

BenET01 (approved):

maxguy2001 (approved):
Minor new style guide changes to be implemented but otherwise all good.

Platiniom64 (approved):

lib/io/hardware_uart.hpp
```diff
@@ -51,26 +53,25 @@ enum class BaudRate {
...
-    const BaudRate baud_rate,  // TODOLater: Figure out a default for this by testing
-    const BitsPerByte bits_per_byte = BitsPerByte::k8);
+    const UartBaudRate baud_rate,  // TODOLater: Figure out a default for this by testing
+    const UartBitsPerByte bits_per_byte);
   Uart(core::ILogger &logger, const int file_descriptor);
   ~Uart();
 
-  core::Result sendBytes(const char *tx, const std::uint8_t length);
-  core::Result readBytes(unsigned char *rx, const std::uint8_t length);
+  virtual core::Result sendBytes(const char *tx, const std::uint8_t length);
```
TomLonergan03, line 73: I assume these pointers are C library moments?

TomLonergan03 (commented):
Looks good, will check includes once I'm at my pc later tonight

lib/io/hardware_uart.hpp
TomLonergan03: finally getting back to my pc it looks like linux do be like that

TomLonergan03 (approved):
looks good

i've not found any missing includes but looking at includes is making my eyes water so I might have missed something

SnickeyX:
in line with what we discussed yesterday, i've made register addresses + constants for adc a private static constexpr since it should never be used anywhere other than the class -> this means confusion between possible overlapping register names between sensors can be avoided entirely @TomLonergan03 @maxguy2001

kshxtij (approved):

PR #102, opened 2023-03-06
Ss brake clean

Knhwong:
Same as brake, makes things cleaner.

### DIFF
--- a/lib/brakes/brakes.cpp
+++ b/lib/brakes/brakes.cpp
@@ -1,6 +1,6 @@
 #include "brakes.hpp"
 
-brakes::Brakes(const std::uint8_t pin, const hyped::io::IGpio gpio) : pin_(pin)
+brakes::Brakes(const std::uint8_t pin, const io::IGpio gpio) : pin_(pin)
 {
   writer = gpio.getWriter(pin);
 }
--- a/lib/brakes/brakes.cpp
+++ b/lib/brakes/brakes.cpp
@@ -1,0 +1,16 @@
+#include "brakes.hpp"
+
+brakes::Brakes(const std::uint8_t pin, const hyped::io::IGpio gpio) : pin_(pin)
+{
+  writer = gpio.getWriter(pin);
+}
+
+void brakes::stop()
+{
+  writer.write(core::DigitalSignal::kHigh);
+}
+
+void brakes::release()
+{
+  writer.write(core::DigitalSignal::kLow);
+}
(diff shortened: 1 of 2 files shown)


lib/brakes/brakes.hpp
```diff
@@ -0,0 +1,17 @@
...
+
+class Brakes {
+ public:
+  Brakes(const std::uint8_t pin_, const io::IGpio gpio);
+  void stop();
+  void release();
+
+ private:
+  const std::uint8_t pin_;
+  io::IGpioWriter writer;
```
maxguy2001, line 16: `writer_`

maxguy2001 (changes requested):

lib/brakes/brakes.cpp
```diff
@@ -0,0 +1,16 @@
+#include "brakes.hpp"
+
+brakes::Brakes(const std::uint8_t pin, const io::IGpio gpio) : pin_(pin)
```
TomLonergan03, line 3: `Brakes::Brakes` and same for all `brakes::` should be `Brakes::`

TomLonergan03 (changes requested):

lib/brakes/brakes.cpp
Knhwong: done

lib/brakes/brakes.hpp
```diff
@@ -0,0 +1,18 @@
...
+// This class deals with stopping and release brakes of the pod.
+namespace hyped::brakes {
+class Brakes {
+ public:
+  Brakes(const std::uint8_t pin, const io::IGpio gpio);
+  void stop();
+  void release();
+
+ private:
+  const std::uint8_t pin_;
```
maxguy2001, line 15: This isn't used anywhere?

maxguy2001: It's explicitly used in constructor for writer but nowhere else?

lib/brakes/brakes.hpp
```diff
@@ -0,0 +1,18 @@
...
+
+#include <core/types.hpp>
+#include <io/gpio.hpp>
+
+// This class deals with stopping and release brakes of the pod.
+namespace hyped::brakes {
+class Brakes {
+ public:
+  Brakes(const std::uint8_t pin, const io::IGpio gpio);
+  void stop();
```
maxguy2001, line 11: Document these?

maxguy2001 (changes requested):

PR #106, opened 2023-03-21
MTC - CAN Extended support

TomLonergan03:
I believe socket automatically deals with extended CAN IDs, all that needs to change is to OR the ID in the sent frame with the CAN_EFF_FLAG.
Added the CAN_EFF_FLAG define for non linux systems

TODO:

- [x] Test socket actually handles it


TomLonergan03:
Tested and works

kshxtij (approved):

PR #107, opened 2023-03-23
Nav imu kalman integration

Fayz-S:
Beginning of kalman integration: transition matrix function needs to be moved to new class


lib/navigation/control/consts.hpp
```diff
@@ -0,0 +1,62 @@
...
+  = Eigen::Matrix<core::Float, extended_dimension, 1>;  //[Jerk, Snap, Crackle, Pop, ...]
+using JacobianMatrix
+  = Eigen::Matrix<core::Float, extended_dimension, state_dimension>;  // higher order derivatives
+                                                                      // for propogation
+
+static constexpr core::Float kTrackLength       = 100.0;  // m
+static constexpr core::Float kBrakingDistance   = 20.0;   // m TODOLater:check!
+static constexpr core::Float kPi                = 3.14159265359;
+static constexpr core::Float kWheelCicumference = kPi * 0.1;  // m TODOLater: check!
+static constexpr core::Float kStripeDistance    = 10.0;       // m TODOLater:check!
```
SnickeyX, line 37: if this is the keyence stripe, i think its a 6m distance from what Greg wrote in slack

lib/navigation/control/crosscheck.hpp
```diff
@@ -44,7 +48,12 @@ class Crosscheck {
...
    */
-  SensorChecks checkEncooderKeyence(const core::EncoderData encoder_data,
-                                    const core::KeyenceData keyence_data);
+  SensorChecks checkEncoderKeyence(const core::Float encoder_displacement,
+                                   const core::Float keyence_displacement);
+
+  core::ILogger &logger_;
+  const core::ITimeSource &time_;
+  static constexpr std::uint8_t kMaxAllowedAccelerometerEncoderDifference = 5;
+  static constexpr std::uint8_t kMaxAllowedKeyenceEncoderDifference       = 12;
```
SnickeyX, line 57: do we have any reasoning for these thresholds? (Just to avoid them feeling like magic)

lib/navigation/control/navigator.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+  return trajectory_;
+}
+
+void Navigator::keyenceUpdate(const core::KeyenceData &keyence_data)
+{
+  // Check keyence strictly increasing
+  if (keyence_data.at(0) < previous_keyence_reading_.at(0)) {
+    logger_.log(core::LogLevel::kFatal, "Keyence data is decreasing");
+  }
```
SnickeyX, line 52: do we plan to add anything other than a logger to indicate this failure?

lib/navigation/filtering/accelerometer_kalman.cpp
```diff
@@ -0,0 +1,104 @@
+/*
```
SnickeyX, line 1: assuming this file is a WIP?

lib/navigation/filtering/extended_kalman_filter.hpp
```diff
@@ -0,0 +1,65 @@
...
+                       const StateVector initial_state,
+                       const ErrorCovarianceMatrix initial_error_covariance)
+      : time_source_(time_source),
+        last_update_time_(time_source->now()),
+        state_estimate_(initial_state),
+        error_covariance_(initial_error_covariance)
+  {
+    static_assert(state_dimension > 0);
+    static_assert(measurement_dimension > 0);
+    static_assert(extended_dimension > 0);
```
SnickeyX, line 29: love to see some asserts!

SnickeyX (changes requested):
minor things. i trust you with the stats o7

lib/navigation/control/navigator.cpp
maxguy2001: such as?

SnickeyX: core::Result?

lib/navigation/control/crosscheck.hpp
SnickeyX: Yeah that would be better. Can't really judge the values lol

lib/navigation/control/navigator.cpp
maxguy2001: What does that do differently than a logged failure?

lib/navigation/control/crosscheck.cpp
```diff
@@ -0,0 +1,62 @@
...
+  SensorChecks encoder_keyence_check
+    = checkEncoderKeyence(encoder_displacement, keyence_displacement);
+  if (encoder_keyence_check == SensorChecks::kUnacceptable) {
+    logger_.log(core::LogLevel::kFatal,
+                "Large disagreement between encoders and keyence. Trajectory cannot be accurately "
+                "calculated.");
+    return SensorChecks::kUnacceptable;
+  }
+
+  logger_.log(core::LogLevel::kInfo, "Trajectory values sucessfully verified");
```
BenET01, line 37: "successfully"

lib/navigation/control/crosscheck.cpp
```diff
@@ -0,0 +1,62 @@
...
+  return SensorChecks::kAcceptable;
+}
+
+SensorChecks Crosscheck::checkEncoderAccelerometer(const core::Float acceleration_displacement,
+                                                   const core::Float encoder_displacement)
+{
+  // check values are within tolerance
+  const core::Float difference = acceleration_displacement - encoder_displacement;
+  if (std::abs(difference) > kMaxAllowedAccelerometerEncoderDifference) {
+    return SensorChecks::kUnacceptable;
```
BenET01, line 48: do we not want to log this?

lib/navigation/control/crosscheck.cpp
```diff
@@ -0,0 +1,62 @@
...
+  }
+  return SensorChecks::kAcceptable;
+}
+
+SensorChecks Crosscheck::checkEncoderKeyence(const core::Float encoder_displacement,
+                                             const core::Float keyence_displacement)
+{
+  const core::Float difference = encoder_displacement - keyence_displacement;
+  if (std::abs(difference) > kMaxAllowedKeyenceEncoderDifference) {
+    return SensorChecks::kUnacceptable;
```
BenET01, line 58: same as above

lib/utils/kalman_higher_derivatives.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+  // update acceleration and timestamps vectors
+  acceleration_values_.pop_back();
+  acceleration_values_.insert(acceleration_values_.begin(), acceleration);
+
+  timestamps_.pop_back();
+  timestamps_.insert(timestamps_.begin(), timestamp);
+
+  // calculate new jerk value from new data and update jerk sum
+  sum_jerk_ -= jerk_values_.at(jerk_values_.size() + 1);
```
BenET01, line 37: probably stupid but not too sure why we do this?

BenET01 (changes requested):

lib/navigation/control/crosscheck.cpp
maxguy2001: but yeah I'll fix :(

lib/navigation/control/crosscheck.cpp
maxguy2001: yeah I'll log that too

lib/utils/kalman_higher_derivatives.cpp
maxguy2001: Instead of summing the whole array each time the function is called (which is $O(n)$), we store the total of the array as a data class member, subtract the element at the back of the array (that we are removing) and add the new element that we're inserting at the front of the array to the total (constant time).

lib/navigation/control/navigator.cpp
TomLonergan03: logging doesn't trigger any behaviour, if we need to react to this event then we must return something indicating an error

lib/navigation/filtering/running_means_filter.cpp
```diff
@@ -0,0 +1,25 @@
+#include "running_means_filter.hpp"
+
+namespace hyped::navigation {
+
+RunningMeansFilter::RunningMeansFilter(core::ILogger &logger, const core::ITimeSource &time)
+    : logger_(logger),
+      time_(time)
+{
+  acceleration_values_.resize(kNumDataPoints);
+  std::fill(acceleration_values_.begin(), acceleration_values_.end(), 0.0F);
```
TomLonergan03, line 10: why is this done this way? seems weird to resize

test/navigation/kalman_filter.cpp
```diff
@@ -9,14 +10,16 @@ TEST(KalmanFilter, construction)
...
+  navigation::ErrorCovarianceMatrix initial_error_covariance;
   KalmanFilter kalman_filter(manual_time, initial_state, initial_error_covariance);
+  /*
   kalman_filter.filter(KalmanFilter::StateTransitionMatrix(),
                        KalmanFilter::StateTransitionCovarianceMatrix(),
                        KalmanFilter::MeasurementMatrix(),
                        KalmanFilter::MeasurementNoiseCovarianceMatrix(),
                        KalmanFilter::StateVector());
+ */
 }
```
TomLonergan03, line 23: commented out code, very cringe

TomLonergan03 (changes requested):
i don't know stats or maths, am trusting that it is right and don't see any obvious issues with implementation

otherwise looks good.

(also 2am i am going insane)

lib/navigation/control/consts.hpp
```diff
@@ -0,0 +1,69 @@
...
+using JacobianMatrix
+  = Eigen::Matrix<core::Float, extended_dimension, state_dimension>;  // higher order derivatives
+                                                                      // for propogation
+
+static constexpr core::Float kTrackLength       = 100.0;  // m
+static constexpr core::Float kBrakingDistance   = 20.0;   // m TODOLater:check!
+static constexpr core::Float kPi                = 3.14159265359;
+static constexpr core::Float kWheelCicumference = kPi * 0.1;  // m TODOLater: check!
+static constexpr core::Float kStripeDistance    = 6.0;
+static constexpr bool is_keyence_active         = true;
```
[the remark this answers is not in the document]
TomLonergan03: also probably `kIsKeyenceActive`

lib/navigation/control/navigator.cpp
maxguy2001: Ok. I'll change the return types of the main navigation interfaces then

lib/navigation/filtering/accelerometer_kalman.cpp
maxguy2001: Yeah this file is either getting fixed or deleted for now, still TBD

lib/navigation/control/navigator.cpp
maxguy2001: Ok. I'll change the return types of the main navigation interfaces then

lib/navigation/filtering/kalman_filter.cpp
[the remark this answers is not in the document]
maxguy2001: Complier was being difficult for some reason. Can't remember why but this was the very stupid solution.

lib/navigation/filtering/kalman_filter.hpp
```diff
@@ -4,24 +4,30 @@
...
     = Eigen::Matrix<core::Float, state_dimension, state_dimension>;
   using ErrorCovarianceMatrix = Eigen::Matrix<core::Float, state_dimension, state_dimension>;
-  using MeasurementMatrix     = Eigen::Matrix<core::Float, measurement_dimension, state_dimension>;
-  using MeasurementVector     = Eigen::Matrix<core::Float, measurement_dimension, 1>;
+  using MeasurementMatrix
+    = Eigen::Matrix<core::Float, measurement_dimension, state_dimension>;          //[1, 0, 0]
+  using MeasurementVector = Eigen::Matrix<core::Float, measurement_dimension, 1>;  //[acc_val, 0, 0]
   using MeasurementNoiseCovarianceMatrix
     = Eigen::Matrix<core::Float, measurement_dimension, measurement_dimension>;
+  */
```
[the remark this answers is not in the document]
maxguy2001: Yeah I put a comment to but forgot lol. Will wipe this section

lib/navigation/filtering/running_means_filter.cpp
maxguy2001: I mean it was this or write a loop? Is  there a better way of doing it?

test/navigation/kalman_filter.cpp
maxguy2001: :(

lib/navigation/filtering/running_means_filter.cpp
TomLonergan03: looking at this awake it makes sense now, you need a circular buffer and this makes sense to initialise it to a specific size.
You can do `acceleration_values_ = std::vector<core::Float>(length,initial_value);` but idk if that would be better form
Alternatively, if you do `acceleration_values_(length, initial_value)` in the initialiser list that also correctly initialises. I think this is the one I like most (but it is a bit more obtuse) but am open to arguments and am aware this is meaningless nitpicking

maxguy2001: I don't see anything directly wrong with current implementation? It just needs to be initialized to a certain size on class construction which is what it does?

maxguy2001 (approved):
Pending mac build fix

TomLonergan03 (approved):
fix linter + mac build and is good to go

BenET01 (approved):
pending mac build

lib/utils/kalman_higher_derivatives.cpp
Fayz-S, line 37: should this be `+=`?

lib/navigation/control/consts.hpp
```diff
@@ -0,0 +1,76 @@
...
+using StateTransitionCovarianceMatrix
+  = Eigen::Matrix<core::Float, state_dimension, state_dimension>;
+using ErrorCovarianceMatrix = Eigen::Matrix<core::Float, state_dimension, state_dimension>;
+using MeasurementMatrix
+  = Eigen::Matrix<core::Float, measurement_dimension, state_dimension>;  //[[1, 0, 0], [0, 0, 0],
+                                                                         //[0, 0, 0]]
+using MeasurementVector = Eigen::Matrix<core::Float, measurement_dimension, 1>;  //[acc_val, 0, 0]
+using MeasurementNoiseCovarianceMatrix
+  = Eigen::Matrix<core::Float, measurement_dimension, measurement_dimension>;
+using ExtendedStateVector
```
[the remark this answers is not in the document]
maxguy2001: Don't think so. @BenET01 should know better but I believe this vector contains *only* the extended states?

lib/navigation/control/consts.hpp
```diff
@@ -0,0 +1,76 @@
...
+static constexpr core::Float kBrakingDistance   = 20.0;   // m TODOLater:check!
+static constexpr core::Float kPi                = 3.14159265359;
+static constexpr core::Float kWheelCicumference = kPi * 0.1;  // m TODOLater: check!
+static constexpr core::Float kStripeDistance    = 6.0;
+static constexpr bool kIsKeyenceActive          = true;
+
+// define sensor checks return type
+enum class SensorChecks { kUnacceptable = 0, kAcceptable };
+
+struct Quartiles {
```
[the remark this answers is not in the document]
maxguy2001: Aw I meant to fix it. It's because encoders preprocessor uses one and imu preprocessor uses the other. In reality they should probably get combined into one I think?

lib/utils/kalman_higher_derivatives.cpp
maxguy2001: No. Here we are subtracting the last element of the vector from the total to remove values that are too old from calculation. We then add the new value onto this

SnickeyX (approved):
good to go

PR #108, opened 2023-03-23
MTC - CAN return processor errors + documentation

TomLonergan03:
Changed CanProcessor and ICanProcessor to return `core::Result::kFailure` if a processor fails to process its message

Also added doc comments to CAN classes

### DIFF
--- a/lib/io/can.hpp
+++ b/lib/io/can.hpp
@@ -41,13 +41,44 @@ struct sockaddr_can {
 
 class ICanProcessor {
  public:
-  virtual void processMessage(const io::CanFrame &frame) = 0;
+  /**
+   * @brief Called by ICan::receive() when a message with the given ID is received
+   * @param frame The CAN message to be processed
+   * @return core::Result::kSuccess if the message was processed successfully,
+   * core::Result::kFailure otherwise
+   */
+  virtual core::Result processMessage(const io::CanFrame &frame) = 0;
 };
 
 class ICan {
  public:
-  virtual core::Result send(const CanFrame &message)                                          = 0;
-  virtual core::Result receive()                                                              = 0;
+  /**
+   * @brief Send a CAN message
+   * @param message The message to be sent
+   * @return core::Result::kSuccess if the message was sent successfully core::Result::kFailure
+   * otherwise
+   * @note To send an extended CAN message, set the EFF flag in the can_id field by ORing it with
+   * CAN_EFF_FLAG
+   */
+  virtual core::Result send(const CanFrame &message) = 0;
+
+  /**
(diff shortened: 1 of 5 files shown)


lib/io/hardware_can.cpp
```diff
@@ -109,10 +109,12 @@ core::Result HardwareCan::receive()
     logger_.log(core::LogLevel::kFatal, "No CanProccessor associated with id %i", message.can_id);
     return core::Result::kFailure;
   }
+  core::Result process_successful = core::Result::kSuccess;
```
maxguy2001, line 112: Why do you make this object instead of just returning `kFailure` when the condition is met and `kSuccess` otherwise?

lib/io/can.hpp
```diff
@@ -41,13 +41,44 @@ struct sockaddr_can {
...
+   * and thus the actual ID will be frame.can_id - 0x80000000
+   */
+  virtual core::Result receive() = 0;
+
+  /**
+   * @brief Registers a processor to be called when a message with the given ID is received
+   * @param id The ID of the message to be processed
+   * @param processor The processor to be called when a message with the given ID is received
+   */
   virtual void addProcessor(const std::uint16_t id, std::shared_ptr<ICanProcessor> processor) = 0;
```
maxguy2001, line 82: Can this not fail too? Not 100% sure on context here but if this could fail then a `core::Result` return value might make sense here too?

lib/io/can.hpp
```diff
@@ -41,13 +41,44 @@ struct sockaddr_can {
 
 class ICanProcessor {
  public:
-  virtual void processMessage(const io::CanFrame &frame) = 0;
+  /**
+   * @brief Called by ICan::receive() when a message with the given ID is received
```
maxguy2001, line 45: Do you not want to include what the function actually does a little bit here?

lib/io/can.hpp
```diff
@@ -41,13 +41,44 @@ struct sockaddr_can {
...
+   * otherwise
+   * @note To send an extended CAN message, set the EFF flag in the can_id field by ORing it with
+   * CAN_EFF_FLAG
+   */
+  virtual core::Result send(const CanFrame &message) = 0;
+
+  /**
+   * @brief Attempt to receive a CAN message
+   * @return core::Result::kSuccess if a message was received and successfully processed by all
+   * registered processors, core::Result::kFailure if a message is not received successfully or if a
```
maxguy2001, line 68: Do we log which of the fail states occur?

maxguy2001 (changes requested):

kshxtij (approved):

davidbeechey (approved):

PR #110, opened 2023-03-27
SNS - IMD

sa-fx:
IMD requirements

- Retrieve the isolation status from the IMD
- Retrieve real-time values for resistance between the positive power supply terminal and the chassis, and the negative power supply terminal and the chassis

### DIFF
--- a/lib/debug/repl.cpp
+++ b/lib/debug/repl.cpp
@@ -724,6 +724,7 @@ void Repl::addImdCommands(const std::string &bus)
       logger_.log(core::LogLevel::kFatal, "Failed to run IMD::updateValues()");
       return;
     }
+    // Allow time for IMD to respond over CAN
     sleep(1);
     core::Result can_receive_result = can->receive();
     if (can_receive_result == core::Result::kFailure) {


lib/debug/repl.cpp
```diff
@@ -609,4 +636,38 @@ void Repl::addTemperatureCommands(const std::uint8_t bus, const std::uint8_t dev
...
 
+void Repl::addLowPowerCurrentCommands(const std::uint8_t bus, const std::uint8_t device_address)
+{
+  const auto optional_i2c = io::HardwareI2c::create(logger_, bus);
+  if (!optional_i2c) {
+    logger_.log(core::LogLevel::kFatal, "Failed to create I2C instance on bus %d", bus);
+    return;
+  }
+  const auto i2c = std::move(*optional_i2c);
+  const auto optional_low_power_current
```
maxguy2001, line 647: You don't check this actually returns something despite using a create function?

test/core/CMakeLists.txt
```diff
@@ -11,6 +11,6 @@ target_link_libraries(${target}
     hyped_core
     hyped_utils
     Eigen3::Eigen
-    Boost::boost
+    # Boost::boost
```
maxguy2001, line 14: ?

maxguy2001 (changes requested):

test/core/CMakeLists.txt
sa-fx: Done so that low power current sensor debug code could build for BBB as Boost cross-compiling wasn't working at this time
All LPC sensor debug code has now (hopefully) been removed from this PR

lib/debug/repl.cpp
sa-fx: An issue which can and should be considered when focusing on the LPC sensor - This code shouldn't have been edited with this PR
As such, all LPC sensor debug code has now (hopefully) been removed from this PR

lib/sensors/imd.hpp
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include <core/logger.hpp>
+#include <core/types.hpp>
+#include <io/can.hpp>
+
+namespace hyped::sensors {
+// Constants defined in IMD CAN reference manual
+constexpr std::uint8_t kRequestDataCanId            = 0xA100101;
+constexpr std::uint8_t kReturnDataCanId             = 0xA100100;
```
TomLonergan03, line 10: these 2 should be `uint32_t`

lib/sensors/imd.hpp
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include <core/logger.hpp>
+#include <core/types.hpp>
+#include <io/can.hpp>
+
```
TomLonergan03, line 6: `#include optional`

lib/sensors/imd.cpp
```diff
@@ -0,0 +1,66 @@
...
+std::optional<std::shared_ptr<Imd>> Imd::create(core::ILogger &logger,
+                                                const std::shared_ptr<io::ICan> can)
+{
+  std::optional<std::shared_ptr<Imd>> optional_imd = std::make_shared<Imd>(logger, can);
+  if (!optional_imd) {
+    logger.log(core::LogLevel::kFatal, "Failed to create IMD");
+    return std::nullopt;
+  }
+  std::shared_ptr<Imd> imd = *optional_imd;
+  can.addProcessor(kReturnDataCanId, imd);
```
TomLonergan03, line 13: also, i have messed up the typing on addProcessor, `id` should be a `uint32_t`. i forgot that extended can used longer ids. if you update hardware_can.[ch]pp and can.hpp then this will work.

lib/sensors/imd.hpp
```diff
@@ -0,0 +1,41 @@
...
+  std::uint16_t getRp();
+  std::uint16_t getRn();
+  std::uint8_t getIsolationState();
+
+ private:
+  core::ILogger &logger_;
+  std::shared_ptr<io::ICan> can_;
+  std::uint16_t rp_;
+  std::uint16_t rn_;
+  std::uint8_t iso_state_;
```
TomLonergan03, line 39: naming

TomLonergan03 (changes requested):
this should get it to build once you fix these issues, my bad on the can.hpp id type

lib/sensors/imd.hpp
sa-fx: Fair enough on _iso_state__
_rp__ and _rn__ came from the CAN reference manual, but I imagine code clarity is more important than consistency with a 3rd party manual

TomLonergan03: in this case i think so. if ctrf+f for "resistance positive" doesn't turn anything up in the manual probably add a comment, but rp and rn are meaningless to everything outside this specific class so it is better to be clear

lib/debug/repl.hpp
TomLonergan03, line 1: are the changes in here part of the PR? if not then remove them

lib/sensors/imd.hpp
```diff
@@ -0,0 +1,44 @@
...
+   */
+  core::Result updateValues();
+
+  /**
+   * @brief Saves values returned from the IMD
+   */
+  core::Result processMessage(const io::CanFrame &frame);
+
+  std::uint16_t getRp();
+  std::uint16_t getRn();
```
TomLonergan03, line 34: naming on these 2 also

TomLonergan03 (changes requested):
2 more things then good to go

TomLonergan03 (approved):

lib/sensors/imd.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+
+core::Result Imd::updateValues()
+{
+  io::CanFrame frame;
+  frame.can_id = CAN_EFF_FLAG | kRequestDataCanId;
+  // Always one byte for Request_mux
+  frame.can_dlc       = 1;
+  frame.data[0]       = kRequestIsolationResistances;
+  frame.data[1]       = 0;
```
maxguy2001, line 28: This feels quite repetitive, are you not better off putting this in a loop?

lib/sensors/imd.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+core::Result Imd::updateValues()
+{
+  io::CanFrame frame;
+  frame.can_id = CAN_EFF_FLAG | kRequestDataCanId;
+  // Always one byte for Request_mux
+  frame.can_dlc = 1;
+  frame.data[0] = kRequestIsolationResistances;
+  // Fill remaining data bits with 0
+  for (int i = 1; i < 8; i++) {
```
maxguy2001, line 29: `std::size_t` and `++i`

maxguy2001 (changes requested):
Tiny style guide thing but otherwise good to go

maxguy2001 (approved):

SnickeyX (approved):
minor changes, otherwise hope it gets tested

lib/debug/repl.cpp
```diff
@@ -700,6 +700,46 @@ void Repl::addTemperatureCommands(const std::uint8_t bus, const std::uint8_t dev
...
+  imd_request_values.name = "imd request values";
+  imd_request_values.description
+    = "Request values from the IMD for positive and negative resistances, and isolation state";
+  imd_request_values.handler = [this, imd, bus, can]() {
+    core::Result result = imd->updateValues();
+    if (result == core::Result::kFailure) {
+      logger_.log(core::LogLevel::kFatal, "Failed to run IMD::updateValues()");
+      return;
+    }
+    sleep(1);
```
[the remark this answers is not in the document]
sa-fx: Was advised by @TomLonergan03 so that enough time is given to allow CAN device to receive message and respond with its own. If there's a better way of doing this though, let me know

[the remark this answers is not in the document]
TomLonergan03: yeah for the repl this is the best way I've worked out to trigger `can_.receive()` and guarantee that the message has been returned. tbh you could probably immediately run `can_.receive()` but I'm not 100% sure.
in the actual code `can_.receive()` will be called in one of the main loops so will be dealt with asynchronously (relative to message send at least) hence why we have to sleep

PR #117, opened 2023-05-28
SNS - Wheel Encoder

TomLonergan03:
- encoder structure
- implement wheel encoder
- make constructor public
- fix logic
- store previous voltage
- add encoder to config file
- wheel encoder repl commands

### DIFF
--- a/lib/sensors/wheel_encoder.cpp
+++ b/lib/sensors/wheel_encoder.cpp
@@ -2,4 +2,28 @@
 
 namespace hyped::sensors {
 
+WheelEncoder::WheelEncoder(core::ILogger &logger, std::shared_ptr<io::IAdc> adc)
+    : logger_(logger),
+      adc_(adc),
+      count_(0)
+{
+}
+
+std::uint64_t WheelEncoder::getCount()
+{
+  return count_;
+}
+
+core::Result WheelEncoder::updateCount()
+{
+  const auto optional_voltage = adc_->readValue();
+  if (!optional_voltage) {
+    logger_.log(core::LogLevel::kFatal, "Failed to read wheel encoder value from ADC");
+    return core::Result::kFailure;
+  }
+  const core::Float voltage = *optional_voltage;
+  if (voltage > kVoltageThreshold) { count_++; }
+  return core::Result::kSuccess;
+}
+
 }  // namespace hyped::sensors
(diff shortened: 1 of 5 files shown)


lib/sensors/wheel_encoder.cpp
```diff
@@ -0,0 +1,36 @@
...
+{
+  const auto optional_voltage = adc_->readValue();
+  if (!optional_voltage) {
+    logger_.log(core::LogLevel::kFatal, "Failed to read wheel encoder value from ADC");
+    return core::Result::kFailure;
+  }
+  const core::Float voltage = *optional_voltage;
+  if (voltage > kVoltageThreshold && previous_voltage_ < kVoltageThreshold) { count_++; }
+  previous_voltage_ = voltage;
+  return core::Result::kSuccess;
```
SnickeyX, line 28: If we keep getting a value over the thresh, at what point do we think the encoders aren't working. Ik nav pays a lot of attention to these so want to make sure we report any possible failure asap.

maxguy2001: Yeah these are our singular most accurate sensor I think. If these give faulty data, we'll only spot it if it's disagreeing with our other data but this data being unreliable could potentially leave us without a ground truth

maxguy2001 (changes requested):

lib/sensors/wheel_encoder.cpp
TomLonergan03: it depends on velocity. we can probably calculate a maximum high period before we stop trusting it based off of minimum velocity we expect to hold

lib/sensors/wheel_encoder.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+core::Result WheelEncoder::updateCount()
+{
+  const auto optional_voltage = adc_->readValue();
+  if (!optional_voltage) {
+    logger_.log(core::LogLevel::kFatal, "Failed to read wheel encoder value from ADC");
+    return core::Result::kFailure;
+  }
+  const core::Float voltage = *optional_voltage;
+  if (voltage > kVoltageThreshold && previous_voltage_ < kVoltageThreshold) { count_++; }
```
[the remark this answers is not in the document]
TomLonergan03: will fix this, however can you remind me what the reason for this is

PR #122, opened 2023-06-03
feat(tel): Write telemetry data and show historical data

arjunnaha:
- Add `MqttIngestion` and `MeasurementService` to read data from MQTT and write it to Influx (database)
- Add `HistoricalService` and `historical-data-plugin` to query data from Influx and return it back to OpenMCT

image


davidbeechey (approved):
Looks great! Like the new `types` package :)

